PR #4, opened 2024-08-17
fix(components): scan panic error, docker-container error handling by converting state.error to string

gyuho:
Before:

Cannot handle error like

> {"level":"debug","ts":"2024-08-17T01:53:44Z","caller":"query/poller.go:134","msg":"polling error","id":"docker-container","error":"Error response from daemon: client version 1.44 is too new. Maximum supported API version is 1.43"}

Screenshot 2024-08-17 at 10 04 37 AM

Screenshot 2024-08-17 at 10 02 07 AM

After:

Screenshot 2024-08-17 at 10 04 51 AM

### DIFF
--- a/cmd/gpud/command/up.go
+++ b/cmd/gpud/command/up.go
@@ -20,7 +20,7 @@ func cmdUp(cliContext *cli.Context) (retErr error) {
 				retErr = err
 			}
 		} else {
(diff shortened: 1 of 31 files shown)


hm2501 (approved):
LGTM

PR #7, opened 2024-08-19
feat(internal/session): make it more unit testable

gyuho:


internal/session/session.go
```diff
@@ -48,144 +71,151 @@ type Body struct {
...
+		reader, writer := io.Pipe()
+		goroutineCloseCh := make(chan any)
+		go func() {
+			log.Logger.Debug("session writer created")
+			for {
+				select {
+				case <-s.writerCloseCh:
+					log.Logger.Debug("session writer closed")
+					writer.Close()
+					return
```
gyuho, line 102: @cardyok We didn't `return` before, now we do. Please check.

internal/session/session.go
```diff
@@ -48,144 +71,151 @@ type Body struct {
...
+		if err != nil {
+			log.Logger.Errorf("session writer: error making request: %v, retrying", err)
 			close(goroutineCloseCh)
-			time.Sleep(3 * time.Second)
+			continue
 		}
-	}()
+
+		log.Logger.Debugf("session writer: unexpected closed, resp: %v %v, reconnecting...", resp.Status, resp.StatusCode)
+		close(goroutineCloseCh)
```
gyuho, line 137: @cardyok  We didn't `close(goroutineCloseCh)`, now we do. Please check.

internal/session/session.go
```diff
@@ -48,144 +71,151 @@ type Body struct {
...
+				s.writerCloseCh <- true
+				break
 			}
-			close(goroutineCloseCh)
-			time.Sleep(3 * time.Second)
+
+			s.reader <- content
 		}
-	}()
+		close(goroutineCloseCh)
```
gyuho, line 205: @cardyok 

Before

```go
			decoder := json.NewDecoder(resp.Body)
			for {
				var content Body
				err = decoder.Decode(&content)
				if err != nil {
					if !errors.Is(err, io.EOF) {
						fmt.Println("Error reading response:", err)
					}
					s.writerCloseCh <- true
					time.Sleep(3 * time.Second)
					break
				}
				s.reader <- content
			}
			close(goroutineCloseCh)
			time.Sleep(3 * time.Second)
```

Now:

```go
		decoder := json.NewDecoder(resp.Body)
		for {
			var content Body
			err = decoder.Decode(&content)
			if err != nil {
				if !errors.Is(err, io.EOF) {
					fmt.Println("Error reading response:", err)
				}

				s.writerCloseCh <- true
				break
			}

			s.reader <- content
		}
		close(goroutineCloseCh)
```

Just cleanups for time.Sleep

internal/session/session.go
cardyok: this is intentionally left out to ensure that only one pair of read/write connection is there..

internal/session/session.go
cardyok: similarly, here is forcing writer's goroutine to be closed after reader finishes closing, but i think returning here should be fine.

internal/session/session.go
gyuho: @cardyok removed

cardyok (approved):
lgtm

PR #10, opened 2024-08-19
fix(dmesg): add fallback dmesg watch command, remove confusing log message

gyuho:
`Commands [][]string` run in sequence, so in case `-w` is not supported in some machines, we should continue to the next commands.

> log.Logger.Debugw("command was terminated (exit code -1) by the root context cancellation", "cmd", sr.cmd.String(), "contextError", sr.ctx.Err())

to clarify the logging (this is expected when we run `gpud down`)

Previous logging was confusing:

> {"level":"warn","ts":"2024-08-19T18:20:17Z","caller":"tail/streamer_command.go:138","msg":"error waiting for command to finish","error":"signal: terminated","cmd":"/usr/bin/bash /tmp/streamer-from-command2548997474.txt"}
{"level":"warn","ts":"2024-08-19T18:27:11Z","caller":"tail/streamer_command.go:138","msg":"error waiting for command to finish","error":"signal: terminated","cmd":"/usr/bin/bash /tmp/streamer-from-command4028497856.txt"}

### DIFF
--- a/components/dmesg/config.go
+++ b/components/dmesg/config.go
@@ -64,8 +64,8 @@ func DefaultConfig() Config {
 			BufferSize: query_log_config.DefaultBufferSize,
 
 			Commands: [][]string{
-				{"dmesg", "--ctime", "--nopager", "--buffer-size", "163920", "-w"},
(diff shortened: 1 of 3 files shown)


cardyok (approved):
lgtm

PR #16, opened 2024-08-20
fix(goreleaser): goreleaser workflow to fetch git tag

gyuho:


.github/workflows/goreleaser.yml
```diff
@@ -79,4 +82,4 @@ jobs:
           prerelease: false
           body: Latest builds from the last commit
           files: |
-            ./dist/gpud-${{ matrix.job.target }}.tar.gz
+            ./dist/gpud_${{steps.get_tag_version.outputs.TAG_VERSION}}_${{ matrix.job.target }}.tar.gz
```
hm2501, line 85: can we make the suffix `.tgz` instead of `.tar.gz`?

hm2501 (approved):
LGTM

PR #26, opened 2024-08-21
feat(pkg/process): initial commit, fix dmesg scan on non-root

gyuho:
We run external commands as an external process within gpud everything (e.g., dmesg, dmesg -w, gpud diagnosis for nvidia* commands). We should have a shared package that runs those commands. (optiona) Will be useful for next phase of GPUd where we can maintain the on-host components.

### DIFF
--- a/components/query/log/tail/streamer_command.go
+++ b/components/query/log/tail/streamer_command.go
@@ -3,12 +3,12 @@ package tail
 import (
 	"bufio"
 	"context"
-	"os"
-	"os/exec"
+	"fmt"
 	"time"
 
 	query_log_filter "github.com/leptonai/gpud/components/query/log/filter"
 	"github.com/leptonai/gpud/log"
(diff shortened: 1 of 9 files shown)


components/query/log/tail/streamer_command.go
```diff
@@ -21,38 +21,29 @@ func NewFromCommand(ctx context.Context, commands [][]string, opts ...OpOption)
...
-	stdoutScanner := bufio.NewScanner(stdout)
 
-	stderr, err := cmd.StderrPipe()
-	if err != nil {
-		return nil, err
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case err := <-p.Wait():
+		return nil, fmt.Errorf("command exited unexpectedly: %w", err)
```
gyuho, line 36: This is the bug fix. Previously, it was just ignored.

hm2501 (approved):
LGTM

PR #27, opened 2024-08-21
adding ubuntu 20.04

sysbot:


gyuho (approved):
Lgtm

Thanks for contribution!

We will need to change our release file name format to include os name. Should be an easy change. Do you have specific needs for Ubuntu 20.04 builds?

gyuho:
@sysbot Our build should be Ubuntu version agnostic, except sqlite dependency. If you have any issues with Ubuntu 20.04, we can certainly provide a separate release binaries. :)

gyuho:
@sysbot  I pushed one more commit to support per Ubuntu version release. We will merge and test this change. /cc @hm2501

PR #29, opened 2024-08-22
feat(update): flag for controlling auto update

hm2501:


internal/session/serve.go
```diff
@@ -62,6 +62,9 @@ func (s *Session) serve() {
 			if !systemdManaged {
 				log.Logger.Debugw("gpud is not managed with systemd")
 				response.Error = fmt.Errorf("gpud is not managed with systemd")
+			} else if !s.enableAutoUpdate {
+				log.Logger.Debugw("auto update is disabled")
+				response.Error = fmt.Errorf("auto update is disabled")
```
gyuho, line 67:
```suggestion
				response.Error = errors.New("gpud is not managed with systemd")
			} else if !s.enableAutoUpdate {
				log.Logger.Debugw("auto update is disabled")
				response.Error = errors.New("auto update is disabled")
```

gyuho (approved):
lgtm

Issue #32, opened 2024-08-22
gpud scan (dmesg) fills the stderr with kernel logs, and fails with unknown flags


eicca:
Hi again :)

I'm running
```
sudo gpud scan 
```
And it works well, except that after displaying the results of the scan, it shows the kernel log from dmesg. This makes it hard to read the scan report, so I use just this for now:
```
sudo ./gpud scan 2>/dev/null
```

It would be amazing if there would be an option to not display dmesg by default.

Thanks in advance :heart:

gyuho:
`gpud scan` might have failed, thus dumping all the output to your terminal.

Could you share the error message for your `gpud scan` command?

And regardless, https://github.com/leptonai/gpud/pull/33 will remove that verbose output from the scan command.

Thanks for the report.

eicca:
The `gpud scan` does all the steps well until the 
```
⌛ scanning dmesg for 5000 lines
```
And then it errors there with:
```
gpud: failed to execute command: exit status 1 ([    0.000000] kernel: Linux version blah blah and other kernel details
```
unfortunately, it doesn't really say why it fails or what kind of command was it.

On my local GPU machine I don't have such error.

gyuho:
> On my local GPU machine I don't have such error.

@eicca Can you share your output `gpud --version`?

gyuho:
We've fixed a few releases ago, and https://github.com/leptonai/gpud/releases/tag/v0.0.1-alpha7 has been released :)

Please give it a try and let us know if things are still breaking :)

eicca:
Hi, I tried the new release and the kernel log doesn't fill the terminal anymore :+1: 

However, the dmesg scan error is still there:

```
⌛ scanning dmesg for 5000 lines
{"level":"warn","ts":"2024-08-29T13:28:24Z","caller":"process/process.go:221","msg":"command exited with non-zero status","error":"exit status 1","cmd":"/usr/bin/bash /tmp/tmpbash3729988194.bash","exitCode":1}
{"level":"warn","ts":"2024-08-29T13:28:24Z","caller":"process/process.go:228","msg":"process exited with error","error":"exit status 1"}
exit status 1
```

gyuho:
@eicca Thanks for the confirmation.

I suspect this is an OS-specific error.

Could you share your output for

> sudo dmesg --ctime --nopager --buffer-size 163920 --since '1 hour ago'

command?

(this is what the gpud scan command is running)

eicca:
I get this error:

[...]

PR #33, opened 2024-08-22
fix(query/log/tail): do not show output on scan failures

gyuho:
Address https://github.com/leptonai/gpud/issues/32.


xiang90:
lgtm

xiang90 (approved):

PR #41, opened 2024-08-27
fix(cmd/gpud): clean up "gpud down" logs

gyuho:
Before:

```
gpud down

 ✘ failed to stop systemd unit 'gpud.service': systemctl stop failed: exit status 5 output: Failed to stop gpud.service: Unit gpud.service not loaded.

gpud: systemctl stop failed: exit status 5 output: Failed to stop gpud.service: Unit gpud.service not loaded.
```

After:

```
gpud down

✔ gpud is not running (no-op)
```

### DIFF
--- a/cmd/gpud/command/down.go
+++ b/cmd/gpud/command/down.go
@@ -4,9 +4,10 @@ import (
 	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


hm2501 (approved):
LGTM

PR #45, opened 2024-08-30
fix(dmesg): fallback in case "dmesg --since" flag doesn't exist in older versions

gyuho:
Address https://github.com/leptonai/gpud/issues/32.


hm2501 (approved):
LGTM

PR #47, opened 2024-09-02
fix(docs): pkg.go.dev links, add Makefile CGO_ENABLED=1

flyer103:
Fix invalid document links, delete invalid document content, and add `CGO_ENABLED=1` configuration when compiling `gpud` in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ FORCE:
 
 define BUILD_BINARY
 @echo "$(WHALE) $@"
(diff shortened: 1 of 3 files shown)


gyuho:
@flyer103 Thanks for the contribution. I've pushed a commit to fix the pkg.go.dev links. Merging on greens.

gyuho (approved):
lgtm

PR #48, opened 2024-09-02
fix(nvidia): skip clock events NVML check if not supported by old drivers

gyuho:
Reported by @cardyok:

image

Not all NVIDIA drivers support this symbol. And if we don't do this, the CGO call just exits the whole process.

(ref. https://github.com/NVIDIA/go-nvml/pull/123)

### DIFF
--- a/components/accelerator/nvidia/query/nvml/nvml.go
+++ b/components/accelerator/nvidia/query/nvml/nvml.go
@@ -43,6 +43,9 @@ var _ Instance = (*instance)(nil)
 type instance struct {
 	mu sync.RWMutex
 
+	driverVersion        string
+	clockEventsSupported bool
+
 	rootCtx    context.Context
 	rootCancel context.CancelFunc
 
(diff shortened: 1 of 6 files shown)


components/accelerator/nvidia/query/nvml/nvml.go
```diff
@@ -91,19 +94,53 @@ type DeviceInfo struct {
...
+	}
+
+	major, minor, patch = parsed[0], parsed[1], parsed[2]
+	return major, minor, patch, nil
+}
+
+// clock events are supported in versions 535 and above
+// otherwise, CGO call just exits with
+// undefined symbol: nvmlDeviceGetCurrentClocksEventReasons
+func ClockEventsSupportedVersion(major, minor, patch int) bool {
```
hm2501, line 140: do we need `minor` and `patch` params?

hm2501 (approved):
LGTM with one minor comment

components/accelerator/nvidia/query/nvml/nvml.go
gyuho: Let me remove :)

PR #50, opened 2024-09-04
feat(nvidia/nvml): include device uuid for xid event

gyuho:


components/accelerator/nvidia/query/nvml/xid.go
```diff
@@ -109,18 +110,30 @@ func (inst *instance) pollXidEvents() {
 			}
 		}
 
+		var deviceUUID string
+		var deviceUUIDErr error
+		deviceUUID, ret = e.Device.GetUUID()
+		if ret != nvml.SUCCESS {
+			// "If we cannot reliably determine the device UUID, we mark all devices as unhealthy."
+			// ref. nvidia/k8s-device-plugin/internal/rm/health.go
+			deviceUUIDErr = fmt.Errorf("failed to get device UUID: %v", nvml.ErrorString(ret))
```
hm2501, line 119: optional nit: it might be friendly if we initialize `deviceUUID` with something like "(failed to get)"?

hm2501 (approved):
LGTM with optional nit

components/accelerator/nvidia/query/nvml/xid.go
gyuho: I think that will complicate our logic. If no device id such as empty we should make it as unhealthy.

hm2501: got it, then it's fine

PR #51, opened 2024-09-06
fix(nvidia/nvml): mark xid 68 as user app error, document

gyuho:
c.f., https://github.com/NVIDIA/k8s-device-plugin/issues/945


hm2501 (approved):
LGTM

PR #52, opened 2024-09-06
fix(components/fd): use system-wide file descriptor limit, add default 1-million threshold limit, remove "_avg" metrics in fd component

gyuho:
We should not use rlimit since it's specific to the process (in this case, fd limit for the gpud process).

### DIFF
--- a/components/fd/component_output.go
+++ b/components/fd/component_output.go
@@ -209,14 +209,11 @@ func CreateGet(cfg Config) query.GetFunc {
 			return nil, err
 		}
 
-		var usedPct float64
-		if limit > 0 {
-			usageVal := runningPIDs // for mac
-			if usage > 0 {
-				usageVal = usage
-			}
-			usedPct = float64(usageVal) / float64(limit) * 100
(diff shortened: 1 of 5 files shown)


components/fd/component_output.go
```diff
@@ -122,72 +161,97 @@ var (
...
 	return defaultPoller
 }
 
-func Get(ctx context.Context) (_ any, e error) {
-	defer func() {
-		if e != nil {
-			components_metrics.SetGetFailed(Name)
-		} else {
-			components_metrics.SetGetSuccess(Name)
+func CreateGet(cfg Config) query.GetFunc {
```
hm2501, line 172: Is this possible to decouple calculation logics from low level syscalls or file access and write UTs?

hm2501 (approved):
I think it's better to have some UTs for calculation logics but overall LGTM

components/fd/component_output.go
gyuho: done

PR #62, opened 2024-09-10
fix(components/docker): do not set not healthy if docker client version incompatible

gyuho:


components/docker/container/component_output.go
```diff
@@ -53,6 +59,12 @@ func ParseStateDockerContainer(m map[string]string) ([]DockerContainer, error) {
 }
 
 func (o *Output) describeReason() string {
+	if o.IsErrDockerClientVersionNewerThanDaemon {
+		if o.DockerClientError == "" {
+			return "not supported; docker client version is newer than the daemon version (needs downgrading docker client in GPUd or upgrading docker daemon in the host)"
```
cardyok, line 63: maybe we shouldn't mention downgrading GPUd in err message..

gyuho: done

cardyok (approved):

PR #66, opened 2024-09-12
feat(internal/session): add missing writer close for session writer

gyuho:


xiang90:
is this a bug? shall we add a test?

gyuho:
@cardyok to confirm. I now remember this was intentional -- if that's the case, we shall document.

cardyok (approved):

PR #77, opened 2024-09-24
fix(nvidia/peermem): do not decide health based on ibcore peermem module

gyuho:


components/accelerator/nvidia/query/query.go
```diff
@@ -294,6 +294,24 @@ func (o *Output) YAML() ([]byte, error) {
 	return yaml.Marshal(o)
 }
 
+func (o *Output) GPUCounts() int {
+	if o == nil {
+		return 0
+	}
+	cnts := o.SMI.AttachedGPUs
```
cardyok, line 301: check SMI nil here?

components/accelerator/nvidia/query/query.go
```diff
@@ -294,6 +294,24 @@ func (o *Output) YAML() ([]byte, error) {
...
+	if cnts == 0 { // in case of "nvidia-smi" failure
+		cnts = len(o.NVML.DeviceInfos)
+	}
+	return cnts
+}
+
+func (o *Output) GPUProductName() string {
+	if o == nil || len(o.SMI.GPUs) == 0 {
```
cardyok, line 309: same

cardyok (approved):

PR #78, opened 2024-09-24
feat(internal/server): dynamically refresh containerd, docker, kubelet components

gyuho:


cardyok:
code lgtm, doing more testing on it

config/default.go
```diff
@@ -351,3 +298,83 @@ func DefaultFifoFile() (string, error) {
...
 }
+
+func DefaultContainerdComponent(ctx context.Context) (any, bool) {
+	if runtime.GOOS != "linux" {
+		log.Logger.Debugw("ignoring default containerd pod checking since it's not linux", "os", runtime.GOOS)
+		return nil, false
+	}
+
+	containerdSocketExists := false
+	containerdRunning := false
```
cardyok, line 309: for containerd/docker when its not running, I think component also need to be registered. just unhealthy, and for kubelet, we can use pidof kubelet instead of check 10255, the rest lgtm

cardyok (approved):

config/default.go
gyuho: done

PR #80, opened 2024-09-25
feat(nvidia): track row remapping, RMA/GPU reset status

gyuho:
This should alert, since this GPU requires a reset to actually remap the rows:

```
    Remapped Rows
        Correctable Error                 : 0
        Uncorrectable Error               : 4
        Pending                           : Yes
```

ref. https://docs.nvidia.com/deploy/a100-gpu-mem-error-mgmt/index.html

### DIFF
--- a/components/accelerator/nvidia/remapped-rows/component_output.go
+++ b/components/accelerator/nvidia/remapped-rows/component_output.go
@@ -13,7 +13,14 @@ import (
 )
 
 func ToOutput(i *nvidia_query.Output) *Output {
-	o := &Output{}
+	if i == nil {
+		return nil
+	}
+
+	o := &Output{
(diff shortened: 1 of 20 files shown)


components/accelerator/nvidia/remapped-rows/component_output.go
```diff
@@ -0,0 +1,166 @@
...
+func ToOutput(i *nvidia_query.Output) *Output {
+	if i == nil {
+		return nil
+	}
+
+	o := &Output{
+		MemoryErrorManagementCapabilities: i.MemoryErrorManagementCapabilities,
+	}
+
+	for _, g := range i.SMI.GPUs {
```
cardyok, line 24: need to check i.SMI != nil

cardyok (approved):

PR #81, opened 2024-09-26
feat(nvidia/xid,sxid): rename Detail.ID to XID, add required actions for XID/SXID events

gyuho:


components/accelerator/nvidia/query/xid/xid.go
```diff
@@ -19,6 +24,15 @@ type Detail struct {
 	BusError               bool   `json:"bus_error"`
 	ThermalIssue           bool   `json:"thermal_issue"`
 	FBCorruption           bool   `json:"fb_corruption"`
+
+	// Set to true if the error requires a GPU reset.
+	// GPU reset and system reboot are independent actions.
+	// Both fields are set to true, if the error requires "any" of the two actions.
+	RequiresGPUReset bool `json:"requires_gpu_reset"`
```
cardyok, line 31: shall we make this field more generic to be a separate type so it can be easily extendable

cardyok (approved):

PR #82, opened 2024-09-27
Update mothership endpoint

cardyok:


components/info/component.go
```diff
@@ -64,7 +67,7 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 		},
 		{
 			Name:      StateNameAnnotations,
-			Healthy:   true,
+			Healthy:   err == nil,
```
gyuho, line 70: Should we have a separate state? since not all gpud has this file?

go-pkg/randutil/randutil.go
```diff
@@ -0,0 +1,31 @@
+package randutil
```
gyuho, line 1: doesn't seem used?

components/accelerator/nvidia/ecc/component_output.go
```diff
@@ -13,19 +13,20 @@ import (
...
-			continue
-		}
+	if i != nil && i.SMI == nil {
+		for _, g := range i.SMI.GPUs {
+			if g.ECCErrors == nil {
+				continue
+			}
 
-		o.ErrorCountsSMI = append(o.ErrorCountsSMI, *g.ECCErrors)
+			o.ErrorCountsSMI = append(o.ErrorCountsSMI, *g.ECCErrors)
```
gyuho, line 22: See https://github.com/leptonai/gpud/pull/83. Cherry-picked this change to a separate PR.

go-pkg/randutil/randutil.go
cardyok: yes, not needed anymore, will remove it

components/info/component.go
cardyok: this is just for manually creating unhealthy state, will be removed once pr is ready for review

cmd/gpud/command/command.go
```diff
@@ -377,6 +408,55 @@ cat summary.txt
...
+					Usage: "provider gpu shape of the machine",
+				},
+				cli.StringFlag{
+					Name:  "provider-instance-type",
+					Usage: "provider instance type of the machine",
+				},
+				cli.Int64Flag{
+					Name:  "total-cpu",
+					Usage: "total cpu of the machine",
+					Value: 30000,
```
gyuho, line 448: Can we comment we will (TODO) auto derive, if not set?

cmd/gpud/command/command.go
```diff
@@ -377,6 +408,55 @@ cat summary.txt
...
+					Name:  "provider-gpu-shape",
+					Usage: "provider gpu shape of the machine",
+				},
+				cli.StringFlag{
+					Name:  "provider-instance-type",
+					Usage: "provider instance type of the machine",
+				},
+				cli.Int64Flag{
+					Name:  "total-cpu",
+					Usage: "total cpu of the machine",
```
gyuho, line 447: why do we need this?

cardyok: will change to /proc/cpuinfo

cmd/gpud/command/command.go
cardyok: sure thing

cmd/gpud/command/join.go
```diff
@@ -0,0 +1,190 @@
...
+	defer rootCancel()
+	endpoint := cliContext.String("endpoint")
+	clusterName := cliContext.String("cluster-name")
+	publicIP := cliContext.String("public-ip")
+	provider := cliContext.String("provider")
+	providerGPUShape := cliContext.String("provider-gpu-shape")
+	providerInstanceType := cliContext.String("provider-instance-type")
+	xrayNeeded := cliContext.Bool("xray-needed")
+	totalCPU := cliContext.Int64("total-cpu")
+	extraInfo := cliContext.String("extra-info")
```
gyuho, line 36: Add TODOs to auto derive default values (e.g., GPU type), and also prompt the user to input if not set.

cmd/gpud/command/join.go
```diff
@@ -0,0 +1,190 @@
...
+	gzipReader, err := gzip.NewReader(body)
+	if err != nil {
+		return "", fmt.Errorf("failed to create gzip reader: %w", err)
+	}
+	defer gzipReader.Close()
+
+	tarReader := tar.NewReader(gzipReader)
+
+	for {
+		header, err := tarReader.Next()
```
gyuho, line 120: need a separate function in a separate package, but for now, we can keep everything here :)

cmd/gpud/command/command.go
```diff
@@ -111,7 +112,38 @@ nohup sudo gpud run &>> <your log file path> &
...
+					Name:  "region",
+					Usage: "region of target cluster",
+				},
+				cli.StringFlag{
+					Name:  "cluster",
+					Usage: "name of target cluster",
+				},
+				cli.StringFlag{
+					Name:  "role",
+					Usage: "role arn",
```
gyuho, line 138:
```suggestion
					Usage: "role",
```

in case we support non-AWS roles?

cmd/gpud/command/join.go
```diff
@@ -0,0 +1,297 @@
...
+
+	_, productName, err := accelerator.DetectTypeAndProductName(rootCtx)
+	if err != nil {
+		return err
+	}
+
+	// network section
+	publicIP, _ := login.PublicIP()
+	region := "unknown"
+	detectProvider := ""
```
gyuho, line 75: Remove this? Currently hard-coded to tailscale only since we only talk to tailscale public DERP servers:
 https://github.com/leptonai/gpud/blob/871533e7efcff60e1b1addeed7dd8c94d9845bbc/pkg/latency/edge/derp.go#L73-L74

(this has nothing to do with node setup or GPUs)

gyuho (approved):

PR #86, opened 2024-09-30
feat(nvidia): track "ECC mode" (enabled/disabled) using nvidia-smi and NVML

gyuho:
Especially, `nvmlDeviceGetMemoryErrorCounter` requires "ECC mode" to be enabled.

See https://docs.nvidia.com/deploy/nvml-api/group__nvmlDeviceQueries.html#group__nvmlDeviceQueries_1g30900e951fe44f1f952f0e6c89b0e2c1.


cardyok:
image
ecc mode needs Enable/Disable for json field

xiang90:
Need UT

gyuho:
@xiang90 @cardyok  Added UTs, in addition to the ones in https://github.com/leptonai/gpud/pull/87 (same package).

PR #87, opened 2024-09-30
feat(nvidia/ecc): rename state name key to "ecc" (from ecc_errors)

gyuho:
https://github.com/leptonai/gpud/pull/86 is now tracking "ECC Mode", regardless of its error counts.

Thus, renaming the state key to "ecc", not "ecc-errors".

### DIFF
--- a/components/accelerator/nvidia/ecc/component_output.go
+++ b/components/accelerator/nvidia/ecc/component_output.go
@@ -78,22 +78,22 @@ func ParseOutputJSON(data []byte) (*Output, error) {
 }
 
 const (
-	StateNameECCErrors = "ecc_errors"
+	StateNameECC = "ecc"
 
-	StateKeyECCErrorsData           = "data"
-	StateKeyECCErrorsEncoding       = "encoding"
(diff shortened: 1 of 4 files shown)


components/accelerator/nvidia/ecc/component_output.go
```diff
@@ -66,22 +66,22 @@ func ParseOutputJSON(data []byte) (*Output, error) {
...
 
-	StateKeyECCErrorsData           = "data"
-	StateKeyECCErrorsEncoding       = "encoding"
-	StateValueECCErrorsEncodingJSON = "json"
+	StateKeyECCData           = "data"
+	StateKeyECCEncoding       = "encoding"
+	StateValueECCEncodingJSON = "json"
 )
 
 func ParseStateECCErrors(m map[string]string) (*Output, error) {
```
xiang90, line 76: do we have a unit test for this one? if not, let us add one.

xiang90:
LGTM

gyuho:
@xiang90 UTs have been added.

cardyok (approved):

PR #88, opened 2024-09-30
feat(nvidia/query): shorter timeouts for "nvidia-smi" calls

gyuho:


cardyok (approved):
lgtm, I think can be even shorter than 30s (in case systemd try to restart service comes earlier)

gyuho:
@cardyok Let's try 30-second first and if stable, we can shorten further :)

PR #89, opened 2024-09-30
feat(nvidia/xid,sxid,remapped rows): add required actions field to /states, /events

gyuho:


components/common/action.go
```diff
@@ -8,4 +8,7 @@ type RequiredActions struct {
 	ResetGPU bool `json:"reset_gpu"`
 	// Set to true if the error requires a system reboot.
 	RebootSystem bool `json:"reboot_system"`
+
+	// A list of descriptions for the required actions.
+	Descriptions []string `json:"descriptions"`
```
xiang90, line 13: Let us name the required action to suggested action.

components/common/action.go
```diff
@@ -8,4 +8,7 @@ type RequiredActions struct {
 	ResetGPU bool `json:"reset_gpu"`
```
xiang90, line 8: let us make this a list of RepaireActionType instead of a few boolean types. Now we should only have two types: 1. Reboot 2. RepairHardware (remove reset GPU. just use reboot)

xiang90:
Need UT

components/common/action.go
gyuho: Done

components/common/action.go
```diff
@@ -1,11 +1,78 @@
...
+		return
+	}
+
+	existingActions := make(map[RepairActionType]struct{})
+	for _, action := range s.RepairActions {
+		existingActions[action] = struct{}{}
+	}
+	for _, action := range other.RepairActions {
+		if _, ok := existingActions[action]; !ok {
+			s.RepairActions = append(s.RepairActions, action)
```
cardyok, line 65: since we only have two types of actions for now, shall we keep appending to the descriptions and do deduplicate for RepairActions, only keep the most significant one? the rest lgtm

gyuho: @cardyok I think we want to keep the both.

Reboot system doesn't necessarily require repair hardware, vice versa.

cardyok (approved):

PR #93, opened 2024-10-01
fix(components/fd): rename "fd_max_file_exists" to "fd_limit_supported", fix get limit on darwin

gyuho:
Screenshot 2024-10-01 at 8 59 22 PM

### DIFF
--- a/components/fd/get_limit_linux.go
+++ b/components/fd/get_limit_linux.go
@@ -1,3 +1,16 @@
+// Copyright 2019 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


components/fd/get_limit_linux.go
```diff
@@ -0,0 +1,58 @@
...
+func getLimit() (uint64, error) {
+	data, err := os.ReadFile(fileMaxLinux)
+	if err != nil {
+		return 0, err
+	}
+	return strconv.ParseUint(strings.TrimSpace(string(data)), 10, 64)
+}
+
+// "process_open_fds" in prometheus collector
+// ref. https://github.com/prometheus/client_golang/blob/main/prometheus/process_collector_other.go
```
xiang90, line 32: do we need add prom license if this code is copied over

gyuho: +1. Added.

cardyok (approved):

PR #96, opened 2024-10-03
feat(pkg/reboot): initial commit

gyuho:


pkg/reboot/reboot.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Set true to run "systemctl reboot".
+func WithSystemctl(b bool) OpOption {
+	return func(op *Op) {
+		op.useSystemctl = b
+	}
+}
+
+// Specifies a file to create before rebooting.
+func WithFileToCreateBeforeReboot(file string) OpOption {
```
cardyok, line 39: these two functions (With*) won't be needed, we can just do a generic reboot, custom file injection will be done on node initialization.

gyuho: removed

cardyok (approved):

PR #99, opened 2024-10-04
feat(nvidia/ibstat): check "Physical state" as fallback

gyuho:
Should not happen... but in case our hard-coded ibstat "State" string match fails (e.g., breaking API changes), we fallback to "Physical state" for link status checking.

```
CA 'mlx5_8'
	CA type: MT4125
	Number of ports: 1
	Firmware version: 22.39.1002
	Hardware version: 0
	Node GUID: 0xe8ebd303003307df
	System image GUID: 0xe8ebd303003307de
	Port 1:
		State: Down
		Physical state: Disabled
```

### DIFF
--- a/components/accelerator/nvidia/query/ibstat.go
+++ b/components/accelerator/nvidia/query/ibstat.go
@@ -42,8 +42,8 @@ func RunIbstat(ctx context.Context) (*IbstatOutput, error) {
 }
 
 var (
-	ErrIbstatOutputBrokenStateDown        = errors.New("ibstat output unexpected; found State: Down")
-	ErrIbstatOutputBrokenPhysicalDisabled = errors.New("ibstat output unexpected; found Physical state: Disabled")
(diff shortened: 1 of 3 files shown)


components/accelerator/nvidia/query/ibstat.go
```diff
@@ -40,10 +41,22 @@ func RunIbstat(ctx context.Context) (*IbstatOutput, error) {
 	return o, nil
 }
 
+var (
+	ErrIbstatOutputBrokenStateDown        = errors.New("ibstat output seems broken; found State: Down")
```
hexfusion, line 45:
```suggestion
	ErrIbstatOutputBrokenStateDown        = errors.New("ibstat output unexpected; found State: Down")
```

hexfusion (commented):
:wave: :)

cardyok (approved):

PR #108, opened 2024-10-09
fix(components/accelerator-nvidia-ecc): do not unhealthy when driver recovers uncorrectable ecc errors

gyuho:
> when an uncorrectable ECC error is detected, the NVIDIA driver software will perform error recovery.

Meaning, as long as NVIDIA performs row remapping, we don't need to mark this unhealthy. Instead, we monitor remapping-rows component for potential system reboot or gpu reset suggested actions

ref. https://docs.nvidia.com/deploy/a100-gpu-mem-error-mgmt/index.html

Fix

Screenshot 2024-10-09 at 4 52 55 PM

Screenshot 2024-10-09 at 7 13 07 PM


cardyok (approved):
nice, lgtm

PR #109, opened 2024-10-09
fix(k8s/pod): "gpud run --kubelet-ignore-connection-errors" to not mark unhealthy when read only port is not open (does not ignore by default)

gyuho:
Fix

Screenshot 2024-10-09 at 7 42 28 PM

### DIFF
--- a/components/k8s/pod/component_output.go
+++ b/components/k8s/pod/component_output.go
@@ -23,6 +23,7 @@ type Output struct {
 	NodeName string      `json:"node_name,omitempty"`
 	Pods     []PodStatus `json:"pods,omitempty"`
 
+	ConnectionError        string `json:"connection_error,omitempty"`
 	ConnectionErrorIgnored bool   `json:"connection_error_ignored"`
 	Message                string `json:"message,omitempty"`
 }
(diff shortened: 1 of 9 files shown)


components/k8s/pod/component_output.go
```diff
@@ -117,6 +122,20 @@ func CreateGet(cfg Config) query.GetFunc {
...
+			}
+			if cfg.IgnoreConnectionErrors && connectionErr {
+				log.Logger.Warnw("failed to list pods from kubelet read-only port but ignoring since ignore_connection_errors is true", "error", err)
+				return &Output{
+					ConnectionError:        err.Error(),
+					ConnectionErrorIgnored: true,
+					Message:                "failed to list pods from kubelet read-only port but ignoring (maybe readOnlyPort not set in kubelet config file)",
+				}, nil
+			}
 			return nil, err
```
cardyok, line 139: return output, err instead of nil, err here?
we are ignoring errors for now but it might be unexpected if we start handling errors in the future..

components/k8s/pod/component_output.go
```diff
@@ -117,6 +122,20 @@ func CreateGet(cfg Config) query.GetFunc {
...
+				connectionErr = true
+			}
+			if cfg.IgnoreConnectionErrors && connectionErr {
+				log.Logger.Warnw("failed to list pods from kubelet read-only port but ignoring since ignore_connection_errors is true", "error", err)
+				return &Output{
+					ConnectionError:        err.Error(),
+					ConnectionErrorIgnored: true,
+					Message:                "failed to list pods from kubelet read-only port but ignoring (maybe readOnlyPort not set in kubelet config file)",
+				}, nil
+			}
```
cardyok, line 138: when ignoring connections errors, we can check pidof kubelet instead to verify kubelet is running or not. If its not even running we can still report unhealthy

cardyok (approved):

PR #110, opened 2024-10-09
fix(docker-container): "gpud run --docker-ignore-connection-errors" to ignorer docker daemon connection errors (do not ignore by default)

gyuho:
c.f., https://github.com/leptonai/gpud/pull/109

Fix

Screenshot 2024-10-09 at 7 46 47 PM


cardyok:
similar to kubelet, when uds not available, we can check pidof docker instead.

cardyok (approved):

PR #111, opened 2024-10-09
feat(nvidia/info): report GPU device count from "/dev" (`DCGM_FR_DEVICE_COUNT_MISMATCH` DCGM)

gyuho:
Implementing `DCGM_FR_DEVICE_COUNT_MISMATCH` logic.

Tested

> ⌛ scanning nvidia accelerators
✔ successfully checked nvidia-smi
✔ GPU device count '8' (from /dev)
✔ GPU count '8' (from nvidia-smi)
✔ GPU count '8' (from NVML)
✔ GPU product name 'NVIDIA H100 80GB HBM3' (from nvidia-smi)
✔ GPU product name 'NVIDIA H100 80GB HBM3' (from NVML)

### DIFF
--- a/components/accelerator/nvidia/query/query.go
+++ b/components/accelerator/nvidia/query/query.go
@@ -62,6 +62,11 @@ func Get(ctx context.Context) (output any, err error) {
 		IbstatExists:          IbstatExists(),
 	}
 
+	o.GPUDeviceCount, err = CountAllDevicesFromDevDir()
+	if err != nil {
+		log.Logger.Warnw("failed to count gpu devices", "error", err)
+	}
+
 	defer func() {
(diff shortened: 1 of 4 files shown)


components/accelerator/nvidia/query/query.go
```diff
@@ -355,6 +383,26 @@ func (o *Output) GPUProductName() string {
 	return ""
 }
 
+func (o *Output) GPUProductNameFromSMI() string {
```
photoszzt, line 386: From this page: https://developer.nvidia.com/management-library-nvml, my interpretation is nvidia-smi uses nvml so I think the ProductName and GPU count won't differ.

gyuho: Removed.

PR #113, opened 2024-10-10
fix(fabric manager, nccl): fix fabric manager regex, add NCCL monitoring using dmesg

gyuho:
To start monitoring NCCL issue like:

> [Thu Oct 10 03:06:53 2024] pt_main_thread[2536443]: segfault at 7f797fe00000 ip 00007f7c7ac69996 sp 00007f7c12fd7c30 error 4 in libnccl.so.2[7f7c7ac00000+d3d3000]

### DIFF
--- a/components/accelerator/nvidia/query/fabric-manager-log/log.go
+++ b/components/accelerator/nvidia/query/fabric-manager-log/log.go
@@ -1,0 +1,15 @@
+package fabricmanagerlog
+
+const (
+	// e.g.,
+	// [Jul 23 2024 07:53:55] [ERROR] [tid 841] detected NVSwitch fatal error 20034 on fid 0 on NVSwitch pci bus id 00000000:86:00.0 physical id 3 port 33
+	RegexNVSwitchFatalSXidFromLog = `.+detected NVSwitch fatal error (\d+)`
+
+	// e.g.,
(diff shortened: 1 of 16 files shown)


components/accelerator/nvidia/query/fabric-manager-log/log.go
```diff
@@ -0,0 +1,15 @@
+package fabricmanagerlog
+
+const (
+	// e.g.,
+	// [Jul 23 2024 07:53:55] [ERROR] [tid 841] detected NVSwitch fatal error 20034 on fid 0 on NVSwitch pci bus id 00000000:86:00.0 physical id 3 port 33
+	RegexNVSwitchFatalSXidFromLog = `.+detected NVSwitch fatal error (\d+)`
```
photoszzt, line 6: When will the Regex got compiled? Since these are fixed string, could we use regexp.MustCompile to compile the regexp once instead of compiling multiple times at runtime?

gyuho: We pass these to the poller config and compile once here:

https://github.com/leptonai/gpud/blob/c60a71308be289704d110a42fd10d247a39abe63/components/query/log/filter/filter.go#L54-L73

photoszzt (approved):
LGTM.

PR #114, opened 2024-10-11
feat(nvidia/nvml): update nvlib to 0.7.0, rename device ID fields

gyuho:
c.f., https://github.com/NVIDIA/go-nvlib/pull/44

### DIFF
--- a/components/accelerator/nvidia/query/query.go
+++ b/components/accelerator/nvidia/query/query.go
@@ -356,10 +356,9 @@ func (o *Output) GPUProductName() string {
 }
 
 const (
-	inProgress       = "\033[33m⌛\033[0m"
-	checkMark        = "\033[32m✔\033[0m"
-	warningSign      = "\033[31m✘\033[0m"
-	warningBlackSign = "\033[30m✘\033[0m"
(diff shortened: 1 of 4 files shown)


components/accelerator/nvidia/query/query.go
```diff
@@ -356,9 +356,10 @@ func (o *Output) GPUProductName() string {
 }
 
 const (
-	inProgress  = "\033[33m⌛\033[0m"
-	checkMark   = "\033[32m✔\033[0m"
-	warningSign = "\033[31m✘\033[0m"
+	inProgress       = "\033[33m⌛\033[0m"
+	checkMark        = "\033[32m✔\033[0m"
+	warningSign      = "\033[31m✘\033[0m"
+	warningBlackSign = "\033[30m✘\033[0m"
```
photoszzt, line 362: Is this const used in other place?

photoszzt (approved):
LGTM. Is this PR for some other commits? I don't see a place that uses warningBlackSign.

components/accelerator/nvidia/query/query.go
gyuho: good catch... let me remove

PR #116, opened 2024-10-12
fix(session): close writer goroutine

cardyok:
revert #7 
still need to close write channel, when control plane is restarting and 504 is thrown, http is disconnected and goroutine will be leak. There will be multiple consumers of s.writer and the old goroutine will keep consuming the data to be written back to control plane and control plane will never get response back.
image


gyuho (approved):
lgtm

PR #119, opened 2024-10-14
feat(nvidia): inspect process zombie status, bad env vars for CUDA per process (`DCGM_FR_BAD_CUDA_ENV`)

gyuho:


photoszzt (approved):
LGTM. For the bad environments, is this a warning also needed to display to user on UI?

gyuho:
@photoszzt Yes, let me surface them in the components

gyuho:
Will create a new PR for global env checks

Issue #103, opened 2024-10-08
Running gpud as a daemonset/pod in an EKS cluster


chatter92:
Hi,

I am trying to run gpud as a privileged pod in an EKS cluster by creating a docker image out of it. Here is my docker image:

```
FROM nvidia/cuda:12.6.1-cudnn-devel-ubuntu24.04

#For gpu visibility
ENV NVIDIA_VISIBLE_DEVICES=all

RUN apt-get update && apt-get install -y curl

RUN mkdir /tmp/test-gpud
RUN curl -L https://github.com/leptonai/gpud/releases/download/v0.0.4/gpud_v0.0.4_linux_amd64_ubuntu24.04.tgz > /tmp/gpud_v0.0.4_linux_amd64_ubuntu24.04.tgz
RUN tar xzf /tmp/gpud_v0.0.4_linux_amd64_ubuntu24.04.tgz -C /tmp/test-gpud
RUN cp -f /tmp/test-gpud/gpud /usr/sbin

EXPOSE 15132

CMD /usr/sbin/gpud run
```

The pod gets created successfully and I can see gpud running in the pod when I exec into it. However, it starts erroring out after sometime and repeatedly keeps restarting because of a panic in the poller.

Here are the error logs I retrieved from a failed pod:

```
{"level":"info","ts":"2024-10-08T12:50:03Z","caller":"config/default.go:186","msg":"auto-detected clock events not supported -- skipping","error":null}
{"level":"info","ts":"2024-10-08T12:50:03Z","caller":"config/default.go:212","msg":"auto-detected gpm not supported -- skipping","error":null}
{"level":"info","ts":"2024-10-08T12:50:03Z","caller":"command/run.go:84","msg":"starting gpud v0.0.4"}
{"level":"info","ts":"2024-10-08T12:50:03Z","caller":"server/server.go:131","msg":"api version","version":"v1"}
{"level":"warn","ts":"2024-10-08T12:50:03Z","caller":"nvml/nvml.go:326","msg":"gpm metrics not supported"}
2024/10/08 12:50:03 Waiting for /var/log/fabricmanager.log to appear...
{"level":"info","ts":"2024-10-08T12:50:55Z","caller":"server/server.go:1008","msg":"serving 0.0.0.0:15132"}
{"level":"info","ts":"2024-10-08T12:50:55Z","caller":"command/run.go:105","msg":"successfully booted","tookSeconds":52.100623932}



✔ serving https://0.0.0.0:15132


[...]

PR #122, opened 2024-10-15
feat(gpud): "gpud run --auto-update-exit-code" for daemon set auto update use case (optional)

gyuho:
Address https://github.com/leptonai/gpud/issues/103.

### DIFF
--- a/internal/session/serve.go
+++ b/internal/session/serve.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
+	"os"
 	"time"
 
 	v1 "github.com/leptonai/gpud/api/v1"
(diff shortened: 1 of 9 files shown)


internal/session/serve.go
```diff
@@ -71,32 +73,51 @@ func (s *Session) serve() {
...
 
 		responseRaw, _ := json.Marshal(response)
 		s.writer <- Body{
 			Data:  responseRaw,
 			ReqID: body.ReqID,
 		}
+
+		if needExit != -1 {
+			log.Logger.Infow("exiting with code for auto update", "code", needExit)
+			os.Exit(needExit)
```
hm2501, line 119: Should this be `s.autoUpdateExitCode`?

gyuho: oh good catch

hm2501 (approved):
LGTM

PR #127, opened 2024-10-18
fix(nvidia): use NVML + lspci to detect NVIDIA GPUs (without running nvidia-smi)

gyuho:


components/accelerator/nvidia/query/detect.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+// Returns true if the local machine has NVIDIA GPUs installed.
+func GPUsInstalled(ctx context.Context) (bool, error) {
+	smiInstalled := SMIExists()
+	if !smiInstalled {
+		return false, nil
+	}
+	log.Logger.Info("nvidia-smi installed")
```
xiang90, line 24: should be debug level?

components/accelerator/nvidia/query/detect.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	log.Logger.Infow("nvidia PCI devices found", "devices", len(pciDevices))
+
+	// now that we have the NVIDIA PCI devices,
+	// call NVML C-based API for NVML API
+	productName, err := LoadProductName(ctx)
+	if err != nil {
+		return false, err
+	}
+	log.Logger.Infow("detected nvidia gpu", "product", productName)
```
xiang90, line 43: productName can also be a network card. do we check the type of the device?

gyuho: It's calling the device.GetName, so changed this logging to GPU device name.

components/accelerator/nvidia/query/detect.go
gyuho: Done

components/accelerator/nvidia/query/detect.go
```diff
@@ -0,0 +1,145 @@
...
+func GPUsInstalled(ctx context.Context) (bool, error) {
+	smiInstalled := SMIExists()
+	if !smiInstalled {
+		return false, nil
+	}
+	log.Logger.Debugw("nvidia-smi installed")
+
+	// now that nvidia-smi installed,
+	// check the NVIDIA GPU presence via PCI bus
+	pciDevices, err := ListPCIs(ctx)
```
xiang90, line 28: it is only useful for printing?

gyuho: Renamed to `ListNVIDIAPCIs`. It's only listing the devices with NVIDIA name in it, thus checking whether the host has NVIDIA devices.

cardyok (approved):

PR #142, opened 2024-10-30
feat(nvidia/error/xid): provide reason in JSON format with detailed error information

gyuho:


cardyok:
todo: need to resolve XidCriticalErrorMarkedByGPUd from poller and report non critical xids in Events interface

cardyok:
seems now dmesg output overrides nvml output, is this expected?

gyuho:
> seems now dmesg output overrides nvml output,

The other way, NVML takes the precedence

components/accelerator/nvidia/error/xid/reason.go
```diff
@@ -0,0 +1,43 @@
...
+	return json.Marshal(r)
+}
+
+// XidError represents an XID error in the reason.
+type XidError struct {
+	// DataSource is the source of the data.
+	DataSource string `json:"data_source"`
+
+	// RawEvent is the raw event from the nvml API.
+	RawEvent nvidia_query_nvml.XidEvent `json:"raw_event"`
```
cardyok, line 32: Maybe dont need to pass whole rawEvent back to control plane, would make body too big, can just extract several necessary fields to XidError struct, like
SuggestedActions *common.SuggestedActions `json:"suggested_actions,omitempty"`
Description            string `json:"description"`
DeviceUUID string `json:"device_uuid"`

cardyok:
Need to mark State as healthy if all xid are !XidCriticalErrorMarkedByGPUd
maybe skip !XidCriticalErrorMarkedByGPUd items in State result for now and add them to Event in the future would make logic in Evaluate easier..

cardyok (approved):
lgtm

PR #147, opened 2024-11-01
feat(components/os): count process counts per status (e.g., zombie, detached)

gyuho:


components/os/component_output.go
```diff
@@ -150,6 +166,15 @@ func ParseStatesToOutput(states ...components.State) (*Output, error) {
 			}
 			o.Uptimes = uptimes
 
+		case StateNameProcessCountsByStatus:
+			counts, err := ParseStateProcessCountsByStatus(state.ExtraInfo)
```
cardyok, line 170: its gonna be messy if we cumulate all possible process states...R,S,D,T,Z with suffix...maybe we should just count Zombie process count in extra info and return unhealthy if exceed threshold..

cardyok (approved):

PR #145, opened 2024-10-31
feat(nvidia/xid,sxid): more accurate criticality, suggested actions by GPUd, catch all Xids

gyuho:

PR #151, opened 2024-11-04
nits(accelerator/nvidia): rename reason fields, Xid/SXid detail fields for clarification

gyuho:
Split some changes from https://github.com/leptonai/gpud/pull/145.


cardyok (approved):

cardyok:
lgtm, but lets wait for everything in #145 merged before releasing 0.1.3 so that control plane can be consistent

PR #161, opened 2024-11-07
feat(nvidia/remapped-rows): suggest reboot/inspection on row remapping

gyuho:


cardyok (approved):

gyuho:
@cardyok Let's keep them separate for now. In case nvidia-smi parsing fails, we need some fallback, and still need to know when nvidia-smi parsing fails.

PR #162, opened 2024-11-07
feat(nvidia/gsp-firmware-mode): initial commit to track GSP modes

gyuho:


components/accelerator/nvidia/gsp-firmware-mode/component.go
```diff
@@ -0,0 +1,99 @@
...
+func (c *component) Name() string { return nvidia_gsp_firmware_mode_id.Name }
+
+func (c *component) States(ctx context.Context) ([]components.State, error) {
+	last, err := c.poller.Last()
+	if err == query.ErrNoData { // no data
+		log.Logger.Debugw("nothing found in last state (no data collected yet)", "component", nvidia_gsp_firmware_mode_id.Name)
+		return []components.State{
+			{
+				Name:    nvidia_gsp_firmware_mode_id.Name,
+				Healthy: false,
```
cardyok, line 46: this shoud be true...actually, this should be true for all components..

PR #174, opened 2024-11-10
feat(pkg/dmesg): use common ctime parser

gyuho:
Break https://github.com/leptonai/gpud/pull/157 into smaller PRs.


gyuho:
> Why do we need a version that doesn't return error?

Let me check if I need it in other PRs. If not, I will remove :)

PR #187, opened 2024-11-15
fix(nvidia/xid-sxid-state): persist xid/sxid in tail scan, better logging

gyuho:


cardyok (approved):
logs should be debug level, the rest lgtm

PR #196, opened 2024-11-21
feat(nvidia/infiniband): make port states configurable

gyuho:


config/op_options.go
```diff
@@ -22,6 +25,12 @@ func WithFilesToCheck(files ...string) OpOption {
 	}
 }
 
+func WithExpectedPortStates(exp infiniband.ExpectedPortStates) OpOption {
```
cardyok, line 28: so we are not using this yet right?

gyuho: Not yet

cardyok (approved):
lgtm after conflict resolved

PR #205, opened 2024-11-23
fix(session): disable http keep alive

cardyok:


internal/session/session.go
```diff
@@ -168,22 +167,7 @@ func (s *Session) startWriter() {
...
-				}).DialContext,
-				ForceAttemptHTTP2:     true,
-				MaxIdleConns:          10,
-				IdleConnTimeout:       30 * time.Second,
-				TLSHandshakeTimeout:   10 * time.Second,
-				ExpectContinueTimeout: 1 * time.Second,
+				DisableKeepAlives: true,
 			},
 		}
 		resp, err := client.Do(req)
```
xiang90, line 173: is there a timeout on this do func?

internal/session/session.go
```diff
@@ -251,22 +235,7 @@ func (s *Session) startReader() {
...
```
xiang90, line 241: it seems there is no timeout here as well...

internal/session/session.go
xiang90: writecloser needs to somehow get passed in here.

xiang90:
basically:

1. create a http client every single time... do not reuse any http client.
2. on the reader side, it reads every x seconds. if there is nothing, then the read side of the tcp connection may be broken. it should close both the reader and the writer.
3. on the writer side, it writes every x seconds. if it fails to write, then the write side of the tcp connection may be broken. it should close both the reader and the writer.

xiang90:
also do extensive logging everywhere.

gyuho (approved):
lgtm

PR #214, opened 2024-11-29
fix(session): close reader channel on fast return

cardyok:


internal/session/session.go
```diff
@@ -252,6 +252,7 @@ func (s *Session) startReader(readerExit chan any) {
 	req, err := http.NewRequestWithContext(s.ctx, "POST", s.endpoint, nil)
 	if err != nil {
 		log.Logger.Debugf("session reader: error creating request: %v", err)
+		close(pipeFinishCh)
```
xiang90, line 255: shall we just put this into the defer? is it easy to add a test?

PR #217, opened 2024-11-30
feat(nvidia/xid-sxid): increase xid/sxid table retention period to 3-hour

gyuho:
To help us debug better


components/accelerator/nvidia/query/xid-sxid-state/state.go
```diff
@@ -231,6 +234,7 @@ FROM %s`,
 }
 
 func Purge(ctx context.Context, db *sql.DB, opts ...OpOption) (int, error) {
+	log.Logger.Infow("purging nvidia xid/sxid events")
```
cardyok, line 237: lgtm, prob debug level

cardyok (approved):

PR #221, opened 2024-11-30
fix(nvidia/query): only evaluate memory error management capabilities when product name found, add missing GPU ID in nvidia-smi parsing remapped rows

gyuho:


cardyok:
maybe we should report healthy if device does not support row remapping?

cardyok (approved):

PR #224, opened 2024-12-02
feat(fd): monitor VFS file-max limit with allocated file handles on Linux

gyuho:
If this limit is reached, the process halts, impacting the whole host:

> [Sun Dec  1 14:54:40 2024] VFS: file-max limit 1000000 reached

and

> sudo: unable to open /etc/sudoers: Too many open files in system
sudo: no valid sudoers sources found, quitting
sudo: error initializing audit plugin sudoers_audit


components/dmesg/filters_fd.go
```diff
@@ -0,0 +1,28 @@
...
+	"k8s.io/utils/ptr"
+)
+
+const (
+	// e.g.,
+	// [...] VFS: file-max limit 1000000 reached
+	//
+	// ref.
+	// https://docs.kernel.org/admin-guide/sysctl/fs.html#file-max-file-nr
+	EventFileDescriptorVFSFileMaxLimitReached = "nvidia_file_descriptor_vfs_file_max_limit_reached"
```
cardyok, line 17: prob not nvidia_

gyuho: good catch... fixed :)

components/fd/component_output.go
```diff
@@ -116,37 +149,54 @@ func (o *Output) States() ([]components.State, error) {
...
-			state.Reason += "-- threshold_used_percent is greater than 95"
-		}
+	if o.FDLimitSupported && o.ThresholdRunningPIDs > 0 && o.RunningPIDs > o.ThresholdRunningPIDs {
+		state.Healthy = false
+		state.Reason += "; running_pids is greater than threshold_running_pids"
+	}
+	if o.FileHandlesSupported && o.ThresholdAllocatedFileHandles > 0 && o.AllocatedFileHandles > o.ThresholdAllocatedFileHandles {
+		state.Healthy = false
+		state.Reason += "; allocated_file_handles is greater than threshold_allocated_file_handles"
 	}
```
cardyok, line 195: here maybe we should check ThresholdAllocatedFileHandlesPercent as well (also give a default value like 80%), imo ThresholdAllocatedFileHandlesPercent is more important than the absolute value of fd count..

cardyok (approved):

PR #226, opened 2024-12-03
feat(server): send components in gossip

cardyok:


internal/server/server.go
```diff
@@ -1303,7 +1303,10 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID
 		}
 	}()
 
-	if err = login.Gossip(endpoint, uid, config.Address); err != nil {
+	ghler.componentNamesMu.RLock()
+	componentNames := ghler.componentNames
+	ghler.componentNamesMu.RUnlock()
```
cardyok, line 1308: @gyuho ptal, not sure if this is the best way to get most up to date component names..

gyuho (approved):
lgtm on CI fixes

internal/server/server.go
gyuho: Correct. This is the latest set of components that are registers:

```go
				ghler.componentNamesMu.Lock()
				ghler.componentNames = newComponentNames
				ghler.componentNamesMu.Unlock()
			}
```

PR #229, opened 2024-12-04
feat(nvidia/hw-slowdown): include GPU UUID in /events, persist smi for /events, dedup hw slowdown events by data source and nearest minutes, do not return hw slowdown clock events in /states, fix nvidia query get function context timeout

gyuho:


cardyok (approved):
lgtm

PR #230, opened 2024-12-04
feat(nvidia/infiniband): better ib ports/rate checking based on port physical/state

gyuho:


components/accelerator/nvidia/query/infiniband/ibstat.go
```diff
@@ -39,6 +99,40 @@ func (cards IBStatCards) CountByRates(rate int, expectedState string, expectedPh
...
+		return nil
+	}
+
+	// some devices are missing -- less than expected ports found
+	// count all ports regardless of the physical states, states
+	portsWithHighEnoughRates := cards.Count("", "", atLeastRate)
+	if portsWithHighEnoughRates >= atLeastPorts {
+		// some ports must be down -- construct error message accordingly
+		return fmt.Errorf(
+			"not enough LinkUp ports, some ports must be down; only %d LinkUp out of %d, expected at least %d ports and %d Gb/sec rate",
```
cardyok, line 118: need to include IBStatPort.Name in the alert for both case, rest lgtm

cardyok (approved):

PR #232, opened 2024-12-05
feat(components/disk): track total mounted ext partitions

gyuho:


xiang90:
seems reasonable for disk checking. but then we need to separate file system and disk, which is what we should do eventually anyway.

cardyok:
tested, lgtm.
can we also find a place to store mounted device of /var/lib/kubelet as well? A new state name in disk component or some field in a already exist state both seems ok
value can be just simply `findmnt /var/lib/kubelet` and extract the device name in GetFunc

gyuho:
@cardyok Now we have

```json
      },
      {
        "name": "mount_target_usages",
        "healthy": true,
        "extra_info": {
          "data": "{\"/var/lib/kubelet\":{\"device\":\"/dev/mapper/vgroot-lvroot\",\"usage\":{\"total_bytes\":1888564924416,\"total_humanized\":\"1.9 TB\",\"free_bytes\":1304401465344,\"free_humanized\":\"1.3 TB\",\"used_bytes\":497757810688,\"
used_humanized\":\"498 GB\",\"used_percent\":\"27.62\",\"inodes_total\":117178368,\"inodes_used\":2140302,\"inodes_free\":115038066,\"inodes_used_percent\":\"1.83\"}}}",
          "encoding": "json"
        }
      }
    ]
  },
```

cardyok (approved):

PR #242, opened 2024-12-11
feat(components/dmesg): catch EDAC correctable errorrs in dmesg

gyuho:
ref. https://github.com/Azure/azurehpc/blob/2d57191cb35ed638525ba9424cc2aa1b5abe1c05/experimental/aks_npd_draino/npd/deployment/node-problem-detector-config.yaml#L51C20-L51C40


components/memory/dmesg/dmesg.go
```diff
@@ -0,0 +1,87 @@
...
+	// May indicate that Dual Inline Memory Module (DIMM) is beginning to fail.
+	//
+	// e.g.,
+	// [...] EDAC MC0: 1 CE memory read error
+	// [...] EDAC MC1: 128 CE memory read error on CPU_SrcID#1_Ha#0_Chan#1_DIMM#1
+	//
+	// ref.
+	// https://serverfault.com/questions/682909/how-to-find-faulty-memory-module-from-mce-message
+	// https://github.com/Azure/azurehpc/blob/2d57191cb35ed638525ba9424cc2aa1b5abe1c05/experimental/aks_npd_draino/npd/deployment/node-problem-detector-config.yaml#L51C20-L51C40
+	RegexEDACCorrectableErrors = `EDAC.*CE memory read error.*`
```
cardyok, line 41: seems azure regex does not have EDAC prefix..i think we can keep current way and iterate on it tho:)

cardyok (approved):

components/memory/dmesg/dmesg.go
gyuho: Removed

PR #245, opened 2024-12-12
feat(pci): move /states to /events for acs srv-valid checks

gyuho:


components/pci/component_output.go
```diff
@@ -86,18 +88,21 @@ func (o *Output) States() ([]components.State, error) {
...
+
+	// no PCI device with Access Control Services (ACS) enabled or running in VM thus ACS is not an option
+	if len(acsReasons) == 0 {
+		return nil, nil
 	}
 
-	return []components.State{
+	// polling happens periodically
+	// so we truncate the timestamp to the nearest minute
+	truncNowUTC := time.Now().UTC().Truncate(time.Minute)
```
cardyok, line 99: if we truncate to current timestamp, event will be sent repeatedly, we should send it only once for now, maybe record it in db after sent once..

cardyok (approved):

PR #250, opened 2024-12-13
fix(os): run machine/boot id get calls only for linux, gpud run exit 1 on non-linux platform

gyuho:


xiang90:
let us not even worry about other OS/platform for now. and let us drop this PR since we do not want to make things complicated.

we should add a simple check in the main() to exit gpud if the os is not linux.

gyuho:
@xiang90 Now `gpud run` exits 1 on non-linux platform. Still need this change since I am signing the release using gpud sign on Mac :)

xiang90:
i do not get it… why do we still need it?

gyuho:
@xiang90 

- `gpud sign` -- run on mac by the release manager
- `dmidecode` -- currently running on every platform since we added a code with `init` function at the package level
- `dmidecode` -- not needed for mac, but this PR disables that call path for non-linux
- plus, this PR adds an exit code for `gpud run` so it cannot join the platform

components/os/component_output.go
```diff
@@ -362,31 +363,35 @@ type MachineMetadata struct {
...
 
 func init() {
-	// e.g., "/proc/sys/fs/file-max" exists on linux
-	if file.CheckFDLimitSupported() {
-		limit, err := file.GetLimit()
-		if limit > 0 && err == nil {
-			// set to 20% of system limit
-			DefaultZombieProcessCountThreshold = int(float64(limit) * 0.20)
+	// Linux-specific operations
+	if runtime.GOOS == "linux" {
```
xiang90, line 367: if runtime.GOOS != linux (typedef this) {
   return
}

gyuho: done

components/os/component_output.go
```diff
@@ -362,7 +363,12 @@ type MachineMetadata struct {
 var currentMachineMetadata MachineMetadata
 
 func init() {
-	// e.g., "/proc/sys/fs/file-max" exists on linux
+	// Linux-specific operations
+	if runtime.GOOS != "linux" {
```
xiang90, line 367: is linux type defined somewhere?

xiang90 (approved):

components/os/component_output.go
gyuho: Only defined as a build constant :/

```go
// Code generated by gengoos.go using 'go generate'. DO NOT EDIT.

//go:build !android && linux

package goos

const GOOS = `linux`
```

PR #256, opened 2024-12-17
fix(disk): add retries for lsblk

gyuho:


components/disk/component_output.go
```diff
@@ -33,90 +34,114 @@ const (
...
+				StateKeyData:     string(diskBlockDevicesData),
+				StateKeyEncoding: StateValueEncodingJSON,
+			},
 		},
 		{
-			Name:    StateNameDiskExtPartition,
-			Healthy: true,
-			Reason:  "",
+			Name:    StateNameDiskBlockDevicesTotal,
+			Healthy: len(o.DiskBlockDevices) > 0,
```
cardyok, line 112: maybe just return true for all the disk states for now..noticed findmnt also flaky..

cardyok (approved):
lgtm after comment resolved

components/disk/component_output.go
gyuho: Setting true where we use `MountTargetsToTrackUsage` (using findmnt)

PR #260, opened 2024-12-20
fix(query/log/tail): fix time parser for initial lines, use correct time for fabric manager /events

gyuho:
The "tail" library emits the existing lines in the file, when we spawn the tail-er on a file, thus requires custom log line time parser function (e.g., fabric manager log)

Fix where the following "old" logs show up in our fabric manager component events:

> {"fabricmanager_nvswitch_log_error":"","fabricmanager_nvswitch_log_filter":"{\"name\":\"accelerator-nvidia-fabric-manager-nvlink-failure\",\"regex\":\".+failed to find the GPU handle \\\\d+ in the multicast team .*\",\"owner_references\":[\"accelerator-nvidia-fabric-manager\"]}","fabricmanager_nvswitch_log_line":"[Apr 17 2024 01:51:39] [ERROR] [tid 2999877] failed to find the GPU handle 10187860174420860981 in the multicast team request setup 5653964288847403984.","fabricmanager_nvswitch_log_unix_seconds":"1734654883"}

### DIFF
--- a/components/query/log/tail/streamer_file.go
+++ b/components/query/log/tail/streamer_file.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"time"
 
+	query_log_common "github.com/leptonai/gpud/components/query/log/common"
 	"github.com/leptonai/gpud/log"
+
 	"github.com/nxadm/tail"
 )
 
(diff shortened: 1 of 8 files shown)


components/query/log/tail/streamer_file.go
```diff
@@ -86,21 +93,38 @@ func (sr *fileStreamer) pollLoops() {
...
 		}
 
-		if line.Time.IsZero() {
-			line.Time = time.Now().UTC()
+		if sr.extractTime != nil {
+			parsedTime, _, err := sr.extractTime([]byte(txt))
+			if err == nil {
+				line.Time = parsedTime
+			} else {
+				log.Logger.Warnw("error extracting time", "error", err)
```
cardyok, line 118: shall we just skip this line if time parse fails?

cardyok (approved):

PR #261, opened 2024-12-20
feat(nvidia/query): bump up nvidia-smi cmd timeout, better debugging info

gyuho:
To better debug the case where nvidia-smi hangs


cardyok:
lgtm, we can fix the ut and do the poller cache in separate pr

cardyok (approved):

PR #264, opened 2024-12-23
feat(pkg/process): label process for better debugging info

gyuho:


cardyok (approved):
lgtm, lets continue debugging the flodding channel full problem after this release

PR #265, opened 2024-12-25
feat(nvidia): use last successful data in shared poller, shared nvidia-smi/nvml poller to still return data if one operation fails

gyuho:
If one operation fails, the shared poller should still return the data so that each component that consumes the shared data can evaluate their own healthiness.

### DIFF
--- a/components/accelerator/nvidia/bad-envs/component.go
+++ b/components/accelerator/nvidia/bad-envs/component.go
@@ -53,27 +53,13 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 	if err != nil {
 		return nil, err
 	}
-	if last.Error != nil {
-		return []components.State{
-			{
-				Healthy: false,
-				Error:   last.Error.Error(),
-				Reason:  "last query failed",
-			},
-		}, nil
-	}
-	if last.Output == nil {
-		return []components.State{
-			{
-				Healthy: false,
-				Reason:  "no output",
-			},
-		}, nil
-	}
 
 	allOutput, ok := last.Output.(*nvidia_query.Output)
 	if !ok {
 		return nil, fmt.Errorf("invalid output type: %T", last.Output)
+	}
+	if lerr := c.poller.LastError(); lerr != nil {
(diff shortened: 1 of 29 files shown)


cardyok (approved):
lgtm, maybe we should consider making poll channel larger (>1) in case something happens to processItem and block forever

xiang90:
even the failed one SHOULD return the previously cached data without failing its downstream. and then we report the poller is failing. let us do this for now

gyuho:
> making poll channel larger (>1) in case something happens to processItem and block forever

Should be fine for now, buffer=1, receive in for-loop:

https://github.com/leptonai/gpud/blob/a7cdf53da3fb277bfa0fe21518211a7d9f2befa4/components/query/poller.go#L96-L98

https://github.com/leptonai/gpud/blob/a7cdf53da3fb277bfa0fe21518211a7d9f2befa4/components/query/poller.go#L192-L197

And the default query queue size is 60 (meaning we go back up to 5 hours)

https://github.com/leptonai/gpud/blob/a7cdf53da3fb277bfa0fe21518211a7d9f2befa4/components/query/config/config.go#L13-L19

> even the failed one SHOULD return the previously cached data without failing its downstream

Yes, we are now using this

```go
// Reads the last item from the queue.
// If requireNoErr is true, it returns the last item with no error.
// If no item is found, it returns ErrNoData.
func (pl *poller) readLast(requireNoErr bool) (*Item, error) {
	pl.lastItemsMu.RLock()
	defer pl.lastItemsMu.RUnlock()

	if len(pl.lastItems) == 0 {
		return nil, ErrNoData
	}

	// reverse iterate
	for i := len(pl.lastItems) - 1; i >= 0; i-- {
		item := pl.lastItems[i]
		if requireNoErr && item.Error != nil {
			log.Logger.Warnw("skipping item due to error", "id", pl.id, "error", item.Error)
			continue
		}
		return &item, nil
	}

	return nil, ErrNoData
}
```

where each item has its own timestamp to detect when it's too stale.

PR #280, opened 2025-01-02
fix(systemd): set shorter context timeout for dbus calls

gyuho:


cardyok (approved):
lgtm, we can log the err

PR #281, opened 2025-01-02
feat(state): separate read-only sqlite instance for better concurrency

gyuho:
c.f., https://github.com/mattn/go-sqlite3/issues/1179#issuecomment-1638083995


cardyok (approved):
lgtm, for GetUID, we can use dbRO only, prob need a new state.GetMachineID func and return err on not found

PR #287, opened 2025-01-06
feat(cpu,memory): return hung task, soft lockup, oom from dmesg via /events, fix log item error type to "*string"

gyuho:


xiang90:
Need UTs.

components/dmesg/component_event.go
```diff
@@ -28,9 +29,11 @@ func ParseEventJSON(data []byte) (*Event, error) {
 const (
 	EventNameDmesgMatched = "dmesg_matched"
 
-	EventKeyDmesgMatchedError = "error"
+	EventKeyDmesgMatchedError   = "error"
+	EventKeyDmesgMatchedLogItem = "log_item"
 )
 
+// TODO: deprecate
```
gyuho, line 36: @cardyok We need to deprecate dmesg component, and just use it as a shared poller.

cardyok: make sense to me.
Or we can keep the component (only report state) representing health stage of the poller

components/dmesg/component_event.go
```diff
@@ -45,7 +48,13 @@ func (ev *Event) Events() []components.Event {
 
 		es := ""
 		if logItem.Error != nil {
-			es = logItem.Error.Error()
+			es = *logItem.Error
```
cardyok, line 51: need nil check

components/accelerator/nvidia/fabric-manager/component.go
```diff
@@ -130,7 +130,7 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
 		b, _ := ev.Matched.JSON()
 		es := ""
 		if ev.Error != nil {
-			es = ev.Error.Error()
+			es = *ev.Error
```
cardyok, line 133: nil check

components/dmesg/component_state.go
```diff
@@ -138,7 +138,7 @@ func (s *State) States() []components.State {
 			b, _ := item.Matched.JSON()
 			es := ""
 			if item.Error != nil {
-				es = item.Error.Error()
+				es = *item.Error
```
cardyok, line 141: nil check

cardyok (approved):
lgtm

components/dmesg/component_event.go
gyuho: > if logItem.Error != nil {

already does :)

components/accelerator/nvidia/fabric-manager/component.go
gyuho: > if ev.Error != nil {

already does :)

components/dmesg/component_state.go
gyuho: > if item.Error != nil {

already does :)

PR #288, opened 2025-01-06
feat(nvidia/peermem): explicitly skip "invalid context" errors

gyuho:
as the latest driver fixes the issue

ref. https://github.com/Mellanox/nv_peer_memory/issues/120

### DIFF
--- a/components/accelerator/nvidia/peermem/component.go
+++ b/components/accelerator/nvidia/peermem/component.go
@@ -116,6 +116,10 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
 		return nil, err
 	}
 
+	return c.getEvents(ctx, since, dmesgTailResults)
+}
+
+func (c *component) getEvents(ctx context.Context, since time.Time, dmesgTailResults *dmesg.State) ([]components.Event, error) {
 	// dedup by minute level
 	seenMinute := make(map[int64]struct{})
 	events := make([]components.Event, 0)
--- a/components/accelerator/nvidia/peermem/component.go
+++ b/components/accelerator/nvidia/peermem/component.go
@@ -11,7 +11,6 @@ import (
 	"github.com/leptonai/gpud/components"
 	nvidia_peermem_id "github.com/leptonai/gpud/components/accelerator/nvidia/peermem/id"
 	nvidia_query "github.com/leptonai/gpud/components/accelerator/nvidia/query"
-	nvidia_query_peermem "github.com/leptonai/gpud/components/accelerator/nvidia/query/peermem"
 	"github.com/leptonai/gpud/components/dmesg"
 	"github.com/leptonai/gpud/components/query"
 	"github.com/leptonai/gpud/log"
@@ -140,8 +139,7 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
 		//
 		// ref. https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-129-03/index.html
 		// ref. https://github.com/Mellanox/nv_peer_memory/issues/120
-		line := logItem.Line
-		if nvidia_query_peermem.HasInvalidContext(line) {
+		if logItem.Matched.Name == dmesg.EventNvidiaPeermemInvalidContext {
 			continue
 		}
 
--- a/components/accelerator/nvidia/peermem/component.go
+++ b/components/accelerator/nvidia/peermem/component.go
@@ -11,6 +11,7 @@ import (
 	"github.com/leptonai/gpud/components"
 	nvidia_peermem_id "github.com/leptonai/gpud/components/accelerator/nvidia/peermem/id"
 	nvidia_query "github.com/leptonai/gpud/components/accelerator/nvidia/query"
+	nvidia_query_peermem "github.com/leptonai/gpud/components/accelerator/nvidia/query/peermem"
 	"github.com/leptonai/gpud/components/dmesg"
 	"github.com/leptonai/gpud/components/query"
 	"github.com/leptonai/gpud/log"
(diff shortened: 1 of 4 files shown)


components/accelerator/nvidia/peermem/component.go
```diff
@@ -98,56 +96,7 @@ const (
 )
 
 func (c *component) Events(ctx context.Context, since time.Time) ([]components.Event, error) {
-	dmesgC, err := components.GetComponent(dmesg.Name)
```
xiang90, line 101: we still need to detect other peer mem related dmsg.

components/accelerator/nvidia/peermem/component.go
```diff
@@ -98,56 +96,7 @@ const (
...
-		if logItem.Matched.Name != dmesg.EventNvidiaPeermemInvalidContext {
-			continue
-		}
-
-		minute := logItem.Time.Unix() / 60
-		if _, ok := seenMinute[minute]; ok {
-			continue
-		}
-		seenMinute[minute] = struct{}{}
-
```
xiang90, line 138: ignore it here. 

if name == dmesg.EventNvidiaPeermemInvalidContext {
    continue
    // some comment here.
}

components/accelerator/nvidia/peermem/component.go
```diff
@@ -98,56 +96,7 @@ const (
...
-	seenMinute := make(map[int64]struct{})
-	events := make([]components.Event, 0)
-	for _, logItem := range dmesgTailResults.TailScanMatched {
-		if logItem.Error != nil {
-			continue
-		}
-		if logItem.Matched == nil {
-			continue
-		}
-		if logItem.Matched.Name != dmesg.EventNvidiaPeermemInvalidContext {
```
xiang90, line 129: each dmsg should have an affect component field. so we should compare that rather than the very specific msg type. we can improve this later.

components/dmesg/filters_nvidia.go
```diff
@@ -58,11 +49,6 @@ func DefaultDmesgFiltersForNvidia() []*query_log_common.Filter {
 			Regex:           ptr.To(nvidia_query_sxid.RegexNVSwitchSXidDmesg),
 			OwnerReferences: []string{nvidia_component_error_sxid_id.Name},
 		},
-		{
-			Name:            EventNvidiaPeermemInvalidContext,
```
xiang90, line 62: keep this.

components/accelerator/nvidia/peermem/component.go
gyuho: Yeah right now, using owner reference is better

components/accelerator/nvidia/peermem/component.go
```diff
@@ -126,7 +126,16 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
...
+
+		// skip this for now as the latest driver https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-560-35-03/index.html#abstract fixes this issue
+		// "4272659 – A design defect has been identified and mitigated in the GPU kernel-mode driver, related to the GPUDirect RDMA support
+		// in MLNX_OFED and some Ubuntu kernels, commonly referred to as the PeerDirect technology, i.e. the one using the peer-memory kernel
+		// patch. In specific scenarios, for example involving the cleanup after killing of a multi-process application, this issue may lead to
+		// use-after-free and potentially to kernel memory corruption."
+		//
+		// ref. https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-129-03/index.html
+		// ref. https://github.com/Mellanox/nv_peer_memory/issues/120
+		if logItem.Matched.Name == dmesg.EventNvidiaPeermemInvalidContext {
```
xiang90, line 138: actually we still need the previous != operator. or other dmsg will go through. put the additional == operator at the exact place i pointed out

components/accelerator/nvidia/peermem/component.go
```diff
@@ -126,10 +127,24 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
...
 
+		// skip this for now as the latest driver https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-560-35-03/index.html#abstract fixes this issue
+		// "4272659 – A design defect has been identified and mitigated in the GPU kernel-mode driver, related to the GPUDirect RDMA support
+		// in MLNX_OFED and some Ubuntu kernels, commonly referred to as the PeerDirect technology, i.e. the one using the peer-memory kernel
+		// patch. In specific scenarios, for example involving the cleanup after killing of a multi-process application, this issue may lead to
+		// use-after-free and potentially to kernel memory corruption."
+		//
+		// ref. https://docs.nvidia.com/datacenter/tesla/tesla-release-notes-535-129-03/index.html
+		// ref. https://github.com/Mellanox/nv_peer_memory/issues/120
+		line := logItem.Line
```
xiang90, line 143: just copy line 130 and change the != to == done. remove all complicated regex stuff.

xiang90: it is ok to repeat that line and the ONLY difference is the operator ( == )

components/accelerator/nvidia/query/peermem/dmesg.go
```diff
@@ -1,9 +1,22 @@
...
 
+import "regexp"
+
 // repeated messages may indicate more persistent issue on the inter-GPU communication
 // e.g.,
 // [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
 // [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
 // [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
 const RegexInvalidContext = `.*ERROR detected invalid context, skipping further processing`
+
```
xiang90, line 12: revert all these changes?

xiang90:
need a ut to make sure this error dmsg is skipped

xiang90:
lgtm. defer to @cardyok

cardyok (approved):
tested and lgtm

PR #291, opened 2025-01-09
fix(nvml): handle "not supported" error to not fail-fast for NVML get calls

gyuho:


components/accelerator/nvidia/query/nvml/gsp_firmware.go
```diff
@@ -22,8 +22,15 @@ func GetGSPFirmwareMode(uuid string, dev device.Device) (GSPFirmwareMode, error)
 	}
 
 	gspEnabled, supported, ret := dev.GetGspFirmwareMode()
+	if IsNotSupportError(ret) {
+		mode.Enabled = false
+		mode.Supported = false
+		return mode, nil
```
gyuho, line 28: @cardyok This is the fix for

image (1)

cardyok (approved):
lgtm, one question, in general,
seems changing ret != nvml.SUCCESS -> IsNotSupportError(ret) ignores errors other than not supported, will this cause unexpected behavior?

gyuho:
> ret != nvml.SUCCESS

This will fail since the ret is nvml NOT SUPPORTED from NVML call, we want to return no error if it's not supported

PR #293, opened 2025-01-13
feat(power-supply): monitor inefficient power on PCIe

gyuho:
c.f., https://github.com/leptonai/gpud/pull/290

### DIFF
--- a/components/power-supply/component.go
+++ b/components/power-supply/component.go
@@ -4,9 +4,12 @@ package powersupply
 import (
 	"context"
 	"fmt"
+	"runtime"
+	"strconv"
 	"time"
 
 	"github.com/leptonai/gpud/components"
+	common_dmesg "github.com/leptonai/gpud/components/common/dmesg"
 	power_supply_id "github.com/leptonai/gpud/components/power-supply/id"
 	"github.com/leptonai/gpud/components/query"
 	"github.com/leptonai/gpud/log"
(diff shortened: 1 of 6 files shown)


components/power-supply/component.go
```diff
@@ -79,7 +88,49 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
...
+			continue
+		}
+
+		events = append(events, components.Event{
+			Time:    logItem.Time,
+			Name:    common_dmesg.EventPowerSupplyInsufficientPowerOnPCIe,
+			Type:    components.EventTypeCritical,
+			Message: "Insufficient power on the PCIe slot",
+			ExtraInfo: map[string]string{
+				"unix_seconds": strconv.FormatInt(logItem.Time.Unix(), 10),
```
xiang90, line 127: should include the pice device uuid and possibly device name and vendor

components/power-supply/component.go
```diff
@@ -79,7 +88,49 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
...
+			continue
+		}
+		if logItem.Matched.Name != common_dmesg.EventPowerSupplyInsufficientPowerOnPCIe {
+			continue
+		}
+
+		events = append(events, components.Event{
+			Time:    logItem.Time,
+			Name:    common_dmesg.EventPowerSupplyInsufficientPowerOnPCIe,
+			Type:    components.EventTypeCritical,
```
xiang90, line 124: for all the newly introduced event, the level should be a warning. after the first release, we monitor it and if we feel good we promote it to critical in the next release.

components/power-supply/component.go
```diff
@@ -79,7 +88,49 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 }
 
 func (c *component) Events(ctx context.Context, since time.Time) ([]components.Event, error) {
```
xiang90, line 90: need ut for this method as well

gyuho:
Will revisit once we stabilize everything

PR #294, opened 2025-01-14
fix(*): add missing close on process that runs commands as bash, rename abort to close

gyuho:
Fix

> open /tmp/tmpbash888507873.bash: too many open files


xiang90:
any way we can test this?

gyuho:
@xiang90 @cardyok unit tested and end to end tested (confirmed that no temp file leaks for the last 24 hour)

xiang90:
lgtm. defer to @cardyok

cardyok (approved):
lgtm

PR #299, opened 2025-01-16
fix(tailscale): make version checks less flaky with retries

gyuho:


xiang90:
this is way too complicated than needed. what is the problem that we are trying to solve?

components/tailscale/component_output.go
```diff
@@ -122,12 +151,28 @@ func Get(ctx context.Context) (_ any, e error) {
 		}
 	}()
 
-	ver, err := CheckVersion()
-	if err != nil {
-		return nil, err
+	var ver *VersionInfo
+	for i := 0; i < 5; i++ {
```
xiang90, line 155: retry on version check makes very little sense... why would this fail... if this fails, let us just wait for the next run to succeed. version check failure should be just a warning anyway. we should check tailscale systemd status.

gyuho: @cardyok Can we just drop tailscale version check here?

https://github.com/leptonai/gpud/blob/ab4037114adede58f63624c82040471deb513dc6/manager/controllers/package_controller.go#L99-L107

already does tailscale version checks

cardyok: yes we can, actually package controller already checking status/version of kubelet/containerd/tailscale

xiang90:
do we check the systemd unit status?

xiang90:
ok. let us get rid of the version check first anyway. it is not useful. let us create a new pr to check tailscale status.

gyuho:
@xiang90 @cardyok Updated. And actually, we do check tailscale systemd status:

https://github.com/leptonai/gpud/blob/a7cdf53da3fb277bfa0fe21518211a7d9f2befa4/config/default.go#L189-L191

(assume, tailscaled is already running or enabled after gpud join)

cardyok (approved):

PR #300, opened 2025-01-16
fix(nvidia): increase hw slowdown event level to critical (from info)

gyuho:


gyuho:
Closing. We should only critical when this happens multiple times

PR #304, opened 2025-01-18
fix(nvidia/hw-slowdown): evaluate state based on clock events per-minute frequency for the last 10-minute

gyuho:
- remove unused component output code
- remove unused clock events channel
- warning level for hw slowdown event
- move "clock events" db inserts to nvml for simplification
- evaluate clock events component state based on the clock events frequency
- use persistent clock events record using sqlite
- improve logging


components/accelerator/nvidia/hw-slowdown/component.go
```diff
@@ -82,7 +142,7 @@ func (c *component) Events(ctx context.Context, since time.Time) ([]components.E
 		convertedEvents = append(convertedEvents, components.Event{
 			Time:    metav1.Time{Time: time.Unix(event.UnixSeconds, 0).UTC()},
 			Name:    EventNameHWSlowdown,
-			Type:    components.EventTypeInfo,
+			Type:    components.EventTypeWarning,
```
gyuho, line 145: @cardyok @xiang90 Now,

1. all hw slowdown events are warning
2. states are healthy as long as for the last 10 minutes, there are <5 hw slowdown warning events
3. states become unhealthy for the last 10 minutes, there are >=5 hw slowdown warning events

xiang90: can we make it a ratio? like 60%? do not use abs value

gyuho: @xiang90 @cardyok Now it's checking the per-minute frequency 60%.

For the last 10 minutes, 6 events make 60% ratio, thus unhealthy.
For the last 10 minutes, 3 events make 30% ratio, thus healthy.
If we use 100 minutes, 60 events make 60% ratio, thus unhealthy.

FillZpp (commented):
Generally LG
just one question: since the window inside the hw-slowdown component is 10min, but what's the inerval that the States() to be called? If the interval is longer than 10min, it may miss some events?

gyuho:
Internal is 1 minute for poller. We are fine with missing events for the first few minutes. Since we wait for the threshold. As long as we have 6 events in 10 min time window we will be able to catch

FillZpp (approved):
lgtm

gyuho:
/events part is tested

Screenshot 2025-01-21 at 10 01 39 PM

Testing /states

gyuho:
```json
  {
    "component": "accelerator-nvidia-hw-slowdown",
    "states": [
      {
        "name": "hw_slowdown",
        "reason": "hw slowdown events frequency per minute 1.00 (total events per minute count 10) exceeded threshold 0.60 for the last 10m0s"
      }
    ]
  },
```

Screenshot 2025-01-21 at 11 10 35 PM

Screenshot 2025-01-21 at 11 10 57 PM

PR #308, opened 2025-01-20
feat(nvidia/ib): remove "gpud run --expected-port-states-nvidia-infiniband" flag, only keep the default detection for backward compatibility

gyuho:
- remove `gpud run --expected-port-states-nvidia-infiniband` flag
- still keep the default ports/rates detection


components/accelerator/nvidia/infiniband/component_output.go
```diff
@@ -94,52 +92,33 @@ func ParseStatesToOutput(states ...components.State) (*Output, error) {
...
-			// no rate is set, use the default rate based on the product
-			atLeastRate := cfg.ExpectedPortStates.AtLeastRate
-			if atLeastRate == 0 {
-				atLeastRate = infiniband.SupportsInfinibandPortRate(o.GPUProductName)
-			}
-
-			if err := o.Ibstat.Parsed.CheckPortsAndRate(atLeastPorts, atLeastRate); err != nil {
-				return err.Error(), false, nil
-			}
+	if len(o.Ibstat.Parsed) > 0 {
```
FillZpp, line 113: Why return "no ibstat issue/error found" when `o.Ibstat.Parsed` is empty?

I think if the node group expects the node to have IB devices, then get empty parsed message should also be an error type?

gyuho: Good catch, fixing.

gyuho: @FillZpp Addressed, PTAL.

gyuho:
TODO: Still need some defaults for migration phase (outside of evaluate function)

gyuho:
TODO: Need a way to set threshold dynamically

cmd/gpud/command/command.go
```diff
@@ -252,11 +250,6 @@ sudo rm /etc/systemd/system/gpud.service
 					Usage: "enable 'kernel-module' component that returns healthy if and only if all the kernel modules are loaded (default: [], use '--kernel-modules-to-check=a --kernel-modules-to-check=b' for multiple modules)",
 					Value: &kernelModulesToCheck,
 				},
-				&cli.StringFlag{
-					Name:        "expected-port-states-nvidia-infiniband",
```
FillZpp, line 256: we already use this flag on prod machines, right? we have to make sure the commandline flag must be removed in next release, otherwise it may fail to start because of unknown flag

FillZpp (approved):

components/accelerator/nvidia/infiniband/component_output.go
```diff
@@ -94,56 +93,58 @@ func ParseStatesToOutput(states ...components.State) (*Output, error) {
 	return nil, errors.New("no state found")
 }
 
+var (
+	msgNoAtLeastPortsOrRateSet = "no at least ports or at least rate set"
```
ccding, line 97: the English is hard to read. change it to `must set ports or rate`. What does "at least" mean? can add it to the message if needed

components/accelerator/nvidia/infiniband/component_output.go
```diff
@@ -94,56 +93,58 @@ func ParseStatesToOutput(states ...components.State) (*Output, error) {
...
+	msgNoIbstatExists          = "no ibstat exists while configured to check ibstat"
+	msgNoIbstatDataFound       = "no ibstat data found while configured to check ibstat"
+	msgNoIbstatIssueFound      = "ok -- no ibstat issue/error found"
```
ccding, line 100: also improve these

cmd/gpud/command/command.go
gyuho: > we already use this flag on prod machines, right

Confirmed that, this is not being used. Safe to drop :)

PR #314, opened 2025-01-20
fix(diagnose): fix "gpud scan" (add missing temp db creation)

gyuho:


xiang90:
test?

gyuho:
@xiang90 Now running `gpud scan` on each e2e test runs. Also, manually tested in GPU machines.

xiang90:
can we also add some unit test?

ccding (approved):

ccding:
lgtm after adding tests

PR #315, opened 2025-01-20
feat(pkg/dmesg): helper for shared dmesg commands and timestamp parse function

gyuho:


components/diagnose/diagnose.go
```diff
@@ -170,7 +170,7 @@ func run(ctx context.Context, dir string, opts ...OpOption) error {
 		return err
 	}
 
-	if commandExists("ipmitool") {
+	if process.CommandExists("ipmitool") {
```
xiang90, line 173: revert all irrelevant changes. the changes should happen only in the dmesg pkg and affected lines in other pkg.

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+	"context"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/leptonai/gpud/pkg/process"
+)
+
+type Commands struct {
```
xiang90, line 13: make this private.

dmesg pkg should only expose this (or a very similar) interface and its implementation.
```
type Watcher interface {
    Watch() -> LogLine
    Close() error
}

type LogLine struct {
    Timestamp time.Time
    Level string
    Facility string
    Content string
}
```

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+	WatchCommands [][]string
+	TailCommands  [][]string
+	ParseTimeFunc func(line []byte) (time.Time, []byte, error)
+}
+
+func GetCommands(ctx context.Context) (*Commands, error) {
+	sinceSupported, err := checkDmesgSupportsSinceFlag(ctx)
```
xiang90, line 20: remove no need since

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+		ParseTimeFunc: ParseTimeFuncJournalctl(),
+	}, nil
+}
+
+const (
+	// defaultDmesgCmd DefaultDmesgCmdWithSince default scan dmesg command (in newer util-linux it works, but older is not)
+	// some old dmesg versions don't support --since, thus fall back to the one without --since and tail the last 200 lines
+	// ref. https://github.com/leptonai/gpud/issues/32
+	defaultDmesgCmd          = "dmesg --time-format=iso --nopager --buffer-size 163920"
+	defaultDmesgCmdWithSince = "dmesg --time-format=iso --nopager --buffer-size 163920 --since '1 hour ago'"
```
xiang90, line 45: remove

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+func WatchCommandsDmesg() [][]string {
+	return [][]string{
+		{defaultDmesgCmdWithSince + " -w || " + defaultDmesgCmd + " -w || true"},
+
+		// run last commands as fallback, in case "dmesg -w" flag only works in some machines
+		{defaultDmesgCmdWithSince + " -W || " + defaultDmesgCmd + " -W"},
+	}
+}
+
+func TailCommandsDmesg() [][]string {
```
xiang90, line 58: remove. we should only support watch which will return ALL log lines and then watch from the end to return future lines.

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+
+func ParseTimeFuncDmesg() func(line []byte) (time.Time, []byte, error) {
+	return ParseDmesgTimeISO
+}
+
+const (
+	// DefaultJournalCtlCmd default scan journalctl command
+	// in case "dmesg --time-format" or "dmesg --since" flags are not supported
+	defaultJournalCtlWatchCmd = "journalctl -qk -o short-iso --no-pager --since '1 hour ago' -f || true"
+	defaultJournalCtlTailCmd  = "journalctl -qk -o short-iso --no-pager --since '1 hour ago' | tail -n 200"
```
xiang90, line 72: remove

pkg/dmesg/commands.go
eahydra, line 13: Do components that use Commands need to call washCommands/tailcommands themselves?
Is it possible that hide these fields? Just let the component call the method watch dmesg data?
I'm curious what the difference is between Watch and Tail? Can I start watching from 1 hour ago? If yes, Tail can be deleted.

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+func WatchCommandsJournalctl() [][]string {
+	return [][]string{
+		{defaultJournalCtlWatchCmd},
+	}
+}
+
+func TailCommandsJournalctl() [][]string {
```
xiang90, line 81: remove. we do not use tournalctl at all

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+
+	line := strings.Join(lines, "\n")
+	line = strings.TrimSpace(line)
+
+	return checkDmesgVersionOutputForSinceFlag(line), nil
+}
+
+// dmesg --version that supports "--since" flag
+// ref. https://github.com/util-linux/util-linux/blob/master/Documentation/releases/v2.37-ReleaseNotes
+const dmesgSinceFlagSupportVersion = 2.37
```
xiang90, line 131: remove

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+
+var dmesgVersionRegPattern = regexp.MustCompile(`\d+\.\d+`)
+
+func checkDmesgVersionOutputForSinceFlag(verOutput string) bool {
```
xiang90, line 135: remove

pkg/dmesg/commands_test.go
```diff
@@ -0,0 +1,38 @@
+package dmesg
+
+import (
+	"testing"
+)
+
+func Test_checkDmesgVersionOutputForSinceFlag(t *testing.T) {
```
xiang90, line 7: remove

pkg/dmesg/dmesg.go
```diff
@@ -36,8 +35,8 @@ func ParseISOtimeWithError(line []byte) (time.Time, []byte, error) {
 	return parsedTime, extractedLine, nil
 }
 
-// ParseShortISOtimeWithError parses the timestamp from "journalctl -o short-iso" output lines.
-func ParseShortISOtimeWithError(line []byte) (time.Time, []byte, error) {
+// ParseJournalctlTimeShortISO parses the timestamp from "journalctl -o short-iso" output lines.
+func ParseJournalctlTimeShortISO(line []byte) (time.Time, []byte, error) {
```
xiang90, line 39: remove

pkg/dmesg/dmesg_test.go
```diff
@@ -18,29 +18,56 @@ func TestParseISOtimeWithFile(t *testing.T) {
...
+		}
 		t.Logf("time: %s, line: %s (original: %s)", time, extractedLine, line)
 	}
 }
 
-func TestParseShortISOtimeWithFile(t *testing.T) {
-	b, err := os.ReadFile("dmesg-time-format-short-iso.log")
-	if err != nil {
-		t.Fatalf("failed to read dmesg-time-format-short-iso.log: %v", err)
+func TestParseJournalctlTimeShortISO(t *testing.T) {
```
xiang90, line 32: remove

pkg/dmesg/dmesg_test.go
```diff
@@ -91,7 +118,7 @@ func TestParseISOtimeWithError(t *testing.T) {
 	}
 }
 
-func TestParseShortISOtimeWithError(t *testing.T) {
+func TestParseJournalctlTimeShortISOWithError(t *testing.T) {
```
xiang90, line 121: remove

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+	}, nil
+}
+
+const (
+	// defaultDmesgCmd DefaultDmesgCmdWithSince default scan dmesg command (in newer util-linux it works, but older is not)
+	// some old dmesg versions don't support --since, thus fall back to the one without --since and tail the last 200 lines
+	// ref. https://github.com/leptonai/gpud/issues/32
+	defaultDmesgCmd          = "dmesg --time-format=iso --nopager --buffer-size 163920"
+	defaultDmesgCmdWithSince = "dmesg --time-format=iso --nopager --buffer-size 163920 --since '1 hour ago'"
+	defaultDmesgTailCmd      = defaultDmesgCmdWithSince + " || " + defaultDmesgCmd + " | tail -n 200"
```
ccding, line 46: since we have checkDmesgSupportsSinceFlag, do we still need the `... || ...|tail` command?

pkg/dmesg/commands.go
```diff
@@ -0,0 +1,146 @@
...
+		TailCommands:  TailCommandsJournalctl(),
+		ParseTimeFunc: ParseTimeFuncJournalctl(),
+	}, nil
+}
+
+const (
+	// defaultDmesgCmd DefaultDmesgCmdWithSince default scan dmesg command (in newer util-linux it works, but older is not)
+	// some old dmesg versions don't support --since, thus fall back to the one without --since and tail the last 200 lines
+	// ref. https://github.com/leptonai/gpud/issues/32
+	defaultDmesgCmd          = "dmesg --time-format=iso --nopager --buffer-size 163920"
```
xiang90, line 44: let us bump the buffer size to 4MB (make sure we test this)... some machine may have configured a lager buffer size so we need a larger buffer here to get all messages.

xiang90:
closing in favor of https://github.com/leptonai/gpud/pull/319

PR #317, opened 2025-01-20
feat(state): add shared events table for all components

gyuho:


gyuho:
Each component gets its own table.

PR #318, opened 2025-01-20
feat(pkg/poller): simpler log poller with benchmarks

gyuho:

**SEE THE FIRST COMMIT FOR MINIMAL DIFF!**


gyuho:
Need simpler implementation

PR #319, opened 2025-01-21
feat(pkg/dmesg): simpler watcher

gyuho:
Addressing all feedback from https://github.com/leptonai/gpud/pull/315.

### DIFF
--- a/pkg/dmesg/watcher.go
+++ b/pkg/dmesg/watcher.go
@@ -176,6 +176,7 @@ func parseDmesgLine(line string) LogLine {
 	}
 
 	if len(line) < len(isoFormat) {
+		logLine.Timestamp = parsedTime
 		logLine.Content = line
 		return logLine
 	}
--- a/pkg/dmesg/watcher.go
+++ b/pkg/dmesg/watcher.go
@@ -182,7 +182,6 @@ func parseDmesgLine(line string) LogLine {
 	}
 
 	extractedLine := strings.TrimSpace(line[len(isoFormat):])
-	logLine.Timestamp = parsedTime
 	logLine.Content = extractedLine
 	return logLine
 }
--- a/pkg/dmesg/watcher.go
+++ b/pkg/dmesg/watcher.go
@@ -12,11 +12,11 @@ import (
 	"github.com/leptonai/gpud/pkg/process"
 )
 
-var DefaultWatchCommands = []string{
-	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -w || true",
+var DefaultWatchCommands = [][]string{
+	{"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -w || true"},
 
 	// run last commands as fallback, in case "dmesg -w" flag only works in some machines
-	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -W || true",
+	{"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -W || true"},
 }
 
 type LogLine struct {
@@ -39,16 +39,16 @@ type Watcher interface {
 }
 
 func NewWatcher() (Watcher, error) {
-	return newWatcher(DefaultWatchCommands...)
+	return newWatcher(DefaultWatchCommands)
 }
 
-func newWatcher(cmds ...string) (Watcher, error) {
+func newWatcher(cmds [][]string) (Watcher, error) {
 	if len(cmds) == 0 {
 		return nil, errors.New("no commands provided")
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	ch, err := watch(ctx, cmds...)
+	ch, err := watch(ctx, cmds)
 	if err != nil {
 		cancel()
 		return nil, err
@@ -71,13 +71,19 @@ func (w *watcher) Close() {
 
 func watch(
 	ctx context.Context,
-	cmds ...string,
+	cmds [][]string,
 ) (<-chan LogLine, error) {
 	ch := make(chan LogLine, 1000)
-	p, err := process.New(
-		process.WithCommand(cmds...),
-		process.WithRunAsBashScript(), // need to run as bash script when dmesg commands are complicated
-	)
+
+	opts := []process.OpOption{}
+	for _, cmd := range cmds {
+		opts = append(opts, process.WithCommand(cmd...))
+	}
+
+	// need to run as bash script when dmesg commands are complicated
+	opts = append(opts, process.WithRunAsBashScript())
+
+	p, err := process.New(opts...)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/dmesg/watcher.go
+++ b/pkg/dmesg/watcher.go
@@ -1,0 +1,197 @@
+package dmesg
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/leptonai/gpud/log"
+	"github.com/leptonai/gpud/pkg/process"
+)
+
+var DefaultWatchCommands = []string{
+	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -w || true",
+
+	// run last commands as fallback, in case "dmesg -w" flag only works in some machines
+	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -W || true",
+}
+
+type LogLine struct {
+	Timestamp time.Time
+	Facility  string
+	Level     string
+	Content   string
+
+	// Error is set when a dmesg command fails.
+	Error string
+}
+
+type Watcher interface {
+	// Watch returns a channel that emits log lines.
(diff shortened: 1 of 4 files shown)


pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,115 @@
...
+	"context"
+	"strings"
+	"time"
+
+	"github.com/leptonai/gpud/log"
+	"github.com/leptonai/gpud/pkg/process"
+)
+
+type LogLine struct {
+	Timestamp time.Time
```
xiang90, line 13: doesnt dmesg have defined level?

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,115 @@
...
+		if err := p.Close(ctx); err != nil {
+			log.Logger.Warnw("failed to abort command", "err", err)
+		}
+		close(ch)
+	}()
+
+	select {
+	case <-ctx.Done():
+		return
+	case <-time.After(100 * time.Millisecond): // wait for the process to start
```
ccding, line 58: should not rely on such things. the sleep time can be unreliable

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,134 @@
...
+	Content   string
+	Error     string
+}
+
+type Watcher interface {
+	Watch() <-chan LogLine
+	Close()
+}
+
+func NewWatcher(cmds ...string) (Watcher, error) {
```
xiang90, line 23: this is bad. watcher should be dmesg specific. cannot just pass in whatever command

pkg/dmesg/watcher.go
gyuho:
```
kern  :warn  : 2025-01-21T04:41:44,285147+00:00 NVRM: Xid (PCI:0000:38:00): 13, pid='<unknown>', name=<unknown>, Graphics Exception: ESR 0x556730=0x117000e 0x556734=0x24 0x556728=0xf81eb60 0x55672c=0x1174
kern  :warn  : 2025-01-21T04:41:44,287197+00:00 NVRM: Xid (PCI:0000:38:00): 43, pid=2924364, name=pt_main_thread, Ch 00000008
```

Let me use `dmesg --decode --time-format=iso --nopager` to get facility + level

pkg/dmesg/watcher.go
xiang90: if you want to enable test, do NewTestWatcher(). also, we should test dmesg directly... not something else. you can write to /dev/kmsg i guess or maybe not. cc @cardyok

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,134 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	ch, err := watch(ctx, cmds...)
```
ccding, line 25: pass in `[]string{"dmesg...."}` if len(cmds)==0

ccding:
need to merge the output of `dmesg -W` and `dmesg` if the kernel does not support `dmesg -w`

pkg/dmesg/watcher.go
gyuho: Now supports facility + level

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,176 @@
...
+	Facility  string
+	Level     string
+	Content   string
+
+	// Error is set when a dmesg command fails.
+	Error string
+}
+
+type Watcher interface {
+	Watch() <-chan LogLine
```
xiang90, line 31: document this. especially when the chan will be closed (dmesg error)

xiang90:
> Attention: Patch coverage is 67.32673% with 33 lines in your changes missing coverage. Please review.

Let us get the test coverage to 80%+

xiang90:
lgtm otherwise. defer to @eahydra

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,176 @@
...
+	return ch, nil
+}
+
+func wait(ctx context.Context, p process.Process, ch chan LogLine) {
+	defer func() {
+		// we don't close until either context is done or the process is exited
+		if err := p.Close(ctx); err != nil {
+			log.Logger.Warnw("failed to abort command", "err", err)
+		}
+		close(ch)
```
eahydra, line 88: Here has a race condition with read. Here close channel `ch` will cause the read panic when writing. So it's better to close ch in read goroutine.

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	extractedLine := strings.TrimSpace(line[len(isoFormat):])
+	logLine.Timestamp = parsedTime
+	logLine.Content = extractedLine
+	return logLine
+}
+
+func findISOTimestampIndex(s string) int {
+	re := regexp.MustCompile(`\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}`)
```
eahydra, line 170: it's better to make it as a global variable value.

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,176 @@
...
+	if len(decoded) > 0 {
+		logLine.Facility = decoded[0]
+	}
+	if len(decoded) > 1 {
+		logLine.Level = decoded[1]
+	}
+
+	line = line[yidx:]
+	if len(line) < len(isoFormat) {
+		return logLine
```
eahydra, line 154: if the yidx is greater than -1, it means the line has a timestamp, right?
so once we get a timestamp, we should parse it before return?

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,176 @@
...
+	); err != nil {
+		ch <- LogLine{Timestamp: time.Now().UTC(), Error: err.Error()}
+	}
+}
+
+const isoFormat = "2006-01-02T15:04:05,999999-07:00"
+
+// parses the timestamp from "dmesg --time-format=iso" output lines.
+// "The definition of the iso timestamp is: YYYY-MM-DD<T>HH:MM:SS,<microseconds>←+><timezone offset from UTC>."
+func parseDmesgLine(line string) LogLine {
```
eahydra, line 126: we need to add some tests for the function.

pkg/dmesg/watcher_test.go
```diff
@@ -0,0 +1,330 @@
...
+		want     LogLine
+		wantTime bool
+	}{
+		{
+			name:  "normal log line",
+			input: "kern  :warn  : 2024-01-21T04:41:44,123456-07:00 Test message",
+			want: LogLine{
+				Facility: "kern",
+				Level:    "warn",
+				Content:  "Test message",
```
eahydra, line 59: is the timestamp `2024-01-21T04:41:44,123456-07:00`?

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	line = line[yidx:]
+
+	parsedTime, err := time.Parse(isoFormat, line[:len(isoFormat)])
+	if err != nil {
+		log.Logger.Warnw("failed to parse timestamp", "err", err, "line", line)
+		logLine.Content = line
+		return logLine
+	}
```
eahydra, line 176: if we parsed timestamp successfully, so we should update the logLine.Timestamp = parsedTime?

eahydra (approved):
lgtm

pkg/dmesg/watcher.go
```diff
@@ -0,0 +1,197 @@
...
+	w.cancel()
+}
+
+func watch(
+	ctx context.Context,
+	cmds ...string,
+) (<-chan LogLine, error) {
+	ch := make(chan LogLine, 1000)
+	p, err := process.New(
+		process.WithCommand(cmds...),
```
ccding, line 78: do we simply run the `dmesg -w` and `dmesg -W` commands? I thought we should try `-w` first, and if it fails, we merge the outputs of `dmesg` and `dmesg -W`

gyuho: Yes

```
dmesg --decode --time-format=iso --nopager --buffer-size 163920 -w || true
dmesg --decode --time-format=iso --nopager --buffer-size 163920 -W
```

so whatever works it blocks

ccding: i mean we should merge the outputs of `dmesg` and `dmesg -W` if the first line fails

gyuho: @ccding Yes, already doing that. 

```go
var DefaultWatchCommands = []string{
	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -w || true",

	// run last commands as fallback, in case "dmesg -w" flag only works in some machines
	"dmesg --decode --time-format=iso --nopager --buffer-size 163920 -W || true",
}

func NewWatcher() (Watcher, error) {
	return newWatcher(DefaultWatchCommands...)
}
```

`TestWatchMultipleCommands` is testing this behavior :)

PR #321, opened 2025-01-21
feat(xid): simplify xid component

cardyok:


components/accelerator/nvidia/error/xid/component.go
```diff
@@ -17,18 +17,73 @@ import (
...
+	go func() {
+		// 1. initialize base state
+		for {
+			// 1. invoke os component in past 3 days, get all reboot events
+			// 2. if succeed
+			// 2.1 update state with reboot events and cache
+			// 3. if failed
+			//NOTE: this might fail if the os component is not ready, and initial xid state might be wrong
+			break
+		}
```
eahydra, line 58: IMO we should warmup these reboot events when initializing, not in a separate goroutine. 
Do we need to read the reboot event regularly? Once rebooted, the world will be destroyed.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -17,18 +17,73 @@ import (
...
+			case <-ctx.Done():
+				panic("context cancelled")
+			case rawDmesgLine, err := <-watcher.Watch():
+				// 1. parse xid
+				// 2. if not xid, skip
+				// 3. generate current event
+				//   3.1 fill in dmesg based xid
+				//   3.2 fill in nvml based xid (from current poller)
+				// 4. check local cache with timestamp/xid, if it is already in cache, skip
+				// 5. insert into db, cache
```
eahydra, line 44: Once we got the latest event, we can insert the event to the event list, and call the function above commented to get the current state.

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+			if err := json.Unmarshal([]byte(event.ExtraInfo[EventKeyErroXidData]), &currXidErr); err != nil {
+				log.Logger.Errorf("failed to unmarshal event %s %s extra info: %s", event.Name, event.Message, err)
+				continue
+			}
+			if event.Type == components.EventTypeCritical {
+				if health == "Unhealthy" {
+					continue
+				} else {
+					healthy = false
+					health = "Degraded"
```
xiang90, line 35: type define this in a common library

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+// StateUpdateBasedOnEvents resolves the state of the XID error component.
+// note: assume events are sorted by time in ascending order
+func StateUpdateBasedOnEvents(events []components.Event) components.State {
+	var xidRebootMap = make(map[uint64]int)
+	var stateExtraInfo map[string]string
+	var suggestedAction *common.SuggestedActions
+	var xidErr *XidError
+	rebootThreshold := 2
+	healthy := true
+	health := components.StateHealthy
```
eahydra, line 21: if `health` is not Healthy,  healthy is `false`, they are duplicates.

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+	stateError := ""
+	for _, event := range events {
+		if event.Name == EventNameErroXid {
+			var currXidErr XidError
+			if err := json.Unmarshal([]byte(event.ExtraInfo[EventKeyErroXidData]), &currXidErr); err != nil {
+				log.Logger.Errorf("failed to unmarshal event %s %s extra info: %s", event.Name, event.Message, err)
+				continue
+			}
+			if event.Type == components.EventTypeCritical {
+				if health == components.StateUnhealthy {
```
eahydra, line 31: we can define number values for these states that can make the implementation more simple.
For example:
```go
const(
     StateHealthy = 0
     StateDegraded = 1
     StateUnhealthy = 2
)
```

So we can do this:
```go
state := StateHealthy
if event.Type == components.EventTypeCritical {
   state = StateDegraded
}
if state <= currentState {
   // TODO: maybe they have different suggested actions, so we need to check it...
   continue
}
// get the latest suggested action
```

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// StateUpdateBasedOnEvents resolves the state of the XID error component.
+// note: assume events are sorted by time in ascending order
+func StateUpdateBasedOnEvents(events []components.Event) components.State {
+	var xidRebootMap = make(map[uint64]int)
+	var stateExtraInfo map[string]string
+	var suggestedAction *common.SuggestedActions
+	var xidErr *XidError
+	rebootThreshold := 2
```
eahydra, line 19: we should make it as a constant value.

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/components/common"
+	"github.com/leptonai/gpud/log"
+)
+
+// StateUpdateBasedOnEvents resolves the state of the XID error component.
+// note: assume events are sorted by time in ascending order
+func StateUpdateBasedOnEvents(events []components.Event) components.State {
+	var xidRebootMap = make(map[uint64]int)
+	var stateExtraInfo map[string]string
```
eahydra, line 16: we should remove this variable value.

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+	var suggestedAction *common.SuggestedActions
+	var xidErr *XidError
```
eahydra, line 18:
```suggestion
	var lastSuggestedAction *common.SuggestedActions
	var lastXidErr *XidError
```

components/accelerator/nvidia/error/xid/helper.go
eahydra, line 21:
```suggestion
	lastHealth := components.StateHealthy
```

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+			} else if event.Type == components.EventTypeFatal {
+				healthy = false
+				health = components.StateUnhealthy
+			} else {
+				continue
+			}
+			stateError = fmt.Sprintf("xid %d detected by %s", currXidErr.Xid, currXidErr.DataSource)
+			xidErr = &currXidErr
+			if currXidErr.SuggestedActionsByGPUd != nil {
+				if currXidErr.SuggestedActionsByGPUd.RepairActions[0] == common.RepairActionTypeRebootSystem {
```
eahydra, line 46: should check if the RepairActions is empty

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/components/common"
+	"github.com/leptonai/gpud/log"
+)
+
+// StateUpdateBasedOnEvents resolves the state of the XID error component.
+// note: assume events are sorted by time in ascending order
+func StateUpdateBasedOnEvents(events []components.Event) components.State {
```
eahydra, line 14:
```suggestion
func EvolveHealthyState(events []components.Event) components.State {
```

components/accelerator/nvidia/error/xid/store/events.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	c.timeIndex = make([]time.Time, 0, len(c.events))
+	for k := range c.events {
+		c.timeIndex = append(c.timeIndex, k.timestamp)
+	}
+
+	sort.Slice(c.timeIndex, func(i, j int) bool {
+		return c.timeIndex[i].Before(c.timeIndex[j])
+	})
```
eahydra, line 82: there is a race condition in these codes. if every time we need to rebuild time index by events,  so don't need to cache the index result.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,227 @@ package xid
...
+			}
 		}
-	}
+	}()
+	return nil
+}
 
-	return o, nil
+func (c *XIDComponent) States(_ context.Context) ([]components.State, error) {
+	return []components.State{c.currState}, nil
```
eahydra, line 167: A race condition occurs here

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
+	"github.com/leptonai/gpud/components/db"
+	os_id "github.com/leptonai/gpud/components/os/id"
 	"github.com/leptonai/gpud/components/query"
 	"github.com/leptonai/gpud/log"
+	pkg_dmesg "github.com/leptonai/gpud/pkg/dmesg"
 )
 
-func New(ctx context.Context, cfg nvidia_common.Config) components.Component {
+const DefaultRetentionPeriod = 3 * 24 * time.Hour
+const DefaultStateUpdatePeriod = 30 * time.Second
```
eahydra, line 28: keep same style as others.

```suggestion
const (
       DefaultRetentionPeriod = 3 * 24 * time.Hour
       DefaultStateUpdatePeriod = 30 * time.Second
)
```

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
+
+type XIDComponent struct {
+	rootCtx      context.Context
+	cancel       context.CancelFunc
+	poller       query.Poller
+	currState    components.State
+	setHealthyCh chan struct{}
+	store        db.Store
+	sync.RWMutex
 }
```
eahydra, line 69: It's better to move these before the `New` function. Additionally, giving the `RWMutex` field a name would improve readability.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
 	if err != nil {
-		return nil, err
+		log.Logger.Errorw("failed to create dmesg watcher", "error", err)
+		return nil
 	}
-
-	o := &Output{}
-	for _, logItem := range dmesgTailResults.TailScanMatched {
-		if logItem.Error != nil {
+	for {
```
eahydra, line 79: Kindly wrap the loop into a function.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
+		ticker := time.NewTicker(DefaultStateUpdatePeriod)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-c.rootCtx.Done():
+				return
+			case <-ticker.C:
+				osComponent, err := components.GetComponent(os_id.Name)
+				if err != nil {
+					log.Logger.Errorw("failed to get os component", "error", err)
```
eahydra, line 120: Missing continue; this will panic at line 122.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
+				}
+				osEvents, err := osComponent.Events(c.rootCtx, time.Now().Add(-3*24*time.Hour))
+				if err != nil {
+					log.Logger.Errorw("failed to get os states", "error", err)
```
eahydra, line 124: I'm curious: was the continue missing because you didn't want to include it, or was it an oversight?

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
-			log.Logger.Warnw("last query failed -- returning cached, possibly stale data", "error", lerr)
+	go func() {
+		ticker := time.NewTicker(DefaultStateUpdatePeriod)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-c.rootCtx.Done():
+				return
+			case <-ticker.C:
+				osComponent, err := components.GetComponent(os_id.Name)
```
eahydra, line 118: I'm curious because I remember we discussed whether we needed to pull the OS state regularly. At that time, we decided it wasn't necessary, but now it seems we need to pull the latest OS state after all.

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,240 @@ package xid
...
 
-func (c *component) Close() error {
-	log.Logger.Debugw("closing component")
-
+func (c *XIDComponent) Close() error {
+	log.Logger.Debugw("closing XIDComponent")
 	// safe to call stop multiple times
 	c.poller.Stop(nvidia_component_error_xid_id.Name)
-
+	c.cancel()
```
eahydra, line 225: I'm not sure if cancel is only called once. If not, we should use sync.Once to ensure it's only called once.

components/accelerator/nvidia/error/xid/component_output.go
```diff
@@ -181,47 +178,10 @@ func (o *Output) GetReason() Reason {
 const (
 	EventNameErroXid = "error_xid"
 
-	EventKeyErroXidUnixSeconds    = "unix_seconds"
-	EventKeyErroXidData           = "data"
-	EventKeyErroXidEncoding       = "encoding"
-	EventValueErroXidEncodingJSON = "json"
+	EventKeyErroXidData = "data"
```
eahydra, line 181: should we remove this file?

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -0,0 +1,136 @@
...
+// note: assume events are sorted by time in ascending order
+func EvolveHealthyState(events []components.Event) (ret components.State) {
+	defer func() {
+		log.Logger.Debugf("EvolveHealthyState: %v", ret)
+	}()
+	var lastSuggestedAction *common.SuggestedActions
+	var lastXidErr *XidError
+	lastHealth := StateHealthy
+	xidRebootMap := make(map[uint64]int)
+	for i := len(events) - 1; i >= 0; i-- {
```
eahydra, line 32: It seems that the events are sorted by time in descending order? This conflicts with the comment.

components/accelerator/nvidia/error/xid/component.go
cardyok: wanted to skip it,  if we continue when os.Events fail we might incorrectly update state

components/accelerator/nvidia/error/xid/component.go
cardyok: yes, during testing, figured that we can manually generate reboot event (without actually rebooting) and this will be skipped -> incorrect state. Even this is not likely to happen on prod, i think its better for Xid component to cover it..
And since we need to periodically update state anyway (to reconcile when events are purged), thought might as well just make the stateUpdate logic generic to both initialization and periodic invoking..Since this only introduce a minute level extra sqlite query, not too heavy..

components/accelerator/nvidia/error/xid/component.go
cardyok: yea its only called once (since component.Close will only be called once)

components/accelerator/nvidia/error/xid/component_output.go
cardyok: yes, actually this whole file can be removed, wanted to create a separate pr to remove all currently unused file

components/accelerator/nvidia/error/xid/helper.go
cardyok: yes, discussed with other maintainers offline, we decided its best practice for Event interface to give back descending order events (thus descending here), comment is wrong, updated..

eahydra (approved):

internal/session/serve.go
```diff
@@ -82,7 +84,29 @@ func (s *Session) serve() {
 
 		case "delete":
 			go s.deleteMachine(ctx, payload)
-
+		case "sethealthy":
```
gyuho, line 87: Should we `setHealthy` to be consistent with `updateConfig` below?

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,242 @@ package xid
...
+				return nil
+			}
+			log.Logger.Errorw("failed to fetch current events", "error", err)
+			time.Sleep(initializeBackoff)
+			continue
 		}
+		break
 	}
-	dmesgTailResults, err := dmesgComponent.TailScan()
+	watcher, err := pkg_dmesg.NewWatcher()
```
gyuho, line 85: Do we need `components.Start` explicitly when we can just call this in the https://github.com/leptonai/gpud/pull/321/files#diff-0de25174214578a8017687aeff4c98186ec09357955a44eec993c3a024f5ec5cR44?

```go
New(...
  go func() { if err := c.updateCurrentState(); err != nil {
  watcher, err := pkg_dmesg.NewWatcher()
  go watcher.start

File: components/accelerator/nvidia/error/xid/component.go [hunk 24]
```diff
@@ -5,150 +5,242 @@ package xid
...
+			if err != nil {
+				log.Logger.Errorw("failed to get all events", "error", err)
+				continue
+			}
+			c.mu.Lock()
+			c.currState = EvolveHealthyState(events)
+			c.mu.Unlock()
+		case dmesgLine := <-watcher.Watch():
+			log.Logger.Debugw("dmesg line", "line", dmesgLine)
+			ev, err := nvidia_query_xid.ParseDmesgLogLine(metav1.Time{Time: dmesgLine.Timestamp}, dmesgLine.Content)
```
gyuho, line 158: We should now use 

https://github.com/leptonai/gpud/blob/02d18149b2f776b9d2f6fc562b2013bca3858556/components/accelerator/nvidia/query/xid/dmesg/dmesg.go#L61-L64

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,242 @@ package xid
...
+			}
+			event := components.Event{
+				Time: metav1.Time{Time: dmesgLine.Timestamp},
+				Name: EventNameErroXid,
+				ExtraInfo: map[string]string{
+					EventKeyErroXidData: strconv.FormatInt(int64(ev.Detail.Xid), 10),
+					EventKeyDeviceUUID:  ev.DeviceUUID,
+				},
+			}
+			currEvent, err := c.store.Find(c.rootCtx, event)
```
gyuho, line 175: Not going to work... Find -> json.Marshal(event struct) -> json marshal .extraInfo -> emits unordered bytes (e.g., {a:b, 1:2} vs. {1:2, a:b}, and the select query will fail.

Let me see if we can make this more deterministic

internal/session/serve.go
cardyok: feel its better to leave separate bc we might sethealthy a lot of components at same time while updateConfig is component specific..

components/accelerator/nvidia/error/xid/component.go
cardyok: yes we do..created Start interface to guarantee ordering since now xid depends on os finish initializing

components/accelerator/nvidia/error/xid/component.go
cardyok: oh good catch..lemme fix

gyuho: See https://github.com/leptonai/gpud/pull/334 for fix.

cardyok: actually prob no need to make it ordered, can just remove the extraInfo query and go through extraInfo in findEvent

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -5,150 +5,239 @@ package xid
...
 
 	cctx, ccancel := context.WithCancel(ctx)
 	getDefaultPoller().Start(cctx, cfg.Query, nvidia_component_error_xid_id.Name)
 
-	return &component{
-		rootCtx: ctx,
-		cancel:  ccancel,
-		poller:  getDefaultPoller(),
+	extraEventCh := make(chan *components.Event, 256)
+	localStore, err := db.NewStore(dbRW, dbRO, "components_accelerator_nvidia_error_xid_events", DefaultRetentionPeriod)
```
gyuho, line 52:
```suggestion
	localStore, err := db.NewStore(dbRW, dbRO, db.CreateDefaultTableName(nvidia_component_error_xid_id.Name), DefaultRetentionPeriod)
```

gyuho:
Also need to use this events DB in:

https://github.com/leptonai/gpud/blob/02d18149b2f776b9d2f6fc562b2013bca3858556/components/accelerator/nvidia/query/nvml/xid.go#L179-L190

---

TODO: need to deprecate old xid matcher:

https://github.com/leptonai/gpud/blob/02d18149b2f776b9d2f6fc562b2013bca3858556/internal/server/server.go#L332-L390

gyuho:
Discussed offline, we will do NVML in a separate PR

components/db/events.go
```diff
@@ -237,11 +237,17 @@ func createTable(ctx context.Context, db *sql.DB, tableName string) error {
...
-		return fmt.Errorf("failed to marshal extra info: %w", err)
+	var extraInfoJSON, suggestedActionsJSON []byte
+	var err error
+	if ev.ExtraInfo != nil {
+		extraInfoJSON, err = json.Marshal(ev.ExtraInfo)
+		if err != nil {
+			return fmt.Errorf("failed to marshal extra info: %w", err)
+		}
+	}
+	if ev.SuggestedActions != nil {
```
gyuho, line 248: nice catch

gyuho (approved):

xiang90:
let us improve the test coverage to make the CI green.

PR #322, opened 2025-01-21
feat(os/events): use read-only for reboot events, use common db pkg

gyuho:
- use read-only sqlite instance for os/reboot events query
- rename boot_unix_timestamp to timestamp
- rename machine_boot_ids to component_os_boot_events
- remove os boot state package to os/state

Tested

```json
  {
    "component": "os",
    "startTime": "2024-11-15T04:48:56Z",
    "endTime": "2025-01-21T16:27:28.879045767Z",
    "events": [
      {
        "time": "2025-01-21T16:24:02Z",
        "name": "reboot",
        "type": "Warning",
        "message": "system reboot detected (3 minutes ago)"
      },
      {
        "time": "2025-01-21T14:49:16Z",
        "name": "reboot",
        "type": "Warning",
        "message": "system reboot detected (1 hour ago)"
      }
    ]
  }
```

Screenshot 2025-01-22 at 2 51 09 AM


components/os/state/state.go
```diff
@@ -0,0 +1,163 @@
...
+		TableNameBootEvents,
+		ColumnBootID,
+		ColumnTimestamp),
+		bootID, ts.UTC().Unix())
+	sqlite.RecordInsertUpdate(time.Since(start).Seconds())
+
+	return err
+}
+
+type BootEvent struct {
```
xiang90, line 120: should this be in event.go? also we SHOULD NOT persist state into sqlite. state is meant to be transient.

gyuho: Yeah wasn't a good file name. Renaming.

components/os/db/events.go
```diff
@@ -7,28 +7,38 @@ import (
...
 
-	ColumnBootID          = "boot_id"
-	ColumnBootUnixSeconds = "boot_unix_seconds"
+	ColumnBootID    = "boot_id"
+	ColumnTimestamp = "timestamp"
 )
 
-func CreateTableBootIDs(ctx context.Context, db *sql.DB) error {
+func CreateTable(ctx context.Context, db *sql.DB) error {
+	if _, err := db.ExecContext(ctx, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, DeprecatedTableNameMachineBootIDs)); err != nil {
```
ccding, line 23: do we delete the table very time we restart gpud?

components/os/db/events.go
```diff
@@ -55,49 +65,50 @@ func GetFirstBootID(ctx context.Context, db *sql.DB) (string, error) {
...
+	query := fmt.Sprintf(`SELECT %s, %s
+FROM %s
+WHERE %s > ?
+ORDER BY %s DESC`,
+		ColumnBootID, ColumnTimestamp,
+		TableNameBootEvents,
+		ColumnTimestamp,
+		ColumnTimestamp,
+	)
+	params := []any{since.UTC().Unix()}
```
ccding, line 94:
```suggestion
	params := []any{since.Unix()}
```

components/os/db/events.go
gyuho: Yes, but will be removed in the next/following GPUd release (e.g., v0.4.1)

ccding: is it for migration? what if a user upgraded directly from 0.3.9 to 0.4.1? will it mess things up?

ccding: i think we should simply use a different table name if we want to significantly change the schema

gyuho: > from 0.3.9 to 0.4.1? will it mess things up?

Should be safe. We always create table (if not exists). This creation statement will succeed in either case.

Worst case, older versions will have dangling, unused tables. We can keep this drop statement for a long time.

ccding: so we are okay to lose all data during gpud restarts?

gyuho: @ccding Yes.

cardyok (approved):
e2e tested and lgtm after fixing the "boot" -> "reboot" problem

PR #324, opened 2025-01-22
feat(components/memory): use common db + dmesg poller for events, move out of "dmesg" component

gyuho:
c.f., https://github.com/leptonai/gpud/pull/322

Tested:

Screenshot 2025-01-25 at 10 38 26 AM

```json
    "component": "memory",
    "startTime": "2024-11-15T04:48:56Z",
    "endTime": "2025-01-25T02:39:01.610443725Z",
    "events": [
      {
        "time": "2025-01-25T02:38:04Z",
        "name": "memory_oom_kill_constraint",
        "type": "Warning",
        "message": "oom kill constraint detected",
        "extra_info": {
          "log_line": "oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=cri-containerd-3fc28fa1c647ceede9f6b340d0b16c9f1f663698972d22a52e296f291638e014.scope,mems_allowed=0-1,oom_memcg=/lxc.payload.ny2g2r5hh3-lxc/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-poda8697f49_441d_4d4f_90d2_6d8e1fa3bbe7.slice,task_memcg=/lxc.payload.ny2g2r5hh3-lxc/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-poda8697f49_441d_4d4f_90d2_6d8e1fa3bbe7.slice/cri-containerd-3fc28fa1c647ceede9f6b340d0b16c9f1f663698972d22a52e296f291638e014.scope,task=node,pid=863987,uid=0"
        },
        "suggested_actions": {
          "descriptions": null,
          "repair_actions": null
        }
      },
      {
        "time": "2025-01-25T02:36:53Z",
        "name": "memory_oom_kill_constraint",
        "type": "Warning",
        "message": "oom kill constraint detected",
        "extra_info": {
          "log_line": "oom-kill:constraint=OTHER_CONSTRAINT"
        },
        "suggested_actions": {
          "descriptions": null,
          "repair_actions": null
        }
      },
      {
        "time": "2025-01-25T02:36:50Z",
        "name": "memory_oom_kill_constraint",
        "type": "Warning",
        "message": "oom kill constraint detected",
        "extra_info": {
          "log_line": "oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=cri-containerd-3fc28fa1c647ceede9f6b340d0b16c9f1f663698972d22a52e296f291638e014.scope,mems_allowed=0-1,oom_memcg=/lxc.payload.ny2g2r5hh3-lxc/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-poda8697f49_441d_4d4f_90d2_6d8e1fa3bbe7.slice,task_memcg=/lxc.payload.ny2g2r5hh3-lxc/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-poda8697f49_441d_4d4f_90d2_6d8e1fa3bbe7.slice/cri-containerd-3fc28fa1c647ceede9f6b340d0b16c9f1f663698972d22a52e296f291638e014.scope,task=node,pid=863987,uid=0"
        },
        "suggested_actions": {
          "descriptions": null,
          "repair_actions": null
        }
      },
      {
        "time": "2025-01-25T02:36:46Z",
        "name": "memory_oom",
        "type": "Warning",
        "message": "oom detected",
        "extra_info": {
          "log_line": "Out of memory: Kill process 456 (python) score 50 or sacrifice child"
        },
        "suggested_actions": {
          "descriptions": null,
          "repair_actions": null
        }
      },
      {
        "time": "2025-01-25T02:36:42Z",
        "name": "memory_oom",
        "type": "Warning",
        "message": "oom detected",
        "extra_info": {
          "log_line": "Out of memory: Killed process 123, UID 48, (httpd)"
        },
        "suggested_actions": {
          "descriptions": null,
          "repair_actions": null
        }
      }
    ]
```


components/memory/db/events.go
```diff
@@ -0,0 +1,169 @@
+package db
```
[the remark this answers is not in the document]
gyuho: `CreateTable` failures should be fine. We just let it fail in server initialization and try again :)

SQLite can easily do >10k reads, so good enough for GPUd.

eahydra:
About the test result:
image
IMO we only need to get an aggregated result? like `OOM happened, killed processA(pid: 123)`?

pkg/dmesg/watcher.go
```diff
@@ -177,18 +181,20 @@ func parseDmesgLine(line string) LogLine {
 
 	line = line[yidx:]
 
-	parsedTime, err := time.Parse(isoFormat, line[:len(isoFormat)])
-	if err != nil {
-		log.Logger.Warnw("failed to parse timestamp", "err", err, "line", line)
+	// need to parse first
+	// if the line doesn't have the iso format header, no need to parse
```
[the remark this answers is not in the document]
gyuho: @eahydra  Actually, the regex was wrong... which was why it was failing in my test. Fixed it to enforce the exact length, so we don't need this change

gyuho:
@eahydra This is the current behavior. Needs confirmation from @cardyok if we need to aggregate.

cardyok:
i think for dmesg errors that contains multiple lines to determine actually issue (like oom), its the best if we can aggregate it, but might make gpud logic too complicated

gyuho:
> can aggregate

Oh actually, these are all different events, different log lines (just happens to be at the same time, since I injected them at the same time)

xiang90:
The name should not be `dmesg_matched`... Let us fix this. Name should be `OOMed` or something more descriptive.

components/memory/db/events.go
```diff
@@ -0,0 +1,169 @@
...
+	ColumnTimestamp = "timestamp"
+	// e.g., "dmesg"
+	ColumnDataSource = "data_source"
+	// e.g., "memory_oom", "memory_oom_kill_constraint", "memory_oom_cgroup", "memory_edac_correctable_errors"
+	ColumnEventType = "event_type"
+	// e.g., "oom_reaper: reaped process 345646 (vector), now anon-rss:0kB, file-rss:0kB, shmem-rss:0"
+	ColumnEventDetails = "event_details"
+)
+
+func CreateTable(ctx context.Context, db *sql.DB) error {
```
ccding, line 29: can we create a struct so we do not have to pass db in every time?
```go
type EventDB struct {
    db *sql.DB
}

func New(ctx context.Context) (*sql.DB, error) {
...
}

func (db *EventDB) Insert(ev Event) error {
...
}
```

components/memory/db/events.go
```diff
@@ -0,0 +1,169 @@
...
+	"time"
+
+	"github.com/leptonai/gpud/log"
+	"github.com/leptonai/gpud/pkg/sqlite"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+const (
+	TableNameMemoryEvents = "component_memory_events"
```
ccding, line 17: it appears most event tables have the same schema. Can we make this package universal and used by all components that reads/writes an event table? only need to pass in a different table name, and everything else should be the same

gyuho: https://github.com/leptonai/gpud/pull/327 addresses this.

components/memory/component.go
```diff
@@ -5,42 +5,62 @@ import (
...
-		cancel:  ccancel,
-		poller:  getDefaultPoller(),
+	eventsStore, err := events_db.NewStore(
+		cfg.Query.State.DBRW,
+		cfg.Query.State.DBRO,
+		events_db.CreateDefaultTableName(memory_id.Name),
+		3*24*time.Hour,
+	)
+	if err != nil {
+		ccancel()
```
ccding, line 34: it is a bit strange to cancel on error. Do we really need to create a `cctx, ccancel := context.WithCancel(ctx)`? Can we just use ctx?

components/memory/watcher.go
```diff
@@ -0,0 +1,120 @@
...
+
+type watcher struct {
+	ctx         context.Context
+	eventsStore events_db.Store
+
+	closeOnce    sync.Once
+	dmesgWatcher pkg_dmesg.Watcher
+}
+
+func startWatch(ctx context.Context, eventsStore events_db.Store) (*watcher, error) {
```
ccding, line 25: newWatcher?

components/memory/watcher.go
```diff
@@ -0,0 +1,120 @@
...
+			ev := components.Event{
+				Time: metav1.Time{Time: line.Timestamp.UTC()},
+				Type: common.EventTypeWarning,
+				ExtraInfo: map[string]string{
+					EventKeyLogLine: line.Content,
+				},
+			}
+
+			switch {
+			case memory_dmesg.HasOOM(line.Content):
```
ccding, line 66: I wonder if we should store a map from regex to <name, message> so we do not have to write a list of switch...case and functions like HasOOM, HasOOMKillConstraint, HasOOMKiller etc

pkg/dmesg/testdata/dmesg.decode.iso.log.1
```diff
@@ -287,7 +287,6 @@ kern  :info  : 2025-01-17T15:36:17,285006+00:00 pci 0000:ff:1e.1: [8086:3259] ty
 kern  :info  : 2025-01-17T15:36:17,292999+00:00 pci 0000:ff:1e.2: [8086:325a] type 00 class 0x088000
 kern  :info  : 2025-01-17T15:36:17,297001+00:00 pci 0000:ff:1e.3: [8086:325b] type 00 class 0x088000
 kern  :info  : 2025-01-17T15:36:17,304997+00:00 pci 0000:ff:1e.4: [8086:325c] type 00 class 0x088000
-kern  :info  : 2025-01-17T15:36:17,308997+00:00 pci 0000:ff:1e.5: [8086:325d] type 00 class 0x088000
```
ccding, line 290: why?

gyuho:
@ccding All fixed, thx

components/memory/component.go
gyuho: creating event store first, so we don't need call `ccancel`

ccding: @cardyok do you have any idea how to make the ctx usage look better?

components/memory/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+				},
+			}
+
+			for _, m := range memory_dmesg.GetMatches() {
+				if m.Check(line.Content) {
+					ev.Name = m.Name
+					ev.Message = m.Message
+					break
+				}
+			}
```
ccding, line 72: move this loop into the `memory_dmesg.GetMatches()` function like `memory_dmesg.GetMatches(line string) (name string, message string)`

components/memory/watcher.go
```diff
@@ -0,0 +1,107 @@
...
+
+			cctx, ccancel := context.WithTimeout(w.ctx, 15*time.Second)
+			found, err := w.eventsStore.Find(cctx, components.Event{
+				Time: ev.Time,
+				Name: ev.Name,
+				Type: ev.Type,
+			})
+			ccancel()
+			if err != nil {
+				log.Logger.Errorw("failed to find event", "error", err)
```
ccding, line 85: also log event name, etc

components/memory/watcher.go
gyuho: done

components/memory/component.go
```diff
@@ -5,42 +5,61 @@ import (
...
-		cancel:  ccancel,
-		poller:  getDefaultPoller(),
+	w, err := newWatcher(cctx, eventsStore)
+	if err != nil {
+		ccancel()
+		return nil, err
 	}
+
+	return &component{
+		rootCtx:     ctx,
```
ccding, line 44: should we use cctx here so it can match the ccancel?

also, why do we call it rootCtx rather than ctx?

ccding (approved):

components/memory/component.go
gyuho: renamed and fixed

PR #327, opened 2025-01-23
feat(components/db): add common events db helper

gyuho:
c.f., https://github.com/leptonai/gpud/pull/324#discussion_r1926266501

### DIFF
--- a/components/db/events.go
+++ b/components/db/events.go
@@ -1,0 +1,247 @@
+package db
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/leptonai/gpud/log"
+	"github.com/leptonai/gpud/pkg/sqlite"
+
+	_ "github.com/mattn/go-sqlite3"
+)
+
+const (
+	// Event timestamp in unix seconds.
+	ColumnTimestamp = "timestamp"
+
+	// e.g., "dmesg", "nvml", "nvidia-smi".
+	ColumnDataSource = "data_source"
+
+	// e.g., "memory_oom", "memory_oom_kill_constraint", "memory_oom_cgroup", "memory_edac_correctable_errors".
+	ColumnEventType = "event_type"
+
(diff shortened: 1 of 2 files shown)


components/db/events.go
```diff
@@ -0,0 +1,247 @@
...
+	ColumnDataSource = "data_source"
+
+	// e.g., "memory_oom", "memory_oom_kill_constraint", "memory_oom_cgroup", "memory_edac_correctable_errors".
+	ColumnEventType = "event_type"
+
+	// e.g., "xid", "sxid".
+	ColumnEventID1 = "event_id_1"
+
+	// e.g., "gpu_id", "gpu_uuid".
+	ColumnEventID2 = "event_id_2"
```
ccding, line 30: as discussed offline, let's rename these

components/db/events.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	return &DB{
+		table: tableName,
+		dbRW:  dbRW,
+		dbRO:  dbRO,
+	}, nil
+}
+
+func (db *DB) Insert(ctx context.Context, ev Event) error {
+	return insertEvent(ctx, db.dbRW, db.table, ev)
```
ccding, line 65: do we allow one of dbRW and dbRO to be nil in the NewDB function? if yes, we should check nil before calling insertEvent

components/db/events.go
```diff
@@ -0,0 +1,247 @@
...
+		); err != nil {
+			return nil, err
+		}
+		if eventID1.Valid {
+			event.EventID1 = eventID1.String
+		}
+		if eventID2.Valid {
+			event.EventID2 = eventID2.String
+		}
+		events = append(events, event)
```
ccding, line 220: wrap this to a function to avoid duplicate code

gyuho: Addressed in https://github.com/leptonai/gpud/pull/332

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/332.

PR #331, opened 2025-01-24
fix(hw_slowdown): fix description

cardyok:


components/accelerator/nvidia/hw-slowdown/component.go
```diff
@@ -152,7 +152,7 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 					common.RepairActionTypeHardwareInspection,
 				},
 				Descriptions: []string{
-					"Check the GPU hardware status",
+					"Hardware slowdown are often caused by GPU overheating, power supply unit (PSU) failing or hardware damage, please do a hardware inspection to mitigate the issue",
```
xiang90, line 155: what is hardware damage... 

> Constant hardware slowdowns are often the result of either a failing power supply (PSU) or elevated GPU temperatures.

cardyok: like cooling system pin broken..will remove it

xiang90:
lgtm

gyuho (approved):

PR #332, opened 2025-01-24
feat(components/db): add common events store (similar to components.Event)

gyuho:
Replace https://github.com/leptonai/gpud/pull/327.

Cherry-picked from https://github.com/leptonai/gpud/pull/321.

### DIFF
--- a/components/db/events.go
+++ b/components/db/events.go
@@ -3,15 +3,25 @@ package db
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
 
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/components/common"
 	"github.com/leptonai/gpud/log"
 	"github.com/leptonai/gpud/pkg/sqlite"
 
 	_ "github.com/mattn/go-sqlite3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// Creates the default table name for the component.
+// The table name is in the format of "components_{component_name}_events_v0_4_0".
+func CreateDefaultTableName(componentName string) string {
+	return fmt.Sprintf("components_%s_events_v0_4_0", componentName)
+}
 
 const (
 	// Event timestamp in unix seconds.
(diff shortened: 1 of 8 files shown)


components/db/events.go
```diff
@@ -0,0 +1,284 @@
...
+)
+
+// Creates a new DB instance with the table created.
+func NewStore(ctx context.Context, dbRW *sql.DB, dbRO *sql.DB, tableName string) (*storeImpl, error) {
+	if dbRW == nil {
+		return nil, ErrNoDBRWSet
+	}
+	if dbRO == nil {
+		return nil, ErrNoDBROSet
+	}
```
ccding, line 66: if we do not allow them to be nil, what's the purpose of separating them? Can we use dbRW for both read and write?

gyuho: Oh we want to isolate reads since we have different open modes for write and reads.

gyuho: https://github.com/leptonai/gpud/blob/main/pkg/sqlite/sqlite.go#L34-L35

ccding: i understand this but what's the impact of GPUd if we use dbRW for both read and write?

gyuho: `getEvents` (reads) will need to wait when write transactiosn ongoing (insert events).

components/db/events.go
```diff
@@ -0,0 +1,284 @@
...
+	// e.g., "reboot"
+	ColumnSuggestedActions = "suggested_actions"
+)
+
+type Event struct {
+	Timestamp        int64  `json:"timestamp"`
+	Name             string `json:"name"`
+	Type             string `json:"type"`
+	ExtraInfo        string `json:"extra_info"`
+	SuggestedActions string `json:"suggested_actions"`
```
cardyok, line 45: why we drop message field tho..

gyuho: good catch... adding it back

components/db/events.go
```diff
@@ -0,0 +1,342 @@
...
+	// event message
+	// e.g., "VFS file-max limit reached"
+	Message string `json:"message"`
+
+	// event extra info
+	// e.g.,
+	// data source: "dmesg", "nvml", "nvidia-smi".
+	// event target: "gpu_id", "gpu_uuid".
+	// log detail: "oom_reaper: reaped process 345646 (vector), now anon-rss:0kB, file-rss:0kB, shmem-rss:0".
+	ExtraInfo string `json:"extra_info"`
```
ccding, line 65: is this a json?

components/db/events.go
```diff
@@ -0,0 +1,342 @@
...
+
+	// event suggested actions
+	// e.g., "reboot"
+	SuggestedActions string `json:"suggested_actions"`
```
ccding, line 69: why do we need the SuggestedActions here?

ccding (approved):

components/db/events.go
ccding: i don't think writes to table A will impact reads from table B... anyways we can keep it like this for now

components/db/events.go
gyuho: Directly maps to our event object 

https://github.com/leptonai/gpud/blob/789b4fa4aece34f67ca9a1a2eb12c87daa3e9362/components/components.go#L79-L86

/cc @cardyok

components/db/events.go
gyuho: Yes

components/db/events.go
ccding: is this calculated based on other fields?

components/db/events.go
ccding: probably not very good. for example, if we want to filter by GPU id

components/db/events.go
cardyok: not exactly..for xid, yes. but for other components might not..

components/db/events.go
gyuho: Currently, we have no use case for those detailed queries. If we have such use case, we will revisit this schema to add more columns :)

cardyok (approved):

PR #336, opened 2025-01-26
feat(nvidia/xid, hw slowdown, fuse, PCI): use common events store, simplify nvml poller initialization

gyuho:
- Use `components/db` to persist NVML-originated Xid events
- Switch to `components/db` to persist HW slowdown events

Should be after https://github.com/leptonai/gpud/pull/321


xiang90:
can we somehow split this into multiple PRs?

PR #337, opened 2025-01-27
feat(sxid): simplify sxid component 

cardyok:
similar to #321 ,make sxid separate pr


internal/session/serve.go
```diff
@@ -87,24 +88,43 @@ func (s *Session) serve() {
...
-				if err != nil {
-					log.Logger.Errorw("failed to get component", "error", err)
-					continue
-				}
-				if watchable, ok := rawComponent.(*metrics.WatchableComponentStruct); ok {
-					if component, ok := watchable.Component.(*xid.XIDComponent); ok {
-						if err = component.SetHealthy(); err != nil {
-							log.Logger.Errorw("failed to set xid healthy", "error", err)
+				switch componentName {
+				case "accelerator-nvidia-error-xid":
```
eahydra, line 92: I'm curious why you only handle these two components. It seems that the logic is general

internal/session/serve.go
```diff
@@ -87,24 +88,43 @@ func (s *Session) serve() {
...
+					rawComponent, err := components.GetComponent("accelerator-nvidia-error-xid")
+					if err != nil {
+						log.Logger.Errorw("failed to get component", "error", err)
+						continue
+					}
+					if watchable, ok := rawComponent.(*metrics.WatchableComponentStruct); ok {
+						if component, ok := watchable.Component.(*xid.XIDComponent); ok {
+							if err = component.SetHealthy(); err != nil {
+								log.Logger.Errorw("failed to set xid healthy", "error", err)
+							}
```
eahydra, line 102: IMO every component should be watchable, and I thought there was already a `Watchable` interface. However, it seems that such an interface doesn't actually exist. It appears that we are using `WatchableComponentStruct` to wrap other components, and WatchableComponentStruct overrides the States method to update component health metrics. Is it possible to remove this wrapper to make the interfaces clearer and improve readability? This would also simplify the code and make it easier to maintain.
/cc @gyuho @cardyok

components/accelerator/nvidia/error/sxid/component.go
```diff
@@ -4,95 +4,235 @@ package sxid
...
+
+			if currEvent != nil {
+				log.Logger.Debugw("no new events created")
+				continue
+			}
+			if err = c.store.Insert(c.rootCtx, event); err != nil {
+				log.Logger.Errorw("failed to create event", "error", err)
+				continue
+			}
+			events, err := c.store.Get(c.rootCtx, time.Time{})
```
gyuho, line 180: why do we need this `Get`?

gyuho: Oh got it, being used for states evaluation

internal/session/serve.go
gyuho: Yeah, the interface/unwrap being too complicated... If this is already tested, we can do the refactoring in a separate PR. @cardyok

internal/session/serve.go
gyuho: Use 

```go
nvidia_component_error_xid_id "github.com/leptonai/gpud/components/accelerator/nvidia/error/xid/id"
nvidia_component_error_xid_id.Name
```

?

cardyok: yes updated package name @gyuho 
yes SetHealthy logic is generic, finally we should make SetHealthy a generic interface to all Components, currently this only applies to xid/sxid so didn't add the interface thought can do it later.. @eahydra

eahydra: Let's create an issue for it?

gyuho (approved):

PR #339, opened 2025-01-28
feat(fuse): use common events store pkg

gyuho:
c.f., https://github.com/leptonai/gpud/pull/336


components/fuse/component.go
```diff
@@ -5,46 +5,53 @@ import (
...
 	"github.com/leptonai/gpud/components/query"
 	"github.com/leptonai/gpud/log"
 
-	"github.com/dustin/go-humanize"
 	"github.com/prometheus/client_golang/prometheus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func New(ctx context.Context, cfg Config) components.Component {
+func New(ctx context.Context, cfg Config) (components.Component, error) {
```
xiang90, line 20: Let us test this new func with a UT. should be easy enough

components/fuse/component.go
```diff
@@ -86,61 +93,8 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
...
-				EventKeyData:        string(eb),
-				EventKeyEncoding:    EventValueEncodingJSON,
-			},
-		})
-	}
-	if len(convertedEvents) == 0 {
-		return nil, nil
-	}
-	return convertedEvents, nil
+	return c.eventsStore.Get(ctx, since)
```
xiang90, line 97: need test. should be easy

components/fuse/component.go
```diff
@@ -171,6 +125,8 @@ func (c *component) Close() error {
 	// safe to call stop multiple times
 	c.poller.Stop(fuse_id.Name)
 
+	c.eventsStore.Close()
```
xiang90, line 128: test

components/fuse/component_output.go
```diff
@@ -64,27 +69,47 @@ func CreateGet(cfg Config) query.GetFunc {
...
+			if err != nil {
+				continue
+			}
+			ev := components.Event{
+				Time:    metav1.Time{Time: now.UTC()},
+				Name:    "fuse_connections",
+				Type:    common.EventTypeCritical,
+				Message: info.DeviceName + ": " + strings.Join(msgs, ", "),
+				ExtraInfo: map[string]string{
+					"data":     string(ib),
```
cardyok, line 100: current eventStore.Find cannot correctly compare ib content..but since we already have info.DeviceName and msg included in Message maybe its fine..
the rest lgtm

components/fuse/component.go
```diff
@@ -5,46 +5,53 @@ import (
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func New(ctx context.Context, cfg Config) components.Component {
+func New(ctx context.Context, cfg Config) (components.Component, error) {
+	eventsStore, err := events_db.NewStore(
+		cfg.Query.State.DBRW,
+		cfg.Query.State.DBRO,
+		events_db.CreateDefaultTableName(fuse_id.Name),
+		3*24*time.Hour,
```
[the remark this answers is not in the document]
gyuho: fixed in https://github.com/leptonai/gpud/pull/340, will rebase

components/fuse/component_output.go
gyuho:
```go
func compareEvent(eventA, eventB components.Event) bool {
	if len(eventA.ExtraInfo) != len(eventB.ExtraInfo) {
		return false
	}
	for key, value := range eventA.ExtraInfo {
		if val, ok := eventB.ExtraInfo[key]; !ok || val != value {
			return false
		}
	}
	return true
}
```

should be able to cover this?

cardyok: yea this only checks kv but does not unmarshal ib..

cardyok (approved):

PR #343, opened 2025-01-28
feat(nvidia/xid): use common events DB for NVML-based xid watcher, disable NVML Xid event watcher in favor of "dmesg" watcher, deprecate redundant "error-xid-sxid" component

gyuho:
- use common events DB for NVML-based xid watcher
- disable NVML Xid event watcher in favor of "dmesg" watcher
- deprecate redundant "error-xid-sxid" component

c.f., https://github.com/leptonai/gpud/pull/336


xiang90:
Can we add tests?

components/accelerator/nvidia/query/nvml/nvml.go
```diff
@@ -88,9 +88,16 @@ type instance struct {
 	// read-only database instance
 	dbRO *sql.DB
 
+	xidEventsStore events_db.Store
```
xiang90, line 91: why do we still need this in the nvml pkg if the decision is to only use dmsg for xid detection for now? cc @cardyok

gyuho: Let me remove, as we decided to use dmesg as a source of event

gyuho: removed

components/accelerator/nvidia/query/nvml/nvml.go
```diff
@@ -385,7 +387,7 @@ func (inst *instance) Start() error {
 		}
 	}
 
-	if inst.xidErrorSupported {
+	if inst.xidErrorSupported && inst.xidWatchWithNVML {
 		go inst.pollXidEvents()
```
xiang90, line 391: what is this? can we delete this func all together?

gyuho: removed as well

cardyok (commented):
lgtm after patch ci green

cardyok (approved):

PR #344, opened 2025-01-28
feat(fd): improve reasons, human-readable

gyuho:
To improve old message format:

> allocated_file_handles: 647456, allocated_percent: 0.00, running_pids: 2463, used_percent: 0.00; threshold_running_pids_percent is greater than 80

### DIFF
--- a/components/fd/component_output.go
+++ b/components/fd/component_output.go
@@ -96,6 +96,20 @@ const (
 	StateKeyFileHandlesSupported = "file_handles_supported"
 	// Set to true if the file descriptor limit is supported.
 	StateKeyFDLimitSupported = "fd_limit_supported"
+
+	// Threshold values for health checks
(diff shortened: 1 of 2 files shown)


components/fd/component_output.go
```diff
@@ -185,29 +185,29 @@ func (o *Output) States() ([]components.State, error) {
 
 	if allocatedPercent, err := o.GetAllocatedFileHandlesPercent(); err == nil && allocatedPercent > 95.0 {
 		state.Healthy = false
-		state.Reason += "; allocated_file_handles_percent is greater than 95"
+		state.Reason += "; file handles allocation exceeds 95%"
```
xiang90, line 188: can you define these as consts? then you can refer them in the tests as well.

xiang90:
LGTM after fixing the minor issue mentioned above.

xiang90 (approved):

PR #353, opened 2025-02-04
test(client/v1): increase unit test coverage

gyuho:


client/v1/healthz.go
```diff
@@ -82,5 +82,7 @@ func BlockUntilServerReady(ctx context.Context, addr string, opts ...OpOption) e
 			return fmt.Errorf("context done: %w", ctx.Err())
 		}
 	}
-	return errors.New("server not ready, timeout waiting")
+	return ErrServerNotReady
 }
+
+var ErrServerNotReady = errors.New("server not ready, timeout waiting")
```
eahydra, line 88: We should define this error variable at the beginning of line 14

client/v1/healthz_test.go
```diff
@@ -214,3 +309,156 @@ func TestBlockUntilServerReadyWithInvalidResponse(t *testing.T) {
 		t.Error("BlockUntilServerReady() with invalid response should return error")
 	}
 }
+
+func TestCheckHealthzConcurrent(t *testing.T) {
```
eahydra, line 313: IMO we don't need to do this test. Concurrency scenarios like this don't actually occur, and even if they do, there won't be any problems because the code is clear and simple.

client/v1/healthz_test.go
```diff
@@ -214,3 +309,156 @@ func TestBlockUntilServerReadyWithInvalidResponse(t *testing.T) {
...
+	err := BlockUntilServerReady(context.Background(), srv.URL,
+		WithHTTPClient(client),
+		WithCheckInterval(10*time.Millisecond))
+	assert.Error(t, err)
+	assert.Equal(t, err, ErrServerNotReady)
+
+	// Test with a client timeout longer than server response time
+	client = &http.Client{Timeout: 100 * time.Millisecond}
+	err = BlockUntilServerReady(context.Background(), srv.URL,
+		WithHTTPClient(client),
```
eahydra, line 381: The caller of BlockUntilServerReady doesn't pass in any client, do we really need this test? I thought for the function `BlockUntilServerReady` we only needed ctx, url, interval, and number of retries?

client/v1/options.go
eahydra, line 1: I think we should make the code simpler... We basically don't need to use the `options pattern` here.

client/v1/healthz_test.go
gyuho: removed

eahydra (approved):
lgtm

PR #355, opened 2025-02-05
test(pkg/systemd): add more unit tests to increasing coverage

gyuho:

### DIFF
--- a/pkg/systemd/dbus.go
+++ b/pkg/systemd/dbus.go
@@ -9,8 +9,14 @@ import (
 	"github.com/coreos/go-systemd/v22/dbus"
 )
 
+type dbusConn interface {
+	Close()
+	Connected() bool
+	GetUnitPropertiesContext(ctx context.Context, unit string) (map[string]interface{}, error)
+}
+
 type DbusConn struct {
-	conn *dbus.Conn
+	conn dbusConn
 }
 
 // Caller should explicitly close the connection by calling Close() on returned connection object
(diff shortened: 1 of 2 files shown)


pkg/systemd/dbus.go
```diff
@@ -37,13 +37,26 @@ func (c *DbusConn) IsActive(ctx context.Context, unitName string) (bool, error)
...
 	if err != nil {
-		return false, fmt.Errorf("unable to get unit properties for %s: %w", unitName, err)
+		return false, fmt.Errorf("unable to get unit properties for %s: %w", formattedUnitName, err)
+	}
+
+	return checkActiveState(props, formattedUnitName)
+}
+
+// formatUnitName ensures the unit name has the correct suffix
+func formatUnitName(unitName string) string {
```
eahydra, line 51: probably we should name the function as `normalizeServiceUnitName`?

pkg/systemd/dbus.go
```diff
@@ -37,13 +37,26 @@ func (c *DbusConn) IsActive(ctx context.Context, unitName string) (bool, error)
 	if !c.conn.Connected() {
 		return false, fmt.Errorf("connection disconnected")
 	}
-	if !strings.HasSuffix(unitName, ".target") && !strings.HasSuffix(unitName, ".service") {
```
eahydra, line 40: IMO we should define an interface for dbus.Conn, we should add an interface to DbusConn,  and we should generate a mock for DbusConn, so that other codes rely on the DbusConn can be testable.

eahydra (approved):

pkg/systemd/dbus.go
gyuho: done

PR #357, opened 2025-02-05
test(rootkeys): add unit tests

gyuho:


rootkeys/fs_test.go
```diff
@@ -0,0 +1,73 @@
+package rootkeys
+
+import (
+	"testing"
+)
+
+func TestRootsFS(t *testing.T) {
```
eahydra, line 7: I'm curious since RootsFs is only used by `third_party/tailscale/distsign`, why not move rootsfs to distsign package?

gyuho: > RootsFs is only used by third_party/tailscale/distsign

The `rootkeys/keys` contains the set of public keys managed by lepton, whereas `third_party/tailscale/distsign` is the code we borrowed from tailscale repo

eahydra: The problem is that the dependency is confusing. Referring to the original implementation at `https://github.com/tailscale/tailscale/blob/v1.76.3/clientupdate/distsign`, we only need to override the roots package. Therefore, I’d like to move the rootkeys package to the distsign package as roots, and revert the roots.go file.

third_party/tailscale/distsign/roots.go
```diff
@@ -11,7 +11,7 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/leptonai/gpud/rootkeys"
+	"github.com/leptonai/gpud/third_party/tailscale/distsign/rootkeys"
```
eahydra, line 14: Actually we can doing as following to reuse the tailscale tests
```go
//go:embed keys
var rootsFS embed.FS
```

and move the keys folder to distsign...

gyuho:
@eahydra Updated PTAL.

eahydra (approved):
lgtm

PR #358, opened 2025-02-05
test(errdefs): use more helpers, increase test coverage

gyuho:


client/v1/status.go
```diff
@@ -0,0 +1,41 @@
...
+	"crypto/tls"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/leptonai/gpud/manager/packages"
+)
+
+func GetStatus(ctx context.Context, url string) ([]packages.PackageStatus, error) {
```
xiang90, line 14: comment

xiang90:
lgtm. defer to @cardyok

client/v1/status.go
```diff
@@ -0,0 +1,46 @@
+package v1
+
+import (
```
eahydra, line 3: we should rename the file name, such as packages.go? or package_status.go? `status.go` is too general.

client/v1/status.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/leptonai/gpud/manager/packages"
+)
+
+// GetPackageStatus fetches the GPUd package status from the GPUd admin API.
+func GetPackageStatus(ctx context.Context, url string) ([]packages.PackageStatus, error) {
+	httpClient := &http.Client{
```
eahydra, line 16: Why not use options here?  It's recommended that if we need to customize http client, it's better to define a type `Client` represents the gpud client, and refactor the function as a Client's method. So that we can just focus on the   core logic in the function.

cmd/gpud/command/status.go
```diff
@@ -56,7 +51,9 @@ func cmdStatus(cliContext *cli.Context) error {
 	fmt.Printf("%s successfully checked gpud health\n", checkMark)
 
 	for {
-		packageStatus, err := getStatus()
+		cctx, ccancel := context.WithTimeout(rootCtx, 15*time.Second)
+		packageStatus, err := client.GetPackageStatus(cctx, fmt.Sprintf("https://localhost:%d/admin/packages", config.DefaultGPUdPort))
```
eahydra, line 55: A URI like `/admin/packages` should be defined in the apis package.

gyuho: Kept it in `/internal/server` for now, we can refactor API paths in a separate PR /cc @cardyok

client/v1/status.go
gyuho: Done

client/v1/status.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// GetPackageStatus fetches the GPUd package status from the GPUd admin API.
+func GetPackageStatus(ctx context.Context, url string, opts ...OpOption) ([]packages.PackageStatus, error) {
+	op := &Op{}
+	if err := op.applyOpts(opts); err != nil {
+		return nil, err
+	}
+
+	httpClient := op.httpClient
```
eahydra, line 20: should move it to line 25

cmd/gpud/command/status.go
cardyok: sgtm, we do not have remote dependencies for this api now.

eahydra (approved):
lgtm

PR #359, opened 2025-02-05
tests(internal/session): add more unit tests with smaller functions

gyuho:

### DIFF
--- a/internal/session/session.go
+++ b/internal/session/session.go
@@ -169,27 +169,8 @@ func (s *Session) keepAlive() {
 	}
 }
 
-func (s *Session) startWriter(ctx context.Context, writerExit chan any) {
-	pipeFinishCh := make(chan any)
-	goroutineCloseCh := make(chan any)
-	defer func() {
-		close(goroutineCloseCh)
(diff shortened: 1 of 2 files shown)


internal/session/session.go
```diff
@@ -288,30 +290,41 @@ func (s *Session) startReader(ctx context.Context, readerExit chan any) {
 		return
 	}
 
-	lastPackageTimestamp := &time.Time{}
-	*lastPackageTimestamp = time.Now()
+	s.processReaderResponse(resp, goroutineCloseCh, pipeFinishCh)
+}
+
+func (s *Session) processReaderResponse(resp *http.Response, goroutineCloseCh, pipeFinishCh chan any) {
+	lastPackageTimestamp := time.Now()
```
eahydra, line 297: there is a race condition with `lastPackageTimestamp`..

gyuho: @eahydra @cardyok Fixed

eahydra (approved):
lgtm, @cardyok PTAL

cardyok (approved):
lgtm

PR #360, opened 2025-02-05
test(nvidia/query/nvml): add clock speed, errors unit tests using mock/nvml

gyuho:
Using https://pkg.go.dev/github.com/NVIDIA/go-nvml@v0.12.4-0/pkg/nvml/mock

### DIFF
--- a/components/accelerator/nvidia/query/nvml/testutil/util.go
+++ b/components/accelerator/nvidia/query/nvml/testutil/util.go
@@ -5,34 +5,67 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/nvml/mock"
 )
 
-var _ device.Device = (*mockDevice)(nil)
+var _ device.Device = (*MockDevice)(nil)
 
-type mockDevice struct {
+type MockDevice struct {
 	*mock.Device
+	Architecture          string
+	Brand                 string
+	CudaComputeCapability string
+	PCIBusID              string
 }
 
-func (d *mockDevice) GetArchitectureAsString() (string, error) {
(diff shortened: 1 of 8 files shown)


components/accelerator/nvidia/query/nvml/testutil/util.go
```diff
@@ -12,27 +13,69 @@ type mockDevice struct {
 }
 
 func (d *mockDevice) GetArchitectureAsString() (string, error) {
-	return "", nil
+	return "Test Architecture", nil
```
xiang90, line 16: can we test this mock? or can we explicitly skip the test coverage check of this file?

gyuho: @xiang90 UTs are added

```
--- PASS: TestCreateDevice (0.00s)
    --- PASS: TestCreateDevice/default_values (0.00s)
    --- PASS: TestCreateDevice/custom_values (0.00s)
PASS
coverage: 97.5% of statements
ok      github.com/leptonai/gpud/components/accelerator/nvidia/query/nvml/testutil      0.258s
```

components/accelerator/nvidia/query/nvml/testutil/options.go
```diff
@@ -0,0 +1,55 @@
+package testutil
+
+type Op struct {
```
eahydra, line 3: Actually we don't need options here? we can just export these fields in mockDevice..

gyuho: removed

PR #361, opened 2025-02-05
fix(components/state): clean up update queries, increase unit test coverage, read last API version using SQLite rowid column

gyuho:

### DIFF
--- a/components/state/api_version.go
+++ b/components/state/api_version.go
@@ -3,9 +3,11 @@ package state
 import (
 	"context"
 	"database/sql"
+	"errors"
 	"fmt"
 	"time"
 
+	"github.com/leptonai/gpud/log"
 	"github.com/leptonai/gpud/pkg/sqlite"
 )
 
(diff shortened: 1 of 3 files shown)


components/state/api_version.go
```diff
@@ -36,20 +36,26 @@ func UpdateAPIVersionIfNotExists(ctx context.Context, db *sql.DB, apiVersion str
 }
 
 func UpdateAPIVersion(ctx context.Context, db *sql.DB, apiVersion string) error {
-	query := fmt.Sprintf(`
-INSERT OR REPLACE INTO %s (%s) VALUES (?)
-`, TableNameAPIVersion, ColumnAPIVersion)
+	// First delete any existing version
+	query := fmt.Sprintf(`DELETE FROM %s`, TableNameAPIVersion)
```
xiang90, line 40: what is the issue with `insert or replace`?

components/state/state.go
```diff
@@ -115,73 +115,73 @@ INSERT OR REPLACE INTO %s (%s, %s) VALUES (?, ?);
 
 func GetLoginInfo(ctx context.Context, dbRO *sql.DB, machineID string) (string, error) {
 	query := fmt.Sprintf(`
-SELECT %s FROM %s WHERE %s = '%s'
+SELECT COALESCE(%s, '') FROM %s WHERE %s = ?
```
xiang90, line 118: why?

components/state/state.go
```diff
@@ -115,73 +115,73 @@ INSERT OR REPLACE INTO %s (%s, %s) VALUES (?, ?);
...
+	if err == sql.ErrNoRows {
+		return "", nil
+	}
 	return token, err
 }
 
 func UpdateLoginInfo(ctx context.Context, dbRW *sql.DB, machineID string, token string) error {
 	query := fmt.Sprintf(`
-UPDATE %s SET %s = '%s' WHERE %s = '%s';
+UPDATE %s SET %s = ? WHERE %s = ?;
```
xiang90, line 139: why?

components/state/api_version.go
gyuho: there was some bug when reading this value, fixed the other way, PTAL.

components/state/state.go
gyuho: Removed, no need to handle this this way

components/state/state.go
gyuho: reverted

gyuho:
@cardyok @xiang90 there was some bug in how we read api version (though wasn't used anywhere), fixed

xiang90:
defer to @cardyok

cardyok (approved):
lgtm after fixing the ci

PR #370, opened 2025-02-07
feat(infiniband): use events db to persist ibstat check thresholds

gyuho:
- **test(events): test purge disable**
- **feat(infiniband): use events db to persist ibstat check thresholds**

### DIFF
--- a/components/accelerator/nvidia/infiniband/threshold/threshold.go
+++ b/components/accelerator/nvidia/infiniband/threshold/threshold.go
@@ -1,0 +1,122 @@
+// Package threshold contains the threshold for the infiniband component.
+package threshold
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"sync"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/components/common"
+	events_store "github.com/leptonai/gpud/components/db"
+)
+
+// Defines the minimum number of ports and the expected rate in Gb/sec.
+type Threshold struct {
+	// The minimum number of ports.
+	// If not set, it defaults to the number of GPUs.
+	AtLeastPorts int `json:"at_least_ports"`
+
+	// The expected rate in Gb/sec.
+	// If not set, it defaults to 200.
+	AtLeastRate int `json:"at_least_rate"`
+}
+
+func (th Threshold) JSON() ([]byte, error) {
+	return json.Marshal(th)
+}
+
+func Parse(b []byte) (*Threshold, error) {
+	var th Threshold
+	err := json.Unmarshal(b, &th)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


components/accelerator/nvidia/infiniband/threshold/threshold.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if ev == nil {
+		return nil, nil
+	}
+	return Parse([]byte(ev.ExtraInfo["data"]))
+}
+
+// Updates the current threshold, if and only if the current threshold is not found
+// or the new threshold is different from the current threshold.
+func (s *Store) CompareAndSet(ctx context.Context, threshold Threshold) error {
```
eahydra, line 86: It is better to use the write lock for the entire function.

components/accelerator/nvidia/infiniband/threshold/threshold.go
```diff
@@ -0,0 +1,122 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/components/common"
+	events_store "github.com/leptonai/gpud/components/db"
+)
+
+// Defines the minimum number of ports and the expected rate in Gb/sec.
+type Threshold struct {
```
eahydra, line 19: Do we support syncing node hardware spec like ib threshold from server?  /cc @cardyok

components/accelerator/nvidia/infiniband/threshold/threshold.go
```diff
@@ -0,0 +1,122 @@
...
+func Parse(b []byte) (*Threshold, error) {
+	var th Threshold
+	err := json.Unmarshal(b, &th)
+	if err != nil {
+		return nil, err
+	}
+	return &th, nil
+}
+
+func (th Threshold) Event(time time.Time) (components.Event, error) {
```
eahydra, line 42: I'm curious that why not store these data to a separate file which stores all node hardware spec or node related thresholds?

components/accelerator/nvidia/infiniband/threshold/threshold.go
gyuho: Yes

https://github.com/leptonai/gpud/blob/9fe57e679884c752af3328d1b60ee0c2729eae03/internal/session/serve.go#L170-L183

(Current issue is this update will reset on GPUd restart)

components/accelerator/nvidia/infiniband/threshold/threshold.go
gyuho: Yes, we can do that as well (separate table might be easier, since we already have all the persistence via sqlite file)

eahydra: OK, but it would be better to use a text file to store this configuration, as it would be more maintainable.

gyuho:
Closing for now

PR #371, opened 2025-02-07
feat(infiniband): do not auto-detect ibstat threshold (disable by default), simplify poller, discard parse error to use last known ibstat status

gyuho:


gyuho:
@eahydra @cardyok We need these first

- https://github.com/leptonai/gpud/pull/377
- https://github.com/leptonai/gpud/pull/378

components/accelerator/nvidia/query/infiniband/infiniband.go
```diff
@@ -12,24 +12,26 @@ import (
...
 func SupportsInfinibandPortRate(gpuProductName string) int {
 	p := strings.ToLower(gpuProductName)
+
 	if strings.Contains(p, "a100") {
 		return 200
 	}
 	if strings.Contains(p, "h100") {
 		return 400
 	}
+	if strings.Contains(p, "b100") {
```
xiang90, line 25: revert

gyuho:
Will also simplify the data fetch logic (remove all complicated goroutines)

gyuho:
Feedback will be addressed in a new PR

PR #373, opened 2025-02-08
fix(xid/sxid): rely on last reboot first

### LINKED ISSUE #372: Improvement: always get the last boot time from the local machine when possible
We should rely on the local machine information as much as possible. If we cannot find things, we may want to fallback to our own record in sqlite.

See

```
last reboot
reboot   system boot  5.15.0-78-generi Tue Feb  4 17:54   still running
reboot   system boot  5.15.0-78-generi Wed Jan 29 10:38 - 17:45 (6+07:06)
reboot   system boot  5.15.0-78-generi Wed Jan 29 08:14 - 17:45 (6+09:30)
reboot   system boot  5.15.0-78-generi Tue Jan 28 14:46 - 17:45 (7+02:58)
reboot   system boot  5.15.0-78-generi Tue Jan 28 14:30 - 17:45 (7+03:14)
reboot   system boot  5.15.0-78-generi Tue Jan 28 13:55 - 17:45 (7+03:50)
reboot   system boot  5.15.0-78-generi Sat Sep 28 03:44 - 17:45 (129+14:00)
reboot   system boot  5.15.0-78-generi Fri Sep 27 07:33 - 03:39  (20:06)
reboot   system boot  5.15.0-78-generi Fri Sep 27 07:28 - 07:30  (00:01)
reboot   system boot  5.15.0-78-generi Fri Sep 27 06:56 - 07:25  (00:29)
reboot   system boot  5.15.0-78-generi Fri Sep 27 06:52 - 06:53  (00:01)
reboot   system boot  5.15.0-78-generi Tue Sep 24 04:14 - 06:49 (3+02:35)
```

cardyok:
fix #372

### DIFF
--- a/components/accelerator/nvidia/error/xid/component.go
+++ b/components/accelerator/nvidia/error/xid/component.go
@@ -203,23 +203,31 @@ func (c *XIDComponent) SetHealthy() error {
 }
 
 func (c *XIDComponent) updateCurrentState() error {
-	osComponent, err := components.GetComponent(os_id.Name)
+	rebootEvents, err := getRebootEvents(c.rootCtx)
 	if err != nil {
-		return fmt.Errorf("failed to get os component: %w", err)
-	}
-	osEvents, err := osComponent.Events(c.rootCtx, time.Now().Add(-DefaultRetentionPeriod))
-	if err != nil {
-		return fmt.Errorf("failed to get os events: %w", err)
+		return fmt.Errorf("failed to get reboot events: %w", err)
 	}
 	localEvents, err := c.store.Get(c.rootCtx, time.Now().Add(-DefaultRetentionPeriod))
 	if err != nil {
 		return fmt.Errorf("failed to get all events: %w", err)
 	}
-	events := mergeEvents(osEvents, localEvents)
+	events := mergeEvents(rebootEvents, localEvents)
 	c.mu.Lock()
 	c.currState = EvolveHealthyState(events)
 	c.mu.Unlock()
 	return nil
+}
+
+func getRebootEvents(ctx context.Context) ([]components.Event, error) {
+	osComponent, err := components.GetComponent(os_id.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get os component: %w", err)
+	}
+	osEvents, err := osComponent.Events(ctx, time.Now().Add(-DefaultRetentionPeriod))
+	if err != nil {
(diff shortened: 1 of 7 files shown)


components/accelerator/nvidia/error/xid/component.go
```diff
@@ -203,25 +205,46 @@ func (c *XIDComponent) SetHealthy() error {
...
 	if err != nil {
-		return fmt.Errorf("failed to get os events: %w", err)
+		return fmt.Errorf("failed to get reboot events: %w", err)
 	}
 	localEvents, err := c.store.Get(c.rootCtx, time.Time{})
 	if err != nil {
 		return fmt.Errorf("failed to get all events: %w", err)
 	}
-	events := mergeEvents(osEvents, localEvents)
+	events := mergeEvents(rebootEvents, localEvents)
```
gyuho, line 216: Make sure `localEvents` is not nil`?

https://github.com/leptonai/gpud/blob/ab3785a997a0feb116f4ed3acab28afdb3521300/components/db/events.go#L352-L355

`Get` returns `nil` if not found

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -203,25 +205,46 @@ func (c *XIDComponent) SetHealthy() error {
...
 	c.mu.Lock()
 	c.currState = EvolveHealthyState(events)
 	c.mu.Unlock()
 	return nil
 }
 
+func getRebootEvents(ctx context.Context) ([]components.Event, error) {
```
gyuho, line 223: This seems useful in a lot of places, should we put it in `gpud/tree/main/components/common`?

cardyok: yes we can, did not find a proper file location to put it tho..prob need a `gpud/components/utils`

components/accelerator/nvidia/error/xid/component.go
cardyok: actually it can be nil here.. localEvents nil is acceptable to mergeEvents

components/accelerator/nvidia/error/xid/component.go
gyuho: yeah either way is fine :)

cardyok:
based on most recent discussion, i think this one can be dropped, will change the code

eahydra (approved):
lgtm

Issue #379, opened 2025-02-09
scan: support IB checking.


xiang90:
User would explicitly specify if they want to do IB checking. The default should be to skip since we do not know if the machine is expected to be ib enabled.

User should only give provide one flag ( --check-ib ). We should:

1. detect the gpu family (a100, h100, h200, b200, etc.)
2. based on the gpu family, find the expected number of ib devices and rates (we need a table based on nvidia reference architecture.)
3. use ibstats to get the data and verify

eahydra:
@gyuho Let’s make the PR title shorter.. We can add detailed description or background in the commit message body..

gyuho:
Done via https://github.com/leptonai/gpud/pull/383

PR #378, opened 2025-02-09
feat(ib): remove GPU product for "ibstat" checks, remove auto-detect "ibstat" threshold

gyuho:
- Remove GPU product name checks in "ibstat"
- Remove default "ibstat" threshold based on GPU product name

PR #380, opened 2025-02-09
feat(gpud): "scan --check-ib", move port/rates to "infiniband" package

gyuho:
Address https://github.com/leptonai/gpud/issues/379.

Base branch https://github.com/leptonai/gpud/pull/378.

### DIFF
--- a/components/accelerator/nvidia/infiniband/component_output.go
+++ b/components/accelerator/nvidia/infiniband/component_output.go
@@ -101,7 +101,7 @@ var (
 
 // Returns the output evaluation reason and its healthy-ness.
 // We DO NOT auto-detect infiniband devices/PCI buses, strictly rely on the user-specified config.
-func (o *Output) Evaluate(cfg ExpectedPortStates) (string, bool, error) {
+func (o *Output) Evaluate(cfg infiniband.ExpectedPortStates) (string, bool, error) {
 	// nothing specified for this machine, gpud MUST skip the ib check
 	if cfg.AtLeastPorts == 0 && cfg.AtLeastRate == 0 {
 		return msgMustSetPortsOrRate, true, nil
--- a/components/accelerator/nvidia/query/infiniband/default.go
+++ b/components/accelerator/nvidia/query/infiniband/default.go
@@ -1,0 +1,50 @@
+package infiniband
+
+import "strings"
+
+// Configures the expected state of the ports.
+type ExpectedPortStates struct {
+	// The minimum number of ports.
+	// If not set, it defaults to the number of GPUs.
+	AtLeastPorts int `json:"at_least_ports"`
+
+	// The expected rate in Gb/sec.
+	// If not set, it defaults to 200.
+	AtLeastRate int `json:"at_least_rate"`
+}
+
+var gpuPortConfigs = map[string]ExpectedPortStates{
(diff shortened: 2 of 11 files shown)


components/diagnose/scan.go
```diff
@@ -214,6 +215,31 @@ func Scan(ctx context.Context, opts ...OpOption) error {
 					}
 				}
 			}
+
+			if op.checkIb {
+				cctx, cancel := context.WithTimeout(ctx, 15*time.Second)
```
eahydra, line 220: The function is too long.. could you wrap these codes as a function?

eahydra: Are these codes duplicated in `accelerator/nvidia/infiniband`? we're going to directly call `ibstat` to detect if there is any ib issue. I think these codes are similar, but different rate configuration..

gyuho: Yes, will be similar, the difference is that, the `accelerator/nvidia/infiniband` component code will set the state to unhealthy

eahydra: Yeah, so it would be better to implement it somewhere that can be reused by `accelerator/nvidia/infiniband` or just put it in `accelerator/nvidia/infiniband` if possible.

components/accelerator/nvidia/query/infiniband/default.go
```diff
@@ -0,0 +1,47 @@
...
+	// "InfiniBand (default): Up to 400Gbps"
+	// ref. https://docs.nvidia.com/dgx/dgxh100-user-guide/introduction-to-dgxh100.html
+	// ref. https://docs.nvidia.com/launchpad/ai/h100-mig/latest/h100-mig-gpu.html
+	"h100": {AtLeastPorts: 8, AtLeastRate: 400},
+	"h200": {AtLeastPorts: 8, AtLeastRate: 400},
+
+	// "HGX B200 supports networking speeds up to 400Gb/s through the NVIDIA Quantum-2 InfiniBand"
+	// ref. https://nvidianews.nvidia.com/news/nvidia-blackwell-platform-arrives-to-power-a-new-era-of-computing
+	"b100": {AtLeastPorts: 8, AtLeastRate: 400},
+	"b200": {AtLeastPorts: 8, AtLeastRate: 400},
```
xiang90, line 31: need to add gb. there is no more b100z

xiang90: let us use actual doc as reference not news

components/accelerator/nvidia/infiniband/component_output.go
```diff
@@ -103,7 +103,7 @@ var (
 // We DO NOT auto-detect infiniband devices/PCI buses, strictly rely on the user-specified config.
 func (o *Output) Evaluate(cfg infiniband.ExpectedPortStates) (string, bool, error) {
 	// nothing specified for this machine, gpud MUST skip the ib check
-	if cfg.AtLeastPorts == 0 && cfg.AtLeastRate == 0 {
+	if cfg.AtLeastPorts == 0 && cfg.AtLeastRateGb == 0 {
```
xiang90, line 106: No... Keep AtLeastRate as it is. GB is the standard Nvidia B. check out gb200 nvl72.

components/diagnose/scan.go
gyuho: done

xiang90:
defer to @eahydra after fixing the GB thing.

PR #381, opened 2025-02-09
feat(infiniband): directly call ibstat for ib states

gyuho:
Replace https://github.com/leptonai/gpud/pull/371.


xiang90:
this is great. defer to @cardyok / @eahydra

cardyok (commented):
overall lgtm, one quick question, seems now we will enable infiniband component on all machines that have nvidia-smi command. shall we keep this behavior?

gyuho:
@cardyok Only enabled (no change in that logic):

https://github.com/leptonai/gpud/blob/31b17f9dc3a15d77a18255c384f4496c260ed7ea/config/default.go#L245-L248

https://github.com/leptonai/gpud/blob/31b17f9dc3a15d77a18255c384f4496c260ed7ea/config/default.go#L310

cardyok (approved):
approved, we still need to fix the false negative issue (when infiniband issue is real, restarting gpud will temporarily incorrectly detect as Healthy) in the future

PR #390, opened 2025-02-10
fix(xid/sxid): only consider 3 day events and do not rely on purge

cardyok:


eahydra:
We fix it according to the 3-day retention logic, but there are still some cases that cannot be solved. For example, an xid critical error occurred 4 days ago, and it has not been processed for some reason. After 4 days, the xid should still be in the unhealthy/degrade state. However, if it is processed according to the 3-day retention, it will become healthy, which is contrary to the fact. /cc @xiang90 @gyuho

components/accelerator/nvidia/error/xid/component.go
```diff
@@ -142,7 +142,7 @@ func (c *XIDComponent) start(watcher pkg_dmesg.Watcher, updatePeriod time.Durati
 				log.Logger.Errorw("failed to create event", "error", err)
 				continue
 			}
-			events, err := c.store.Get(c.rootCtx, time.Time{})
+			events, err := c.store.Get(c.rootCtx, time.Now().Add(-DefaultRetentionPeriod))
```
eahydra, line 145: Let's mock the db.Store interface to do more tests.

eahydra (approved):

PR #392, opened 2025-02-11
debug(infiniband): output raw ibstat when issue found

gyuho:


components/accelerator/nvidia/infiniband/component.go
```diff
@@ -79,6 +80,8 @@ func (c *component) getStates(ctx context.Context, thresholds infiniband.Expecte
 				"potential infiniband switch/hardware issue needs immediate attention",
 			},
 		}
+
+		fmt.Printf("[DEBUG] ibstat issue found!\n\n%s\n\n", string(o.Raw))
```
eahydra, line 84: We should use logger?

eahydra (approved):
lgtm

PR #395, opened 2025-02-11
feat(ib, disk): use combined output for ibstat, lsblk

gyuho:
c.f., https://github.com/leptonai/gpud/pull/393


components/accelerator/nvidia/infiniband/component.go
```diff
@@ -58,7 +59,24 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 }
 
 func (c *component) getStates(ctx context.Context, thresholds infiniband.ExpectedPortStates) ([]components.State, error) {
-	o, err := infiniband.GetIbstatOutput(ctx, []string{c.toolOverwrites.IbstatCommand})
+	var o *infiniband.IbstatOutput
+	var err error
+	for i := 0; i < 3; i++ {
+		o, err = infiniband.GetIbstatOutput(ctx, []string{c.toolOverwrites.IbstatCommand})
+		if errors.Is(err, infiniband.ErrIbstatOutputEmpty) {
```
ccding, line 66: do shorter branches first. for example
```go
for i := 0; i < 3; i++ {
  o, err = infiniband.GetIbstatOutput(ctx, []string{c.toolOverwrites.IbstatCommand})
  if err != nil || !errors.Is(err, infiniband.ErrIbstatOutputEmpty) {
      break
  }
  log.Logger.Warnw("ibstat output is empty, retrying", "attempt", i+1, "error", err)
  select {
  case <-ctx.Done():
      return nil, ctx.Err()
  case <-time.After(300 * time.Millisecond):
  }
}

File: components/accelerator/nvidia/infiniband/component.go [hunk 2]
```diff
@@ -58,7 +59,20 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
...
 func (c *component) getStates(ctx context.Context, thresholds infiniband.ExpectedPortStates) ([]components.State, error) {
-	o, err := infiniband.GetIbstatOutput(ctx, []string{c.toolOverwrites.IbstatCommand})
+	var o *infiniband.IbstatOutput
+	var err error
+	for i := 0; i < 3; i++ {
+		o, err = infiniband.GetIbstatOutput(ctx, []string{c.toolOverwrites.IbstatCommand})
+		if err == nil || !errors.Is(err, infiniband.ErrIbstatOutputEmpty) {
+			break
+		}
+		log.Logger.Warnw("ibstat output is empty, retrying", "attempt", i+1, "error", err)
```
eahydra, line 69: It doesn't make sense unless we can figure out the root cause of why we got empty output. If we add a retry here, similar scenarios should follow the same logic, but that would be too strange.

gyuho: removed

eahydra (approved):
lgtm

PR #398, opened 2025-02-13
fix(os): log which command fails when context timeout in get

gyuho:

### DIFF
--- a/components/os/component_output.go
+++ b/components/os/component_output.go
@@ -426,7 +426,7 @@ func setDefaultPoller(cfg Config, eventsStore events_db.Store) {
 		defaultPoller = query.New(
 			os_id.Name,
 			cfg.Query,
-			CreateGet(cfg, eventsStore),
+			createGet(cfg, eventsStore, pkg_host.SystemdDetectVirt),
 			nil,
 		)
 	})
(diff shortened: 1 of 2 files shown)


components/os/component_output.go
```diff
@@ -436,7 +436,9 @@ func getDefaultPoller() query.Poller {
 	return defaultPoller
 }
 
-func CreateGet(cfg Config, eventsStore events_db.Store) func(ctx context.Context) (_ any, e error) {
+type getSystemdDetectVirtFunc func(ctx context.Context) (pkg_host.VirtualizationEnvironment, error)
+
+func createGet(cfg Config, eventsStore events_db.Store, getSystemdDetectVirt getSystemdDetectVirtFunc) func(ctx context.Context) (_ any, e error) {
```
eahydra, line 441: We don't need to refactor the function prototype.  If we want to test it, we can do as following:
```
var getSystemdDetectVirtFunc = pkg_host.SystemdDetectVirt

func createGet(.........) {
   // .... something
  virtEnv, err := getSystemdDetectVirtFunc()
  // blablabla....
}

func Test_createGet(t *testing.T) {
    defer func() {
       getSystemdDetectVirtFunc = pkg_host.SystemdDetectVirt
    }() 
   getSystemdDetectVirtFunc = func() (bool, error) {
        return false, context.DeadlineExceeded
    }
}

```

components/os/component_output.go
```diff
@@ -448,11 +450,11 @@ func CreateGet(cfg Config, eventsStore events_db.Store) func(ctx context.Context
 
 		o := &Output{}
 
-		cctx, ccancel := context.WithTimeout(ctx, 10*time.Second)
-		virtEnv, err := pkg_host.SystemdDetectVirt(cctx)
```
eahydra, line 452: I read the function `pkg_host.SystemdDetectVirt`, it used Process. Should we append command into error message or log the command when getting error?

components/os/component_output.go
gyuho: done

components/os/component_output.go
gyuho: done

eahydra (approved):
lgtm

PR #399, opened 2025-02-13
feat(dmesg): dedup logs by seconds if same content, bump up log channel buffer

gyuho:
- Deduplicate log lines if same second, same content
- Bump up dmesg log watcher channel buffer to 3,000

### DIFF
--- a/pkg/dmesg/testdata/dmesg.peermem.log.0
+++ b/pkg/dmesg/testdata/dmesg.peermem.log.0
@@ -1,0 +1,8 @@
+kern  :err   : 2025-02-10T16:28:06,502716+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,514050+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,525389+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,536726+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,548051+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,559449+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,570939+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:07,570939+00:00 test
--- a/pkg/dmesg/watcher.go
+++ b/pkg/dmesg/watcher.go
@@ -53,13 +53,18 @@ func NewWatcher() (Watcher, error) {
 	return NewWatcherWithCommands(DefaultWatchCommands)
 }
 
+const (
+	DefaultCacheExpiration    = 5 * time.Minute
+	DefaultCachePurgeInterval = 10 * time.Minute
+)
+
 func NewWatcherWithCommands(cmds [][]string) (Watcher, error) {
 	if len(cmds) == 0 {
 		return nil, errors.New("no commands provided")
 	}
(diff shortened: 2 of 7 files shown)


pkg/dmesg/testdata/dmesg.peermem.log.0
```diff
@@ -0,0 +1,7 @@
+kern  :err   : 2025-02-10T16:28:06,502716+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,514050+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,525389+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,536726+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,548051+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,559449+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+kern  :err   : 2025-02-10T16:28:06,570939+00:00 nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
```
eahydra, line 7: We should add some different logs here, you can check if we get these logs

pkg/dmesg/watcher.go
```diff
@@ -221,3 +241,34 @@ func findISOTimestampIndex(s string) int {
...
+
+// isCached returns true if the log line is already in the cache.
+func (d *deduper) isCached(l LogLine) bool {
+	_, found := d.cache.Get(l.cacheKey())
+	return found
+}
+
+// addCache adds the log line to the cache.
+func (d *deduper) addCache(l LogLine) {
+	d.cache.Set(l.cacheKey(), struct{}{}, cache.DefaultExpiration)
```
eahydra, line 273: we can record the times to the value, and print it when evicting.

pkg/dmesg/watcher_test.go
```diff
@@ -622,3 +681,431 @@ func TestReadContextCancellation(t *testing.T) {
 		t.Error("timeout waiting for channel to close after context cancellation")
 	}
 }
+
+func TestDeduper(t *testing.T) {
+	t.Run("new deduper should create cache with correct expiration", func(t *testing.T) {
+		d := newDeduper(5*time.Minute, 10*time.Minute)
+		assert.NotNil(t, d)
+		assert.NotNil(t, d.cache)
+	})
```
eahydra, line 690: actually we can remove it.. the following cases can cover it.

pkg/dmesg/watcher.go
```diff
@@ -83,6 +91,8 @@ func (w *watcher) Close() {
 func watch(
 	ctx context.Context,
 	cmds [][]string,
+	cacheExpiration time.Duration,
+	cachePurgeInterval time.Duration,
 ) (<-chan LogLine, error) {
 	ch := make(chan LogLine, 1000)
```
eahydra, line 97: Should we need to extend the channel buffer size?

pkg/dmesg/watcher.go
```diff
@@ -221,3 +241,34 @@ func findISOTimestampIndex(s string) int {
 	}
 	return loc[0]
 }
+
+func (l LogLine) cacheKey() string {
+	h := sha256.New()
```
eahydra, line 246: Do we need to use sha256 here?  Perhaps we should a simple hash function like fnv-64 or the simplest way is use fmt.Sprintf("%d-%s", l.Timestamp.Unix(), l.Content)

gyuho: Actually go-cache already hash map... let me just raw string https://github.com/patrickmn/go-cache/blob/46f407853014144407b6c2ec7ccc76bf67958d93/cache.go#L42

pkg/dmesg/testdata/dmesg.peermem.log.0
gyuho: done

pkg/dmesg/watcher.go
gyuho: done

pkg/dmesg/watcher.go
gyuho: done

eahydra (approved):
lgtm

PR #400, opened 2025-02-13
feat(components/fd): use new dmesg poller (simplified)

gyuho:


components/fd/dmesg_matcher.go
```diff
@@ -11,7 +11,9 @@ const (
 	//
 	// ref.
 	// https://docs.kernel.org/admin-guide/sysctl/fs.html#file-max-file-nr
-	RegexVFSFileMaxLimitReached = `VFS: file-max limit \d+ reached`
+	EventVFSFileMaxLimitReached   = "error_vfs_file_max_limit_reached"
```
eahydra, line 14: shall we remove the error prefix? We have the Event.Type.

components/fd/dmesg_matcher.go
```diff
@@ -11,7 +11,9 @@ const (
...
+	RegexVFSFileMaxLimitReached   = `VFS: file-max limit \d+ reached`
```
eahydra, line 15: nit, if it's possible, we should not export these constant values and the functions below.

components/fd/watcher.go
```diff
@@ -0,0 +1,101 @@
...
+			} else {
+				log.Logger.Infow("successfully inserted event", "event", ev.Name)
+			}
+		}
+	}
+}
+
+func (w *watcher) close() {
+	w.closeOnce.Do(func() {
+		w.dmesgWatcher.Close()
```
eahydra, line 99: we should make the dmesgWatcher.Close close-safe... should add closeOnce to dmesgWatcher.

eahydra (commented):
overall lgtm

components/fd/dmesg_matcher.go
gyuho: @cardyok Please confirm if it's ok

components/fd/watcher.go
gyuho: fixed and others fixed in https://github.com/leptonai/gpud/pull/404

components/fd/dmesg_matcher.go
gyuho: done

components/fd/dmesg_matcher.go
cardyok: yea its is ok

gyuho:
@eahydra @xiang90 codecov failures due to this deleting code, once we clean up, we will increase the test coverage for the new code base

eahydra (approved):
lgtm

PR #402, opened 2025-02-13
nits(components/memory): set missing message field in events, update "New"

gyuho:


eahydra (approved):
lgtm

PR #403, opened 2025-02-13
feat(nvidia/peermem): use new dmesg poller (simplified)

gyuho:


components/accelerator/nvidia/peermem/watcher.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+type watcher struct {
+	ctx          context.Context
+	eventsStore  events_db.Store
+	dmesgWatcher pkg_dmesg.Watcher
+}
+
+func newWatcher(ctx context.Context, eventsStore events_db.Store) (*watcher, error) {
```
eahydra, line 22: LOL.. These watchers are same.. Later we can make the watcher as a helper in the dmesg package, renamed to WatcherPipeline?  Caller just needs to pass eventdb store and matcher.

components/accelerator/nvidia/peermem/dmesg_matcher.go
```diff
@@ -0,0 +1,46 @@
...
+import "regexp"
+
+const (
+	// repeated messages may indicate more persistent issue on the inter-GPU communication
+	// e.g.,
+	// [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+	// [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+	// [Thu Sep 19 02:29:46 2024] nvidia-peermem nv_get_p2p_free_callback:127 ERROR detected invalid context, skipping further processing
+	eventPeermemInvalidContext   = "nvidia_peermem_invalid_context"
+	regexPeermemInvalidContext   = `.*ERROR detected invalid context, skipping further processing`
```
eahydra, line 12: I'm not sure, are there any peermem logs that don't have the nvidia-peermem prefix? If not, shall we add this prefix in the regex?

components/accelerator/nvidia/peermem/component.go
```diff
@@ -5,42 +5,60 @@ package peermem
...
+
 	if nvidia_query.GetDefaultPoller() == nil {
+		ccancel()
 		return nil, nvidia_query.ErrDefaultPollerNotSet
 	}
 
 	cfg.Query.SetDefaultsIfNotSet()
-
-	cctx, ccancel := context.WithCancel(ctx)
 	nvidia_query.GetDefaultPoller().Start(cctx, cfg.Query, nvidia_peermem_id.Name)
```
eahydra, line 43: I think we could remove it?

xiang90:
> Attention: Patch coverage is 47.00000% with 53 lines in your changes missing coverage. Please review.

Can we improve

components/accelerator/nvidia/peermem/watcher.go
gyuho: @eahydra Yeah, please https://github.com/leptonai/gpud/pull/406 review for this.

components/accelerator/nvidia/peermem/component.go
gyuho: commented... still need it for lsmod polling. We will deprecate this in the following PRs

components/accelerator/nvidia/peermem/dmesg_matcher.go
gyuho: So far, we only seen this specific pattern. If we find new pattern, we can add it later.

eahydra (approved):
lgtm

PR #404, opened 2025-02-13
nits(components): close dmesg watcher without sync.Once

gyuho:
Already thread-safe, as its underlying watcher simplify calls conext
cancel func


eahydra (approved):
lgtm

PR #406, opened 2025-02-13
feat(dmesg): log line processor for dmesg

gyuho:

### DIFF
--- a/pkg/dmesg/log_line_processor.go
+++ b/pkg/dmesg/log_line_processor.go
@@ -1,0 +1,103 @@
+package dmesg
+
+import (
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/leptonai/gpud/components"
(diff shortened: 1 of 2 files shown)


pkg/dmesg/events_watcher.go
```diff
@@ -0,0 +1,103 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/pkg/common"
+	events_db "github.com/leptonai/gpud/pkg/events-db"
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+type eventsWatcher struct {
```
eahydra, line 15: We already have a type named `Watcher`, so we should not call the new type `eventsWatcher`. We can call it `LogLineProcessor`, `LogEventProcessor`, `LogLineSink`?

gyuho: @eahydra Renamed

PR #407, opened 2025-02-14
 feat(nvidia): remove nvidia-smi fallback for row remapping issues

gyuho:
- **feat(nvidia): remove nvidia-smi fallback for row remapping issues**


eahydra (approved):
lgtm

PR #414, opened 2025-02-14
fix(infiniband): do not call "ibstat" on empty thresholds

gyuho:
Fix the case:
1. thresholds empty (0 and 0)
2. call "ibstat" that returns an empty output

"2" should not happen when the thresholds are not set.


eahydra (approved):
lgtm

PR #419, opened 2025-02-14
feat(nvidia/nccl): use new dmesg log line processor (simplified)

gyuho:


components/accelerator/nvidia/nccl/component.go
```diff
@@ -24,28 +25,29 @@ func New(ctx context.Context, cfg nvidia_common.Config) (components.Component, e
...
-	cctx, ccancel := context.WithCancel(ctx)
-	w, err := newWatcher(cctx, eventsStore)
+	w, err := dmesg.NewWatcher()
 	if err != nil {
-		ccancel()
 		return nil, err
 	}
 
+	cctx, ccancel := context.WithCancel(ctx)
+	logLineProcessor := dmesg.NewLogLineProcessor(cctx, w, Match, eventsStore)
```
eahydra, line 34: Sorry for forgetting to left comments,  shall we create the NewWatcher in the LogLineProcessor if the w is nil?

gyuho: Oh yes, let's do that

gyuho: Done https://github.com/leptonai/gpud/pull/428

eahydra (approved):
lgtm

Issue #409, opened 2025-02-14
cannot check ib state


eahydra:
```
{"level":"error","ts":"2025-02-14T10:04:22Z","caller":"log/log.go:60","msg":"failed to invoke component state","operation":"GetStates","component":"accelerator-nvidia-infiniband","error":"ibstat not found. cannot check ib state","stacktrace":"github.com/leptonai/gpud/log.(*LeptonLogger).Errorw\n\t/home/runner/work/gpud/gpud/log/log.go:60\ngithub.com/leptonai/gpud/internal/session.(*Session).getStates\n\t/home/runner/work/gpud/gpud/internal/session/serve.go:358\ngithub.com/leptonai/gpud/internal/session.(*Session).serve\n\t/home/runner/work/gpud/gpud/internal/session/serve.go:80"}
```

We should check if the configuration existed, if not, just return.

PR #420, opened 2025-02-14
fix(infiniband): set /states unhealthy when thresholds set but ibstat not found

gyuho:
Fix https://github.com/leptonai/gpud/issues/409.


eahydra (approved):
lgtm

PR #423, opened 2025-02-14
project(pkg*): rename more gpud specific packages

gyuho:


gyuho:
let's revisit this

PR #424, opened 2025-02-14
feat(gpud update): trim space when fetching version from "_latest.txt"

gyuho:
Fix

> {"level":"info","ts":"2025-02-11T17:57:17-08:00","caller":"update/update.go:206","msg":"starting gpud update","version":"v0.4.2\n","url":"https://pkg.gpud.dev/","requireRoot":true,"useSystemd":true}
{"level":"info","ts":"2025-02-11T17:57:18-08:00","caller":"update/update.go:34","msg":"Downloading \"https://pkg.gpud.dev/gpud_v0.4.2%0A_linux_amd64_ubuntu24.04.tgz\""}

### DIFF
--- a/cmd/gpud/command/update.go
+++ b/cmd/gpud/command/update.go
@@ -5,6 +5,7 @@
 package command
 
 import (
+	"bytes"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/gpud/command/update.go
```diff
@@ -43,7 +43,7 @@ func detectLatestVersion() (string, error) {
 		return "", err
 	}
 
-	fetchedVer, err := distsign.Fetch(defaultURLPrefix+track+"_latest.txt", 100)
+	fetchedVer, err := distsign.FetchAndTrimSpace(defaultURLPrefix+track+"_latest.txt", 100)
```
eahydra, line 46: it's a bit weird... we should not rely on distsign package here.. we can write this function in the update.go, should call the function `fetchVersion`

gyuho: done

cmd/gpud/command/update.go
```diff
@@ -83,3 +82,20 @@ func cmdUpdateCheck(cliContext *cli.Context) error {
 	fmt.Printf("latest version: %s\n", ver)
 	return nil
 }
+
+func fetchVersion(track string) (string, error) {
```
eahydra, line 86: I re-read the update-related codes and I feel that my previous suggestions were not very good, I'm sorry for wasting your time.
I have two questions:
1. Why do we have a situation like `v0.4.3\n`? Do we have a bug somewhere else that causes this problem?
2. If Trim is required, can it be modified in `detectLatestVersion`? It should be enough to just trim the fetchedVer variable once.

gyuho: > Why do we have a situation like v0.4.2\n

Sometimes, release owner uploaded the unstable_latest.txt file with trailing whitespaces (as it's not fully automated yet).

> can it be modified in detectLatestVersion

Let me check

eahydra (approved):
lgtm

eahydra:
```
codecov/project Failing after 1s — 24.12% (-0.06%) compared to 396e407
```
Please add UT to cover the new codes.

PR #426, opened 2025-02-14
fix(gpud-state): do not rollback if apiversion update committed successfully

gyuho:


eahydra (approved):
lgtm

PR #427, opened 2025-02-15
feat(systemd): do not return error on uptime check failures

gyuho:
Just log as error

### DIFF
--- a/components/systemd/component_output.go
+++ b/components/systemd/component_output.go
@@ -224,7 +224,7 @@ func ConnectDbus() error {
 			defaultDbusConn, err = systemd.NewDbusConn(ctx)
 			if err != nil {
 				defaultDbusConnCancel()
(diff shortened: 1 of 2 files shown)


components/systemd/component_output.go
```diff
@@ -188,6 +183,12 @@ func CreateGet(cfg Config) query.GetFunc {
 
 			uptimeSeconds := int64(0)
 			uptimeDescription := "n/a"
+
+			uptime, err := systemd.GetUptime(unit)
+			if err != nil {
```
eahydra, line 188: We can consider correcting the same errors that have occurred before, if there are not too many. 
In addition, we need to find out the root cause.

eahydra: Like line 227 in that file, we should change the logger level to error.

eahydra (approved):
lgtm

Issue #391, opened 2025-02-11
Improve XID/SXID suggest action


eahydra:
Some XID/SXID issues have two actions. For example, XID 94:
https://github.com/leptonai/gpud/blob/7e9003bd94ea975a37348cea712f3cab0aafbeda/components/accelerator/nvidia/query/xid/xid.go#L3493-L3496

We should improve the XID/SXID-related suggested actions to help users resolve issues more effectively.
1. Only provide one suggestion based on the current state. For example, when we first encounter XID 94, we should suggest that the user reboot the system. If XID 94 persists after two reboots, we should then suggest a HARDWARE_INSPECTION.
2. In my opinion, we have collected too many references to help users understand what XID/SXID is, even though users may not fully understand them. Perhaps we should reduce these references?

gyuho:
Related discussion: https://github.com/leptonai/gpud/pull/432#discussion_r1958195967

gyuho:
https://github.com/leptonai/gpud/blob/b2d7e85fdd823f6465e6ba148677f38b4b4d8ae5/components/accelerator/nvidia/error/xid/helper.go#L85-L91

PR #432, opened 2025-02-17
feat(dmesg): remove component, use match func diagnose/scan

gyuho:


pkg/diagnose/scan.go
```diff
@@ -234,7 +233,7 @@ func Scan(ctx context.Context, opts ...OpOption) error {
 			query_log_tail.WithDedup(true),
 			query_log_tail.WithCommands(pkg_dmesg.DefaultDmesgScanCommands),
 			query_log_tail.WithLinesToTail(op.lines),
-			query_log_tail.WithSelectFilter(dmesg.DefaultDmesgFiltersForNvidia()...),
+			query_log_tail.WithSelectFilter(defaultDmesgFilters()...),
```
eahydra, line 236: We should refactor the dmesgCheck feature. It's repetitive across other components, even though it composes these components here.

IMO we should call the existing dmesg check components directly or their check functions to determine if the log line is matched, we should hide some logic here ,for example we should use https://github.com/leptonai/gpud/blob/2175d7d3671ccde724a6145ddbb87aa3f36e2cbd/pkg/nvidia-query/xid/dmesg/dmesg.go#L61 instead of line 247..

components/accelerator/nvidia/error/xid/helper.go
```diff
@@ -83,8 +111,7 @@ func EvolveHealthyState(events []components.Event) (ret components.State) {
 	if lastXidErr == nil {
 		reason = "XIDComponent is healthy"
 	} else {
-		xidErrBytes, _ := lastXidErr.JSON()
-		reason = string(xidErrBytes)
+		reason = fmt.Sprintf("xid %d detected by %s", lastXidErr.Xid, lastXidErr.DataSource)
```
gyuho, line 114: @eahydra @cardyok This is where we set the reason that's displayed in the nodes dashboard (prev. the whole JSON structure of

https://github.com/leptonai/gpud/blob/6bc689b15fa8257f61c6156043e0626b4a1ff67e/components/accelerator/nvidia/error/xid/reason.go#L25-L45

now simplified to this one-liner, let me know if you have any suggestions (same to SXid)

gyuho: c.f., https://github.com/leptonai/gpud/issues/391

gyuho: This change has been reverted (same as before) /cc @cardyok

eahydra (approved):
lgtm

PR #436, opened 2025-02-18
feat(dmesg, log): support match func in log scanner

gyuho:
- **feat(dmesg): match func with regex**
- **tail**

### DIFF
--- a/components/accelerator/nvidia/nccl/dmesg_matcher.go
+++ b/components/accelerator/nvidia/nccl/dmesg_matcher.go
@@ -1,18 +1,20 @@
 package nccl
 
-import "regexp"
+import (
+	"regexp"
+)
 
 const (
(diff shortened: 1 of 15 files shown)


components/accelerator/nvidia/nccl/dmesg_matcher.go
```diff
@@ -22,23 +24,24 @@ func HasNCCLSegfaultInLibnccl(line string) bool {
 	return false
 }
 
-func Match(line string) (name string, message string) {
+func Match(line string) (eventName string, regex string, message string) {
```
eahydra, line 27: let's not return the regex unless it's necessary.

gyuho: @eahydra Done, PTAL

eahydra (approved):
lgtm

Issue #410, opened 2025-02-14
log size is too big


eahydra:
I found gpud didn't rotate log file, caused the log file size is too big .

```shell
root@xxxx:~# ll -h /var/log/gpud.log
-rw-r--r-- 1 root root 9.5G Feb 14 11:45 /var/log/gpud.log
```

eahydra:
resolved.

PR #437, opened 2025-02-18
cmd(gpud): support "run --log-file" with built-in log rotation

gyuho:
Address https://github.com/leptonai/gpud/issues/410

### DIFF
--- a/cmd/gpud/command/logs.go
+++ b/cmd/gpud/command/logs.go
@@ -12,7 +12,7 @@ import (
 )
 
 func cmdLogs(cliContext *cli.Context) error {
-	if _, err := os.Stat("/var/log/gpud.log"); err != nil {
+	if _, err := os.Stat(logFile); err != nil {
 		return fmt.Errorf("log file %s does not exist", logFile)
 	}
 
--- a/cmd/gpud/command/logs.go
+++ b/cmd/gpud/command/logs.go
@@ -11,10 +11,8 @@ import (
 	"github.com/urfave/cli"
 )
 
-const logFile = "/var/log/gpud.log"
-
 func cmdLogs(cliContext *cli.Context) error {
-	if _, err := os.Stat(logFile); err != nil {
+	if _, err := os.Stat("/var/log/gpud.log"); err != nil {
 		return fmt.Errorf("log file %s does not exist", logFile)
 	}
 
--- a/cmd/gpud/command/run.go
+++ b/cmd/gpud/command/run.go
@@ -28,21 +28,11 @@ func cmdRun(cliContext *cli.Context) error {
 		os.Exit(1)
 	}
 
-	var zapLvl zap.AtomicLevel = zap.NewAtomicLevel() // info level by default
-	if logLevel != "" && logLevel != "info" {
-		var err error
-		zapLvl, err = zap.ParseAtomicLevel(logLevel)
-		if err != nil {
-			return err
-		}
+	zapLvl, err := log.ParseLogLevel(logLevel)
+	if err != nil {
+		return err
 	}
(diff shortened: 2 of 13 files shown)


cmd/gpud/command/logs.go
```diff
@@ -11,10 +11,8 @@ import (
 	"github.com/urfave/cli"
 )
 
-const logFile = "/var/log/gpud.log"
-
 func cmdLogs(cliContext *cli.Context) error {
-	if _, err := os.Stat(logFile); err != nil {
+	if _, err := os.Stat("/var/log/gpud.log"); err != nil {
```
eahydra, line 15: should use logFile here?

cmd/gpud/command/run.go
```diff
@@ -30,15 +30,20 @@ func cmdRun(cliContext *cli.Context) error {
...
 		var err error
 		zapLvl, err = zap.ParseAtomicLevel(logLevel)
 		if err != nil {
 			return err
 		}
+	}
+	if logFile != "" {
+		log.Logger = log.CreateLoggerWithLumberjack(logFile, 128, zapLvl.Level())
+	} else {
+		lCfg := log.DefaultLoggerConfig()
```
eahydra, line 42: The gpud systemd service is not changed in this PR, so I'm wondering how we plan to upgrade the systemd service to remove the stdout/err appen parameters?

https://github.com/leptonai/gpud/blob/2175d7d3671ccde724a6145ddbb87aa3f36e2cbd/pkg/gpud-manager/systemd/gpud.service#L14-L15

cmd/gpud/command/run.go
```diff
@@ -30,15 +30,20 @@ func cmdRun(cliContext *cli.Context) error {
...
 		lCfg.Level = zapLvl
 		log.Logger = log.CreateLogger(lCfg)
 	}
+
```
eahydra, line 46: We should merge them together and pass the logfile to the log package. If the logfile is empty, we should print the logs to stdout/stderr. Otherwise, we should output logs to both the file and stdout/stderr. Additionally, we should add a flag to disable output to stdout/stderr (by default, it should be enabled). In gpud.service, we should only print to the file.

pkg/log/log.go
```diff
@@ -24,6 +25,28 @@ func DefaultLoggerConfig() *zap.Config {
 	return &c
 }
 
+func CreateLoggerWithLumberjack(logFile string, maxSize int, logLevel zapcore.Level) *LeptonLogger {
+	w := zapcore.AddSync(&lumberjack.Logger{
+		Filename:   logFile,
+		MaxSize:    maxSize, // megabytes
+		MaxBackups: 3,
```
eahydra, line 32: We can keep more backups, the file size is not very large after compression.

cmd/gpud/command/run.go
gyuho: For now, this is optional (only enabled when `run --log-file` is not empty /cc @cardyok 

We can drop this once we migrate

cmd/gpud/command/up.go
```diff
@@ -48,11 +48,6 @@ func cmdUp(cliContext *cli.Context) (retErr error) {
 		return err
 	}
 
-	if err := systemd.LogrotateInit(); err != nil {
```
eahydra, line 51: If we remove it, we should update gpud.service together? otherwise the file cannot be rotated.

gyuho: discussed offline, this is safe, once we roll out the lumberjack based logrotates

eahydra (approved):
lgtm

eahydra (approved):

PR #439, opened 2025-02-19
feat(reboot): introduce initializing state for critical components on reboot

### LINKED ISSUE #431: After the machine restarts, gpud outputs persistence mode is enabled
GPUd version: v0.4.3

Case: reboot the system
Result: got the following output
```
persistence mode is not enabled on GPU 00000000:0F:00.0 (nvidia-smi); persistence mode is not enabled on GPU 00000000:34:00.0 (nvidia-smi); persistence mode is not enabled on GPU 00000000:48:00.0 (nvidia-smi);
```

Exptected: Wait for a while to confirm that persistence mode is initialized successfully and then check again

cardyok:


components/accelerator/nvidia/persistence-mode/component_output.go
```diff
@@ -125,5 +130,12 @@ func (o *Output) States() ([]components.State, error) {
 			StateKeyPersistenceModeEncoding: StateValueMemoryUsageEncodingJSON,
 		},
 	}
+	if !healthy {
+		lastReboot, err := reboot.LastReboot(context.Background())
+		if err == nil && time.Since(lastReboot) < initializeGracePeriod {
+			state.Health = components.StateInitializing
```
eahydra, line 136: Did that issue only occur during the reboot stage?

components/containerd/pod/component.go
```diff
@@ -57,6 +58,17 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 		return nil, err
 	}
 	if last.Error != nil {
+		lastReboot, err := reboot.LastReboot(context.Background())
+		if err == nil && time.Since(lastReboot) < initializeGracePeriod {
```
eahydra, line 62: Is it possible to check this condition in the state collection handler? And if this condition is true, tell the controle-plane: `I am initializing, please wait a moment`?

components/accelerator/nvidia/persistence-mode/component_output.go
```diff
@@ -58,6 +61,8 @@ const (
 	StateKeyPersistenceModeData       = "data"
 	StateKeyPersistenceModeEncoding   = "encoding"
 	StateValueMemoryUsageEncodingJSON = "json"
+
+	initializeGracePeriod = 3 * time.Minute
```
eahydra, line 65: Should we define constant values ​​somewhere to avoid duplication? Also, 3 minutes is a bit long. Do we have past data that 3 minutes is reasonable? 1 minute might be enough.

cardyok: our control plane collects every 1 min and I checked some historical data seems it will still fail on second collection, 3 min is about 3 collections, just to be secure..

components/containerd/pod/component.go
cardyok: yea thats doable..but will affect all components (all components will be default initializing when machine just boot up)..if this checking is local to each component is more flexible..

eahydra: In your implementation, only check if the duration from last reboot is less than initialize grace period when the component is unhealthy. We can do same thing in the state collection handler...

components/accelerator/nvidia/persistence-mode/component_output.go
eahydra: Got it

pkg/session/serve.go
```diff
@@ -364,6 +368,20 @@ func (s *Session) getStates(ctx context.Context, payload Request) (v1.LeptonStat
...
+			if !componentState.Healthy {
+				if lastRebootTime.IsZero() {
+					lastRebootTime, err = reboot.LastReboot(context.Background())
+					if err != nil {
+						log.Logger.Errorw("failed to get last reboot time", "error", err)
+					}
+				}
+				if time.Since(lastRebootTime) < initializeGracePeriod {
+					currState.States[i].Health = components.StateInitializing
+					currState.States[i].Healthy = true
```
eahydra, line 381: Should add warn log here.

eahydra (commented):
lgtm

eahydra (approved):
lgtm

PR #441, opened 2025-02-19
feat(session): do parallel component collection

cardyok:


pkg/session/serve.go
```diff
@@ -336,35 +367,49 @@ func (s *Session) getStates(ctx context.Context, payload Request) (v1.LeptonStat
 	if len(payload.Components) > 0 {
 		allComponents = payload.Components
 	}
+	var mu sync.Mutex
```
xiang90, line 370: var (
...
)

pkg/session/serve.go
```diff
@@ -246,38 +247,58 @@ func (s *Session) getEvents(ctx context.Context, payload Request) (v1.LeptonEven
...
+					"error", err,
+				)
+			} else {
+				log.Logger.Debugw("successfully got events", "component", componentName)
+				currEvent.Events = event
+			}
+			eventsBuf <- currEvent
+		}()
+	}
+	wg.Wait()
```
eahydra, line 299: Actually we don't need to wait for all goroutines here. We should ensure all gorooutines send result to the channel even if it failed.

eahydra: You have already done this..so we can remove wg here..

eahydra (approved):
lgtm

PR #443, opened 2025-02-19
feat(disk/lsblk): more debug info when "lsblk --version" parse fails

gyuho:


eahydra (approved):
lgtm

PR #448, opened 2025-02-21
feat(infiniband): use events store to track historical ibstat status

gyuho:

### DIFF
--- a/components/accelerator/nvidia/infiniband/component.go
+++ b/components/accelerator/nvidia/infiniband/component.go
@@ -4,15 +4,19 @@ package infiniband
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"sync"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/leptonai/gpud/components"
 	nvidia_infiniband_id "github.com/leptonai/gpud/components/accelerator/nvidia/infiniband/id"
 	"github.com/leptonai/gpud/pkg/common"
 	nvidia_common "github.com/leptonai/gpud/pkg/config/common"
+	events_db "github.com/leptonai/gpud/pkg/events-db"
 	"github.com/leptonai/gpud/pkg/log"
 	"github.com/leptonai/gpud/pkg/nvidia-query/infiniband"
 )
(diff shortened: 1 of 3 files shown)


components/accelerator/nvidia/infiniband/component.go
```diff
@@ -59,72 +180,51 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
...
-			Descriptions: []string{
-				"potential infiniband switch/hardware issue needs immediate attention",
-			},
-		}
-
-		log.Logger.Warnw("ibstat issue found", "reason", reason, "output", o.Raw)
+	state := components.State{
+		Name:             lastEvent.Name,
+		Healthy:          true,
+		Health:           "",
```
cardyok, line 214: should be Healthy

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,18 +43,135 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+func (c *component) checkIbstatOnce(ts time.Time, thresholds infiniband.ExpectedPortStates) error {
+	if thresholds.AtLeastPorts <= 0 && thresholds.AtLeastRate <= 0 {
+		return nil
+	}
+
+	cctx, ccancel := context.WithTimeout(c.rootCtx, 15*time.Second)
+	o, err := infiniband.GetIbstatOutput(cctx, []string{c.toolOverwrites.IbstatCommand})
+	ccancel()
+	if err != nil && !errors.Is(err, infiniband.ErrNoIbstatCommand) {
+		return err
```
eahydra, line 104: should we record the error as event?

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -59,72 +180,51 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 	return c.getStates(ctx, GetDefaultExpectedPortStates())
 }
 
+var noDataEvents = []components.State{
```
eahydra, line 183: the type is State, but the variable name is events. it's a bit weird..

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+	rootCtx        context.Context
+	cancel         context.CancelFunc
+	eventsStore    events_db.Store
 	toolOverwrites nvidia_common.ToolOverwrites
+
+	lastEventMu sync.RWMutex
+	lastEvent   *components.Event
+}
+
+func (c *component) getLastEvent() *components.Event {
```
eahydra, line 80: should we remove the function? we can get last event from checkIbStatOnce

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+		SuggestedActions: ev.SuggestedActions,
+	}
+	if len(ev.ExtraInfo) > 0 {
+		state.Healthy = ev.ExtraInfo["state_healthy"] == "true"
+		state.Health = ev.ExtraInfo["state_health"]
+	}
+	return state
+}
+
+func (c *component) checkIbstatOnce(ts time.Time, thresholds infiniband.ExpectedPortStates) error {
```
eahydra, line 101: the function should return last event. If ibstat is healthy, we can return nil.

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+	if thresholds.AtLeastPorts <= 0 && thresholds.AtLeastRate <= 0 {
+		return nil
+	}
+
+	ev := components.Event{
+		Time:    metav1.Time{Time: ts},
+		Name:    "ibstat",
+		Type:    common.EventTypeUnknown,
```
eahydra, line 109: I am not sure if we should use Unknown here.

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+			"state_healthy": "true",
+			"state_health":  components.StateHealthy,
+		},
+		SuggestedActions: nil,
+	}
+
+	c.lastEventMu.Lock()
+	defer c.lastEventMu.Unlock()
+
+	cctx, ccancel := context.WithTimeout(c.rootCtx, 15*time.Second)
```
eahydra, line 121: We should check if the duration from last call this function to now exceeds a threshold, and return the last event if it does not exceed the threshold.

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+
+			log.Logger.Warnw("ibstat command not found", "reason", ev.Message)
+		} else {
+			ev.Type = common.EventTypeWarning
+			ev.Message = fmt.Sprintf("ibstat command failed: %v", err)
+			ev.ExtraInfo["state_healthy"] = "false"
+			ev.ExtraInfo["state_health"] = components.StateUnhealthy
+
+			log.Logger.Warnw("ibstat command failed", "error", err)
+		}
```
eahydra, line 140:
```suggestion
        ev.Type = common.EventTypeWarning
        ev.ExtraInfo["state_healthy"] = "false"
		ev.ExtraInfo["state_health"] = components.StateUnhealthy
		if errors.Is(err, infiniband.ErrNoIbstatCommand) {
			ev.Message = fmt.Sprintf("ibstat threshold set but %v", err)
		} else {
			ev.Message = fmt.Sprintf("ibstat command failed: %v", err)
		}
	    log.Logger.Warnw("failed to run ibstat command", "reason", ev.Message)
```

components/accelerator/nvidia/infiniband/component.go
```diff
@@ -39,92 +43,218 @@ func SetDefaultExpectedPortStates(states infiniband.ExpectedPortStates) {
...
+	} else {
+		reason, healthy, err := evaluate(o, thresholds)
+		if err != nil {
+			return err
```
eahydra, line 144: should log the error

components/accelerator/nvidia/infiniband/component.go
gyuho: Oh it's done outside of this function

```go
	// last event already happened within the last 10 seconds, skip the check
	if lastEvent != nil && now.UTC().Sub(lastEvent.Time.Time) < 10*time.Second {
		return []components.State{convertToState(lastEvent)}, nil
	}

	if err := c.checkIbstatOnce(now.UTC(), thresholds); err != nil {
		return nil, err
	}
```

gyuho: let me put that logic inside

eahydra (approved):
overall lgtm. @cardyok PTAL

cardyok (approved):

PR #450, opened 2025-02-23
feat(nvidia-query): use "ERROR_ARGUMENT_VERSION_MISMATCH" to decide whether GPM is supported or not

gyuho:

### DIFF
--- a/pkg/nvidia-query/nvml/error.go
+++ b/pkg/nvidia-query/nvml/error.go
@@ -6,21 +6,27 @@ import (
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 )
 
+// IsVersionMismatchError returns true if the error indicates a version mismatch.
+func IsVersionMismatchError(ret nvml.Return) bool {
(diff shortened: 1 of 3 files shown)


pkg/nvidia-query/nvml/error.go
```diff
@@ -11,6 +11,16 @@ func IsNotSupportError(ret nvml.Return) bool {
 	if ret == nvml.ERROR_NOT_SUPPORTED {
 		return true
 	}
+
+	// "Argument version mismatch" indicates that the NVML library
+	// is not compatible with the corresponding API call
+	// thus marking the operation as not supported.
+	if ret == nvml.ERROR_ARGUMENT_VERSION_MISMATCH {
+		return true
+	}
```
eahydra, line 20: let's not mix them together. Create a function for VersionMisatch, and just call it in GPM-related code.

gyuho: addressed

eahydra (approved):
lgtm

PR #451, opened 2025-02-24
cleanup bash scripts with distinct prefix

gyuho:
- **feat(pkg/process): support custom bash script file name and directory**
- **add remove bash files util func**
- **rename "DefaultBashScriptFilePattern"**


eahydra (approved):
lgtm

PR #453, opened 2025-02-24
fix(fd): take file_max into limit setting

cardyok:


gyuho (approved):

components/fd/component_output.go
```diff
@@ -298,7 +299,7 @@ func CreateGet(cfg Config) query.GetFunc {
 		fileHandlesSupported := file.CheckFileHandlesSupported()
 		var thresholdAllocatedFileHandlesPct float64
 		if cfg.ThresholdAllocatedFileHandles > 0 {
-			thresholdAllocatedFileHandlesPct = calcUsagePct(usage, cfg.ThresholdAllocatedFileHandles)
+			thresholdAllocatedFileHandlesPct = calcUsagePct(usage, utils.Uint64Min(cfg.ThresholdAllocatedFileHandles, limit))
```
eahydra, line 302: golang has a built-in min function..

eahydra (approved):
lgtm

Issue #348, opened 2025-02-02
feature: detect pcie power supply issue from dmesg


xiang90:
Example log 

> Detected insufficient power on the PCIe slot

Once this is detected, we should emit a critical event (with the PCIe name /address ) and change the specific PICe device to a degraded state. 

The suggested action should be `Hardware inspection`. For now, we do not automatically resolve this event. Once the power supply is fixed (hardware repair), we manually change it back to the normal state.

gyuho:
https://github.com/leptonai/gpud/pull/293/files had detected

> [Sun Dec  1 14:54:40 2024] ny2g1r12hh2 kernel: mlx5_core 0000:5c:00.0: mlx5_pcie_event:299:(pid 268269): Detected insufficient power on the PCIe slot (27W).

(now removed)

Also, we can monitor power brake values for NVIDIA-specific data:

https://github.com/leptonai/gpud/blob/6bc689b15fa8257f61c6156043e0626b4a1ff67e/pkg/nvidia-query/nvml/clock_events.go#L149

/cc @eahydra

eahydra:
It seems that the nvidia clock event can help find the power brake. 
The `insufficient powner on the PCIe slot` is general for all PCI devices?  I think we should reopen #293 and improve it.

PR #467, opened 2025-02-26
feat(nvidia/infiniband): watch mellanox power events via dmesg/kernel

gyuho:
Fix https://github.com/leptonai/gpud/issues/348.

### DIFF
--- a/components/accelerator/nvidia/infiniband/component.go
+++ b/components/accelerator/nvidia/infiniband/component.go
@@ -16,6 +16,7 @@ import (
 	nvidia_infiniband_id "github.com/leptonai/gpud/components/accelerator/nvidia/infiniband/id"
 	"github.com/leptonai/gpud/pkg/common"
 	nvidia_common "github.com/leptonai/gpud/pkg/config/common"
+	"github.com/leptonai/gpud/pkg/dmesg"
 	events_db "github.com/leptonai/gpud/pkg/events-db"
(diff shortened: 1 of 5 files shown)


eahydra:
Would you please split the PR into two parts?  One is related to dmesg and the other is to remove the power component.

gyuho:
@eahydra Will rebase on
- https://github.com/leptonai/gpud/pull/468
- https://github.com/leptonai/gpud/pull/469

xiang90:
does this only apply for mlx devices? basically just ib device? can we also check power issue for gpu cards?

gyuho:
> does this only apply for mlx devices? basically just ib device

Yes, these are all from mellanox linux driver code: https://github.com/torvalds/linux/blob/ac9c34d1e45a4c25174ced4fc0cfc33ff3ed08c7/drivers/net/ethernet/mellanox/mlx5/core/events.c#L295-L299

eahydra (approved):

eahydra:
> 
> > power issue for gpu cards?
> 
> We already have this power component:
> 
> https://github.com/leptonai/gpud/blob/a6b6735fc073aeefd6ef7ff67668a10ec76786cd/components/accelerator/nvidia/power/component_output.go#L94-L115
> 
> @xiang90 @eahydra Let me know if we prefer have these in nvidia/power component

Did we find power issue for GPUs in dmesg? If not, we can continue with the code above.

PR #470, opened 2025-02-26
feat(pkg/login): unit testable code base, add unit tests

gyuho:

### DIFF
--- a/pkg/netutil/netutil.go
+++ b/pkg/netutil/netutil.go
@@ -1,0 +1,54 @@
+package netutil
+
+import (
+	"context"
+	"io"
+	"net"
+	"net/http"
+	"strings"
+	"time"
+)
+
+func PublicIP() (string, error) {
+	return publicIP("https://ifconfig.me")
+}
+
+func publicIP(target string) (string, error) {
+	// Create a transport that forces IPv4
+	transport := &http.Transport{
+		// Force IPv4 by using tcp4 network
+		DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
+			return (&net.Dialer{
+				Timeout:   5 * time.Second,
+				KeepAlive: 30 * time.Second,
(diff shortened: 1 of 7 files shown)


pkg/netutil/netutil.go
```diff
@@ -0,0 +1,16 @@
+package netutil
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func PublicIP() (string, error) {
+	cmd := exec.Command("curl", "-4", "ifconfig.me")
```
eahydra, line 9: There is no need to use curl if there is another way to do the same thing. Just use http.Client to get the public IP address.

gyuho: @eahydra done

pkg/netutil/netutil_test.go
```diff
@@ -0,0 +1,17 @@
+package netutil
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPublicIP(t *testing.T) {
```
eahydra, line 9: Currently we use http.Client to request the public ip,  we'd mock the `ifconfig.me` to verify the function.

pkg/netutil/netutil.go
```diff
@@ -0,0 +1,50 @@
...
+		},
+	}
+
+	client := &http.Client{
+		Timeout:   10 * time.Second,
+		Transport: transport,
+	}
+
+	// Create a request to mimic curl behavior
+	req, err := http.NewRequest("GET", "https://ifconfig.me", nil)
```
eahydra, line 30: make `ifconfig.me` as a param, if the param is empty, use `ifconfig.me` by default.

pkg/netutil/netutil_test.go
gyuho: done

eahydra (approved):
lgtm

PR #471, opened 2025-02-27
feat(pkg/latency): "Closest" helper for "join" region detection

gyuho:


pkg/latency/latency.go
```diff
@@ -26,6 +26,19 @@ func (l Latencies) RenderTable(wr io.Writer) {
 	table.Render()
 }
 
+func (l Latencies) Closest() Latency {
```
xiang90, line 29: add comment.

eahydra (approved):
lgtm

PR #476, opened 2025-02-28
feat(netutils/latency): move pkg, add ut from tailscale code base

gyuho:


eahydra (approved):
lgtm

PR #480, opened 2025-02-28
fix(session/serve): fix response error JSON encoding

gyuho:
otherwise, the server will receive

> "error": {}


cardyok (commented):
lgtm, need more testing with 0.4.5 rc release

eahydra (approved):
lgtm

PR #485, opened 2025-03-03
nit(session): do not shadow component name var in metrics get within for-loops

gyuho:
in case someone uses Go 1.22<


eahydra (approved):
lgtm

PR #487, opened 2025-03-04
feat(query/log): remove, and simplify "scan" dmesg func for one-off operations, remove "gpud logs"

gyuho:
TODO: replace this with kmsg reader


pkg/diagnose/scan.go
```diff
@@ -8,8 +8,8 @@ import (
 	"runtime"
 	"time"
 
-	nvidia_component_error_sxid "github.com/leptonai/gpud/components/accelerator/nvidia/error/sxid"
-	nvidia_component_error_xid "github.com/leptonai/gpud/components/accelerator/nvidia/error/xid"
+	error_sxid "github.com/leptonai/gpud/components/accelerator/nvidia/error/sxid"
+	error_xid "github.com/leptonai/gpud/components/accelerator/nvidia/error/xid"
```
eahydra, line 12: These two aliases are confusing and can easily mislead us into thinking that these two packages define error types.

eahydra (approved):

pkg/diagnose/scan.go
gyuho: renamed to `nvidia_x/sxid`

Issue #442, opened 2025-02-19
Programmatically monitor kernel messages (vs. watch dmesg output)


gyuho:
Some references:
 
- https://github.com/kubernetes/kubernetes/blob/1b71dc77f2f2ec190c044a9b78f4434b1c33da6b/pkg/kubelet/oom/oom_watcher_linux.go
- https://github.com/google/cadvisor/blob/v0.51.0/utils/oomparser/oomparser.go#L38
- https://github.com/euank/go-kmsg-parser

For comparison, kubelet does:

>   Warning  SystemOOM                36m                kubelet          System OOM encountered, victim process: vector, pid: 3052236

gyuho:
Discussion: experiment "euank/go-kmsg-parser" as direct parser with syscall

> simple cli-tool that functions similarly to dmesg --ctime --follow is included

(https://github.com/euank/go-kmsg-parser -- we will adapt the code logic into gpud)

(c.f., https://github.com/leptonai/gpud/issues/440#issuecomment-2681046151)

gyuho:
Done. Closing in favor of https://github.com/leptonai/gpud/issues/512.

PR #488, opened 2025-03-04
feat(kmsg): use "euank/go-kmsg-parser" for kernel message scan (experimental)

gyuho:
Address https://github.com/leptonai/gpud/issues/442.

c.f., https://github.com/euank/go-kmsg-parser


pkg/diagnose/scan_linux.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	defer kl.Close()
+
+	ch := make(chan kmsgparser.Message, 1024)
+	if err := kl.Parse(ch); err != nil {
+		log.Logger.Warnw("error parsing kmsg", "error", err)
+		return
+	}
+
+	donec := make(chan struct{})
```
eahydra, line 34: better to use buffer channel

pkg/diagnose/scan_linux.go
```diff
@@ -0,0 +1,60 @@
...
+	fmt.Printf("%s scanning kmsg\n", inProgress)
+	kl, err := kmsgparser.NewParser(kmsgparser.WithNoFollow())
+	if err != nil {
+		log.Logger.Warnw("error creating kmsg parser", "error", err)
+		return
+	}
+	defer kl.Close()
+
+	ch := make(chan kmsgparser.Message, 1024)
+	if err := kl.Parse(ch); err != nil {
```
eahydra, line 29: Parse is blocked, so we should use a separate goroutine for it.  
We can try use [errgroup](golang.org/x/sync/errgroup) to make the code clean..

```go
ch := make(chan kmsgparser.Message, 1024)
group, _ := errgroup.WithContext(c)
group.Go(func() error{
   return kl.Parse(ch)
})

cnt := 0
for msg := range ch {
    cnt++
	ts := humanize.RelTime(msg.Timestamp, now, "ago", "from now")

	if found := nvidia_xid.Match(msg.Message); found != nil {
		fmt.Printf("[XID found] (%s) %q\n", ts, msg.Message)
	}
	if found := nvidia_sxid.Match(msg.Message); found != nil {
		fmt.Printf("[SXID found] (%s) %q\n", ts, msg.Message)
	}
}

err := group.Wait()
if err != nil {
   log.Logger.Errorw("failed to parse kmsg", "error", err) 
}

```

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/491.

PR #489, opened 2025-03-05
feat(nvidia-query): remove some redundant nvidia-smi dependencies

gyuho:
in favor of direct nvml calls


eahydra (approved):
lgtm, should fix the UT

PR #493, opened 2025-03-06
feat(containerd): simplify /states with goroutines

gyuho:


pkg/server/server.go
```diff
@@ -1126,21 +1115,8 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID
...
-								continue
-							}
-							ccfg = *parsed
-						}
-						if err := ccfg.Validate(); err != nil {
-							log.Logger.Errorw("failed to validate component %s config: %w", name, err)
-							continue
-						}
-						componentsToAdd = append(componentsToAdd, containerd_pod.New(ctx, ccfg))
+					if exists := containerd_pod.CheckContainerdRunning(ctx); exists {
```
eahydra, line 1118: This code runs in a loop. We should check if the containerd component exists in componentsToAdd.. And the weird thing is that the loop is an array of component names. There should be a common factory, or simply iterate over an array of factories.  The pseudo code should be similar to the following
```go
for _, factoryFn := range []func(ctx) components.Component {
           containerd_pod.New, 
           docker_container.New, 
           k8s_pod.New,
} {
    // move these check such containerd_pod.CheckContainerdRunning into their factory fn
     component := factoryFn(ctx)
     componentsToAdd = append(componentsToAdd, component)
}
```

gyuho: @eahydra Yes, let us address in the following PR https://github.com/leptonai/gpud/issues/494, since we need changes in those 3 components

components/containerd/pod/component.go
```diff
@@ -3,113 +3,313 @@ package pod
...
-				Name:    containerd_pod_id.Name,
-				Healthy: false,
-				Error:   last.Error.Error(),
-				Reason:  reason,
-			},
-		}, nil
+		c.lastMu.Lock()
+		c.lastErr = err.Error()
+		c.lastReason = reason
+		c.lastMu.Unlock()
```
eahydra, line 129: we should reset lastPods if failed?

components/containerd/pod/component.go
```diff
@@ -3,113 +3,313 @@ package pod
...
+	c := &component{
+		rootCtx:  ctx,
+		cancel:   cancel,
+		endpoint: defaultContainerRuntimeEndpoint,
 	}
+	go func() {
+		ticker := time.NewTicker(time.Minute)
+		defer ticker.Stop()
+		for {
+			select {
```
eahydra, line 32: The first time we have to wait for 1 minute. perhaps we should run checkOnce, then wait for ticker?

eahydra (approved):
lgtm

PR #491, opened 2025-03-05
feat(kmsg): initial commit (experiment to replace dmesg watcher)

gyuho:
```
⌛ scanning kmsg
✔ scanned kmsg file -- found 4040 line(s)
✔ first kmsg line is 1 week ago old
```

to replace https://github.com/leptonai/gpud/pull/488 and address https://github.com/leptonai/gpud/issues/442.

Also close https://github.com/leptonai/gpud/issues/478.

c.f., https://github.com/euank/go-kmsg-parser

PR #496, opened 2025-03-06
feat(components): watch with kmsg (experimental)

gyuho:
c.f., https://github.com/leptonai/gpud/pull/491

### DIFF
--- a/pkg/kmsg/events.go
+++ b/pkg/kmsg/events.go
@@ -1,0 +1,45 @@
+package kmsg
+
+import (
+	"os"
+	"runtime"
+
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+// CreateEventsWatcher creates a new events watcher that will watch the kmsg.
+// Experimental.
+// TODO: integrate with the events store.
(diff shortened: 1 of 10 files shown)


pkg/kmsg/events.go
```diff
@@ -0,0 +1,45 @@
...
+	"os"
+	"runtime"
+
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+// CreateEventsWatcher creates a new events watcher that will watch the kmsg.
+// Experimental.
+// TODO: integrate with the events store.
+func CreateEventsWatcher(matchFunc func(line string) (eventName string, message string)) (Watcher, error) {
```
eahydra, line 13: is it better that rename to StartWatch? because it's in kmsg package.

pkg/kmsg/events.go
```diff
@@ -0,0 +1,45 @@
...
+	if os.Getenv("GPUD_EXPERIMENTAL_KMSG_WATCHER") != "true" {
```
eahydra, line 14: probably we can remove the feature-gate,  and don't fatal the process if  failed to start watch.

eahydra (approved):
lgtm

PR #501, opened 2025-03-07
feat(nvidia): only keep "nvidia-smi --query"

gyuho:
To deprecate nvidia-smi --query as well soon


eahydra (approved):
lgtm, should fix ut.

Issue #494, opened 2025-03-06
Cleanup, simplify dynamic component detection


gyuho:
https://github.com/leptonai/gpud/pull/493#discussion_r1982662421

> This code runs in a loop. We should check if the containerd component exists in componentsToAdd.. And the weird thing is that the loop is an array of component names. There should be a common factory, or simply iterate over an array of factories.  The pseudo code should be similar to the following

```go
for _, factoryFn := range []func(ctx) components.Component {
           containerd_pod.New, 
           docker_container.New, 
           k8s_pod.New,
} {
    // move these check such containerd_pod.CheckContainerdRunning into their factory fn
     component := factoryFn(ctx)
     componentsToAdd = append(componentsToAdd, component)
}
```

/cc @eahydra

PR #505, opened 2025-03-07
feat(components): simplify kubelet/containerd/docker states

gyuho:
Fix https://github.com/leptonai/gpud/issues/494.


components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+	defaultSocketFile               = "/run/containerd/containerd.sock"
+	defaultContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
+)
+
+// NOTE
+// DO NOT USE https://github.com/kubernetes/kubernetes/blob/v1.32.0-alpha.0/staging/src/k8s.io/cri-client/pkg/remote_runtime.go yet
+// it fails with
+// "code = Unavailable desc = name resolver error: produced zero addresses"
+
+const (
```
eahydra, line 192: move all these cri related code to cri.go

components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
-func (o *Output) States() ([]components.State, error) {
-	if len(o.Pods) == 0 {
-		return []components.State{{
-			Name:    containerd_pod_id.Name,
-			Health:  components.StateHealthy,
-			Healthy: true,
-			Reason:  "no output",
-		}}, nil
+// createConn creates a new CRI service client.
+// Cannot use "k8s.io/kubernetes/pkg/kubelet/cri/remote.NewRemoteRuntimeService" directly
```
eahydra, line 234: Since we don't really need to depend on the kubelet/cri package, we can remove the comment.

components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+	return runtimeClient, imageClient, conn, nil
 }
 
-const (
-	defaultSocketFile               = "/run/containerd/containerd.sock"
-	defaultContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
-)
+func CheckContainerdRunning(ctx context.Context) bool {
+	if runtime.GOOS != "linux" {
+		log.Logger.Debugw("ignoring default containerd pod checking since it's not linux", "os", runtime.GOOS)
```
eahydra, line 276: actually if we don't support non-linux os, we should log with warning level

components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+	containerdSocketExists := false
+	containerdRunning := false
 
-func listSandboxStatus(ctx context.Context, endpoint string) ([]*runtimeapi.PodSandboxStatusResponse, error) {
+	if _, err := os.Stat(defaultSocketFile); err == nil {
+		log.Logger.Debugw("containerd default socket file exists, containerd installed", "file", defaultSocketFile)
+		containerdSocketExists = true
+	} else {
+		log.Logger.Debugw("containerd default socket file does not exist, skip containerd check", "file", defaultSocketFile, "error", err)
+	}
```
eahydra, line 295: IMO we should check the unix socket first, then check binary.. because we rely on the socket.
And I'm not sure that if we stopped containerd service, does the socket file exist?

components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
-			Healthy: true,
-			Reason:  "no output",
-		}}, nil
+// createConn creates a new CRI service client.
+// Cannot use "k8s.io/kubernetes/pkg/kubelet/cri/remote.NewRemoteRuntimeService" directly
+// as it causes a bunch of go module errors, importing the whole kubernetes repo.
+// ref. https://github.com/kubernetes-sigs/cri-tools/blob/master/cmd/main.go
+// ref. https://github.com/kubernetes/kubernetes/blob/v1.29.2/pkg/kubelet/cri/remote/remote_runtime.go
+// ref. https://github.com/kubernetes/kubernetes/blob/v1.32.0-alpha.0/staging/src/k8s.io/cri-client/pkg/remote_runtime.go
+func createConn(ctx context.Context, endpoint string) (runtimeapi.RuntimeServiceClient, runtimeapi.ImageServiceClient, *grpc.ClientConn, error) {
```
eahydra, line 239: should return conn and error... init runtime service client and image service client by need on caller side.

components/docker/container/component.go
```diff
@@ -3,83 +3,137 @@ package container
...
+	lastMu   sync.RWMutex
+	lastData Data
 }
 
 func (c *component) Name() string { return docker_container_id.Name }
 
-func (c *component) Start() error { return nil }
+func (c *component) Start() error {
+	go func() {
+		ticker := time.NewTicker(1)
```
eahydra, line 52: typo

components/docker/container/component.go
```diff
@@ -3,83 +3,137 @@ package container
...
+	err error `json:"-"`
+	// set to true if the error is the connection error to the docker daemon
+	connErr bool `json:"-"`
+}
+
+func (d Data) marshalJSON() ([]byte, error) {
+	return json.Marshal(d)
+}
+
+func (d Data) DescribeReason() string {
```
eahydra, line 92: All of these should be updated referring to the comments in containerd/pod/component.go

components/containerd/pod/component.go
```diff
@@ -100,95 +149,170 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+
+	imageClient := runtimeapi.NewImageServiceClient(conn)
+	return runtimeClient, imageClient, conn, nil
 }
 
-const (
-	defaultSocketFile               = "/run/containerd/containerd.sock"
-	defaultContainerRuntimeEndpoint = "unix:///run/containerd/containerd.sock"
-)
+func CheckContainerdRunning(ctx context.Context) bool {
```
eahydra, line 274: Actually we don't need this function because the component should be enabled by default. and we should add a flag to control which components can be disabled or enabled(for those components disabled by default).

components/docker/container/component.go
```diff
@@ -3,83 +3,137 @@ package container
...
-
-	cctx, ccancel := context.WithCancel(ctx)
-	getDefaultPoller().Start(cctx, cfg.Query, docker_container_id.Name)
-
-	return &component{
-		rootCtx: ctx,
-		cancel:  ccancel,
-		poller:  getDefaultPoller(),
-		cfg:     cfg,
+func New(ctx context.Context, ignoreConnectionErrors bool) components.Component {
```
eahydra, line 24: IMO we should not support docker... or we should disable it by default.

components/kubelet/pod/component.go
```diff
@@ -3,87 +3,142 @@ package pod
...
+	lastMu   sync.RWMutex
+	lastData Data
 }
 
 func (c *component) Name() string { return kubelet_pod_id.Name }
 
-func (c *component) Start() error { return nil }
+func (c *component) Start() error {
+	go func() {
+		ticker := time.NewTicker(1)
```
eahydra, line 58: 1 again

components/kubelet/pod/component.go
```diff
@@ -3,87 +3,142 @@ package pod
...
+	c.lastMu.RLock()
+	lastData := c.lastData
+	c.lastMu.RUnlock()
+	return lastData.getStates(c.ignoreConnectionErrors)
+}
+
+type Data struct {
+	KubeletPidFound bool        `json:"kubelet_pid_found"`
+	NodeName        string      `json:"node_name,omitempty"`
+	Pods            []PodStatus `json:"pods,omitempty"`
```
eahydra, line 85: should updated by comments in contaienrd/pod

components/kubelet/pod/component.go
```diff
@@ -99,8 +154,270 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+	if string(b) != "ok" {
+		return fmt.Errorf("kubelet read-only port /healthz expected 'ok', got %q", string(b))
+	}
+
+	return nil
+}
+
+// CheckKubeletReadOnlyPortListening checks if the kubelet read-only port is listening.
+// It first checks if the kubelet is running and then checks if the port is open.
+func CheckKubeletReadOnlyPortListening(ctx context.Context, port int) bool {
```
eahydra, line 239: we should remove it , enable the component by default

pkg/server/server.go
```diff
@@ -131,12 +130,33 @@ type Server struct {
 	autoUpdateExitCode    int
 }
 
-func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID string, packageManager *gpud_manager.Manager, opts ...gpud_config.OpOption) (_ *Server, retErr error) {
-	options := &gpud_config.Op{}
-	if err := options.ApplyOpts(opts); err != nil {
-		return nil, err
-	}
+type createComponentFunc func(context.Context, *lepconfig.Config) components.Component
```
eahydra, line 133: We should export the function type, and should return error
```suggestion
type NewComponentFunc func(context.Context, *lepconfig.Config) (components.Component, error)
```

pkg/server/server.go
```diff
@@ -131,12 +130,33 @@ type Server struct {
...
+
+var createComponentFuncs = map[string]createComponentFunc{
+	containerd_pod_id.Name: func(ctx context.Context, _ *lepconfig.Config) components.Component {
+		exists := containerd_pod.CheckContainerdRunning(ctx)
+		if !exists {
+			return nil
+		}
```
eahydra, line 140: don't check here, move check into New function if need, and if doesn't support it, should mock a dummy or should return a special error like `Unsupported`..

We should hide some information that components need, otherwise this information will be leaked everywhere.

pkg/server/server.go
```diff
@@ -131,12 +130,33 @@ type Server struct {
...
 }
 
-func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID string, packageManager *gpud_manager.Manager, opts ...gpud_config.OpOption) (_ *Server, retErr error) {
-	options := &gpud_config.Op{}
-	if err := options.ApplyOpts(opts); err != nil {
-		return nil, err
-	}
+type createComponentFunc func(context.Context, *lepconfig.Config) components.Component
+
+var createComponentFuncs = map[string]createComponentFunc{
```
eahydra, line 135:
```suggestion
var newComponentFuncMap = map[string]createComponentFunc{
```

eahydra:
> Fix #494.

BTW, the PR only can solve parts of initialization code, should do more work.

components/containerd/pod/component.go
```diff
@@ -5,86 +5,135 @@ import (
...
+func (d Data) DescribeReason() string {
+	reason := fmt.Sprintf("total %d pod sandboxe(s)", len(d.Pods))
 
-func (c *component) Start() error { return nil }
+	if d.err != nil {
+		// this is the error from "ListSandboxStatus"
+		//
+		// e.g.,
+		// rpc error: code = Unimplemented desc = unknown service runtime.v1.RuntimeService
+		reason = "failed gRPC call to the containerd socket"
```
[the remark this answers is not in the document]
gyuho: Yes, to overwrite with the error message

components/containerd/pod/component.go
```diff
@@ -5,86 +5,135 @@ import (
...
-				Error:   lastErr,
-				Reason:  lastReason,
-			},
-		}, nil
+func (d Data) GetHealth() (string, bool) {
+	if d.err != nil {
+		return components.StateUnhealthy, false
 	}
+	return components.StateHealthy, true
+}
```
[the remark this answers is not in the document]
gyuho: now private method inside each component pkg

components/accelerator/nvidia/bad-envs/component.go
```diff
@@ -31,7 +31,7 @@ func New(ctx context.Context, cfg nvidia_common.Config) (components.Component, e
 	}, nil
 }
 
-var _ components.Component = (*component)(nil)
+var _ components.Component = &component{}
```
eahydra, line 34: Lolll, Thanks. And should we move this and the component struct before the New function?

components/containerd/pod/component.go
```diff
@@ -5,86 +5,135 @@ import (
...
+	if len(d.Pods) == 0 { // no pod found yet
+		return []components.State{state}, nil
 	}
-	return o.States()
+
+	b, _ := d.marshalJSON()
+	state.ExtraInfo = map[string]string{
+		"data":     string(b),
+		"encoding": "json",
+	}
```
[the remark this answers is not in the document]
gyuho: Let's address in the following PR https://github.com/leptonai/gpud/issues/508

eahydra (approved):

PR #513, opened 2025-03-10
feat(docker, containerd, kubelet, tailscale): clean up systemd service checks

gyuho:
c.f., https://github.com/leptonai/gpud/pull/505#discussion_r1987155164


components/tailscale/component.go
```diff
@@ -1,40 +1,79 @@
...
+			defer conn.Close()
+			return conn.IsActive(ctx, "tailscaled")
+		},
+	}
+	return c
+}
+
+func (c *component) Name() string { return Name }
+
+func (c *component) Start() error {
```
eahydra, line 54: I didn't find where it was referenced..

components/tailscale/component.go
```diff
@@ -50,5 +89,88 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
...
+		Name:   Name,
+		Reason: d.getReason(),
+	}
+	state.Health, state.Healthy = d.getHealth()
+
+	b, _ := json.Marshal(d)
+	state.ExtraInfo = map[string]string{
+		"data":     string(b),
+		"encoding": "json",
+	}
```
eahydra, line 174: Does Tailscaled component need to append the data into the extraInfo?

components/kubelet/pod/component.go
```diff
@@ -41,7 +44,15 @@ func New(ctx context.Context, kubeletReadOnlyPort int, ignoreConnectionErrors bo
...
 		kubeletReadOnlyPort:      kubeletReadOnlyPort,
-		ignoreConnectionErrors:   ignoreConnectionErrors,
+		checkServiceActive: func(ctx context.Context) (bool, error) {
+			conn, err := systemd.NewDbusConn(ctx)
+			if err != nil {
+				return false, err
+			}
+			defer conn.Close()
+			return conn.IsActive(ctx, "kubelet")
+		},
```
eahydra, line 54: We should make it as a function in systemd package
```go
func CheckServiceActive(ctx context.Context, name string) (bool, error)
```

components/containerd/pod/component.go
```diff
@@ -27,7 +28,11 @@ type component struct {
 
 	// returns true if the dependency is installed, thus requires the component checks
 	checkDependencyInstalled func() bool
-	endpoint                 string
+
+	// return true if the systemd service is active, otherwise false
```
eahydra, line 32: actually we can remove the comment.

components/tailscale/component.go
gyuho: https://github.com/leptonai/gpud/blob/760ef3b77506ae493e47765aaf3da0ac27c0e4a1/internal/server/server.go#L1247-L1252

eahydra: loll... Overlooked

eahydra (approved):
lgtm

PR #514, opened 2025-03-10
feat(nvidia): remove unused "error" component, smi usage for temp/power checks, remove smi parse

gyuho:


components/accelerator/nvidia/info/component_output.go
```diff
@@ -40,15 +37,6 @@ func ToOutput(i *nvidia_query.Output) *Output {
 	if i.SMI != nil {
 		o.Driver.Version = i.SMI.DriverVersion
```
eahydra, line 38: go-nvml supports querying driver and cuda version.

pkg/nvidia-query/detect.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/leptonai/gpud/pkg/process"
 )
 
+// Returns true if the local machine runs on Nvidia GPU
+// by running "nvidia-smi".
+func SMIExists() bool {
```
eahydra, line 19: Do we still need this function?

gyuho: Yes, used for some nvidia detection.

eahydra (approved):
lgtm

PR #515, opened 2025-03-11
feat(nvidia): remove nvidia-smi dependency in all places (use NVML instead)

gyuho:
```
⌛ scanning nvidia accelerators
✔ found library libnvidia-ml.so at /usr/lib/x86_64-linux-gnu/libnvidia-ml.so.535.216.03
✔ found library libcuda.so at /usr/lib/x86_64-linux-gnu/libcuda.so.535.216.03
```


eahydra (approved):
lgtm

PR #517, opened 2025-03-11
feat(cpu, memory, fd): simplify data poll

gyuho:


components/cpu/component.go
```diff
@@ -167,3 +155,352 @@ func (c *component) RegisterCollectors(reg *prometheus.Registry, dbRW *sql.DB, d
...
+	if err != nil {
+		d.Info.err = err
+		return
+	}
+	if len(infos) > 0 {
+		d.Info.Arch = arch
+		d.Info.CPU = infos[0].ModelName
+		d.Info.Family = infos[0].Family
+		d.Info.Model = infos[0].Model
+		d.Info.ModelName = infos[0].ModelName
```
eahydra, line 209: This data is immutable. So we could create a package called `os` or `system` or something and generate this data during the init phase.

components/cpu/component.go
```diff
@@ -167,3 +155,352 @@ func (c *component) RegisterCollectors(reg *prometheus.Registry, dbRW *sql.DB, d
...
+		return
+	}
+	cctx, ccancel = context.WithTimeout(c.ctx, 10*time.Second)
+	logicalCores, err := cpu.CountsWithContext(cctx, true)
+	ccancel()
+	if err != nil {
+		d.Cores.err = err
+		return
+	}
+	d.Cores.Physical = physicalCores
```
eahydra, line 227: Perhaps we can use https://pkg.go.dev/github.com/prometheus/procfs instead of these... it's more simple and used by node-exporter..

components/cpu/component.go
```diff
@@ -167,3 +155,352 @@ func (c *component) RegisterCollectors(reg *prometheus.Registry, dbRW *sql.DB, d
...
+		return
+	}
+
+	now := time.Now().UTC()
+	nowUnix := float64(now.Unix())
+	metrics.SetLastUpdateUnixSeconds(nowUnix)
+
+	prev := getPrevTimeStat()
+	cur := timeStats[0]
+	if prev == nil {
```
eahydra, line 249: we should wrap these into a function

components/memory/component.go
```diff
@@ -171,3 +150,150 @@ func (c *component) RegisterCollectors(reg *prometheus.Registry, dbRW *sql.DB, d
...
+	cctx, ccancel := context.WithTimeout(c.ctx, 5*time.Second)
+	vm, err := mem.VirtualMemoryWithContext(cctx)
+	ccancel()
+	if err != nil {
+		d.err = err
+		return
+	}
+
+	d.TotalBytes = vm.Total
+	d.TotalHumanized = humanize.Bytes(vm.Total)
```
eahydra, line 177: Why generate human-friendly information here? Who uses the data and who is responsible for formatting it

PR #519, opened 2025-03-11
feat(nvidia/remappined-rows): only suggest hw inspection if not reboot

gyuho:


eahydra (approved):
lgtm

PR #524, opened 2025-03-13
feat(nvidia): simulate "get remapped rows" (optional)

gyuho:
> {"level":"info","ts":"2025-03-13T06:37:35Z","caller":"nvml/nvml.go:39","msg":"injecting remapped rows pending","corrRows":0,"uncRows":10,"isPending":true,"failureOccurred":false}
> ...
> {"level":"warn","ts":"2025-03-13T06:38:09Z","caller":"remapped-rows/component_output.go:141","msg":"suggested actions","suggestedActions":"REBOOT_SYSTEM"}

### DIFF
--- a/pkg/nvidia-query/nvml/nvml.go
+++ b/pkg/nvidia-query/nvml/nvml.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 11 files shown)


pkg/nvidia-query/nvml/nvml.go
```diff
@@ -583,6 +450,190 @@ func (inst *instance) Get() (*Output, error) {
 	return st, joinedErr
 }
 
+func initAndCheckNVMLSupported(nvmlLib nvml.Interface) (bool, error) {
+	log.Logger.Infow("initializing nvml library")
+	ret := nvmlLib.Init()
+	if ret == nvml.SUCCESS {
+		return true, nil
```
eahydra, line 457: don't need to call `nvmlLib.Shutdown()`?

gyuho: oh only used init, we call shutdown outside of this function

eahydra (approved):
lgtm

PR #527, opened 2025-03-13
nit(pkg/controllers): remove redundant log header

gyuho:


eahydra:
loll.. actually `[package controller]` is a keyword helps me to easily filter out these logs...

gyuho:
oh got it, let's revisit :)

PR #528, opened 2025-03-13
feat(nvidia): clean up suggested actions (simplify)

gyuho:


pkg/common/action.go
```diff
@@ -43,3 +43,33 @@ func (sa *SuggestedActions) DescribeActions() string {
...
+		Descriptions: nil,
+	}
+
+	actionsToKeep := make(map[RepairActionType]bool)
+	for _, act := range sa.RepairActions {
+		actionsToKeep[act] = true
+	}
+	if _, ok := actionsToKeep[RepairActionTypeHardwareInspection]; ok {
+		delete(actionsToKeep, RepairActionTypeRebootSystem)
+	}
```
eahydra, line 63: We cannot do this since such as xid 79 defined two actions: reboot and hardware-inspection.. XID component sets the action by the history events.. for example, xid component suggests rebooting system for the first and second 79 error, suggests hardware-inspection to the third times...

gyuho: right, let me create a new PR

PR #530, opened 2025-03-13
fix(xid + sxid): truncate repair actions to pick hw inspection after reboot

gyuho:


components/accelerator/nvidia/xid/helper.go
```diff
@@ -62,6 +62,7 @@ func EvolveHealthyState(events []components.Event) (ret components.State) {
 						currXidErr.SuggestedActionsByGPUd.RepairActions[0] = common.RepairActionTypeHardwareInspection
 					}
 				}
+				currXidErr.SuggestedActionsByGPUd.RepairActions = currXidErr.SuggestedActionsByGPUd.RepairActions[:1]
```
eahydra, line 65: SXID has the same issue.

eahydra (approved):

PR #531, opened 2025-03-14
feat(os, peermem): ignore lsmod, system virt cmd timeouts (as non-actionable)

gyuho:

### DIFF
--- a/components/os/component_output.go
+++ b/components/os/component_output.go
@@ -2,6 +2,7 @@ package os
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"runtime"
 	"sync"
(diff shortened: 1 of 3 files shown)


components/os/component_output.go
```diff
@@ -278,15 +279,19 @@ func getDefaultPoller() query.Poller {
...
 		cctx, ccancel := context.WithTimeout(ctx, 15*time.Second)
 		virtEnv, err := getSystemdDetectVirtFunc(cctx)
 		ccancel()
 		if err != nil {
-			return nil, fmt.Errorf("failed to get virtualization environment using 'systemd-detect-virt': %w", err)
+			// ignore "context.DeadlineExceeded" since it's not a critical error and it's non-actionable
+			if !errors.Is(err, context.DeadlineExceeded) {
+				return nil, fmt.Errorf("failed to get virtualization environment using 'systemd-detect-virt': %w", err)
+			}
+			log.Logger.Warnw("failed to get virtualization environment using 'systemd-detect-virt'", "error", err)
```
eahydra, line 294: should we return here if failed to getSystemdDetectVirtFunc?  BTW, We just need to check if virtual env in initialization stage...

eahydra: we can improve it in next version

PR #532, opened 2025-03-14
feat(os): correctly detect first boot event based on machine-id file

gyuho:


gyuho:
Two are different data sources... not reliable...

PR #540, opened 2025-03-17
feat(remapped-rows): handle "system not ready" nvml error, refactor remapped-rows component

gyuho:
Handles the following system not ready error:

> {"level":"warn","ts":"2025-03-14T23:13:22Z","caller":"nvidia-query/query.go:209","msg":"nvml get failed","error":"failed to get clock events: failed to get device clock event reasons: System is not in ready state (GPU uuid GPU-509665ad-b600-ac93-3616-d754b23d636d)\nfailed to get device clock info for nvml.CLOCK_GRAPHICS: System is not in ready state (GPU uuid GPU-509665ad-b600-ac93-3616-d754b23d636d)"}
> {"level":"warn","ts":"2025-03-14T23:13:54Z","caller":"remapped-rows/component.go:73","msg":"last query failed -- returning cached, possibly stale data","error":"nvidia-smi check failed with 1 error(s); [error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field NvidiaSMIGPU.GPU4.Clocks Event Reasons of type query.SMIClockEventReasons]"}


pkg/nvidia-query/nvml/instance.go
```diff
@@ -0,0 +1,219 @@
...
+
+func (inst *instanceV2) Shutdown() error {
+	ret := inst.nvmlLib.Shutdown()
+	if ret != nvml.SUCCESS {
+		return fmt.Errorf("failed to shutdown nvml library: %s", ret)
+	}
+	return nil
+}
+
+var (
```
eahydra, line 145: please move these variable values and related code to a separate file.

eahydra (approved):

PR #555, opened 2025-03-22
fix(server): do not prematurely shutdown nvml instance for row remapping component

gyuho:

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -176,11 +176,6 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID 
 		if err != nil {
 			return nil, fmt.Errorf("failed to create NVML instance: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -958,7 +953,15 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID
 
 	go s.updateToken(ctx, dbRW, uid, endpoint)
 
-	go func() {
+	go func(nvmlInstance nvidia_query_nvml.InstanceV2) {
+		defer func() {
```
eahydra, line 957: We can temporarily fix this. Eventually, we need to ensure that there is only one nvml instance globally, and only one place is responsible for initializing and destroying the nvml instance. This is also what we need to solve in the next version.

eahydra (approved):

PR #561, opened 2025-03-24
fix: panic on special environments

eahydra:
During testing, it was found that gpud would panic randomly at any place, and the stack would point to the same place:

```go
runtime.throw({0x24db382?, 0xc0027c0808?})
	/usr/local/go/src/runtime/panic.go:1101 +0x48 fp=0xc002b122a0 sp=0xc002b12270 pc=0x497dc8
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:939 +0x1e5 fp=0xc002b122d0 sp=0xc002b122a0 pc=0x49a365
runtime.buildTypeAssertCache.func1(0x88?, 0x0?)
	/usr/local/go/src/runtime/iface.go:532 +0xc fp=0xc002b122f0 sp=0xc002b122d0 pc=0x42f14c
runtime.buildTypeAssertCache(0x24db33c?, 0x219d380, 0x0)
	/usr/local/go/src/runtime/iface.go:544 +0x1aa fp=0xc002b12370 sp=0xc002b122f0 pc=0x42f0ca
runtime.typeAssert(0x321c630, 0x219d380)
	/usr/local/go/src/runtime/iface.go:497 +0xae fp=0xc002b123a8 sp=0xc002b12370 pc=0x42ee6e
```

We all believe the interface type assertion cannot be failed. I found out that as long as function nvml.DeviceGetFieldValues is not called, there will be no panic.

I didn't delve into the root cause, but here's a guess: the machine with the problem was installed with NVIDIA 570 version driver, and when we called the nvml function, somehow the call stack or some memory was modified incorrectly. This is similar to heap corruption affecting the golang interface type switch cache.

We don't actually need nvlink tx/rx related data now, but there are other ways to get it later even if we need it.

### DIFF
--- a/components/accelerator/nvidia/nvlink/component.go
+++ b/components/accelerator/nvidia/nvlink/component.go
@@ -112,16 +112,8 @@ func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.
 	if err != nil {
 		return nil, fmt.Errorf("failed to read crc errors: %w", err)
 	}
-	rxBytes, err := nvidia_query_metrics_nvlink.ReadRxBytes(ctx, since)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read rx bytes: %w", err)
-	}
-	txBytes, err := nvidia_query_metrics_nvlink.ReadTxBytes(ctx, since)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read tx bytes: %w", err)
-	}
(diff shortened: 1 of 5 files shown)


gyuho (approved):

eahydra:
It has been running continuously for 9 hours on 4 machines and no panic occurred again.

Issue #510, opened 2025-03-10
Remove unnecessary dependencies


eahydra:
1. Remove fabric-manager component's dependency on systemd component.
2. Remove the xid and sxid components' dependency on os component.

For the 2nd, we should store the reboot event into a separate table. It can be implemented in a separate package, depends on the database, and other components just need to query the reboot event via interface(not database).

eahydra:
solved.

PR #570, opened 2025-03-25
feat(pci): refactor to clean up poller

gyuho:
Remove old poller.


components/accelerator/nvidia/sxid/component.go
```diff
@@ -217,7 +219,7 @@ func (c *SXIDComponent) SetHealthy() error {
 
 func (c *SXIDComponent) updateCurrentState() error {
 	var rebootErr string
-	rebootEvents, err := getRebootEvents(c.rootCtx)
+	rebootEvents, err := os.GetRebootEvents(c.rootCtx, c.eventStore, time.Now().Add(-DefaultRetentionPeriod))
```
eahydra, line 222: it would be great if we can move the `GetRebootEvents` out of os component, refer to #510

eahydra (commented):
should we split the PR into two or three PRs?

components/accelerator/nvidia/sxid/component.go
gyuho: how about `pkg/reboot` and implement `record/read Reboot(ctx, eventStore)`

eahydra: yea, sgtm!

gyuho: @eahydra Done. Will make corresponding changes for control plane once this PR gets merged in.

pkg/reboot/event.go
```diff
@@ -0,0 +1,62 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/pkg/common"
+	"github.com/leptonai/gpud/pkg/eventstore"
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+func RecordEvent(ctx context.Context, eventStore eventstore.Store, bucketName string, lastRebootTime func(context.Context) (time.Time, error)) error {
```
eahydra, line 16: Let's try not to pass in function references, especially if it's just for testing. 
We should make the function `RecordEvent` private and should add functions like `UpdateOSRebootTime` or `SyncOSRebootTime`

```suggestion
func RecordEvent(ctx context.Context, eventStore eventstore.Store, bucketName string) error {
```

gyuho: @eahydra Done

components/accelerator/nvidia/sxid/component.go
```diff
@@ -217,7 +219,7 @@ func (c *SXIDComponent) SetHealthy() error {
 
 func (c *SXIDComponent) updateCurrentState() error {
 	var rebootErr string
-	rebootEvents, err := getRebootEvents(c.rootCtx)
+	rebootEvents, err := reboot.GetEvents(c.rootCtx, c.eventStore, "os", time.Now().Add(-DefaultRetentionPeriod))
```
eahydra, line 222: The caller should not know the `os`, even it should not pass the eventstore to the reboot package.. we can create a global reboot interface object like eventstore.. some static information in os component should be initialized once, and pass them for each component which depends on them.

components/os/component.go
```diff
@@ -3,140 +3,365 @@ package os
...
+		ts: time.Now().UTC(),
 	}
+	defer func() {
+		c.lastMu.Lock()
+		c.lastData = &d
+		c.lastMu.Unlock()
+	}()
+
+	cctx, ccancel := context.WithTimeout(c.ctx, 15*time.Second)
+	virtEnv, err := pkg_host.SystemdDetectVirt(cctx)
```
eahydra, line 114: We move all these static information (such as virtEnv, kernel version, SystemManufacturer, reboot time, host ID, arch, etc.) into separate packages and initialize them when gpud starts.

components/os/component.go
```diff
@@ -3,140 +3,365 @@ package os
...
+	now := time.Now().UTC()
+	d.Uptimes = Uptimes{
+		Seconds:             uptime,
+		SecondsHumanized:    humanize.RelTime(now.Add(time.Duration(-int64(uptime))*time.Second), now, "ago", "from now"),
+		BootTimeUnixSeconds: boottime,
+		BootTimeHumanized:   humanize.RelTime(time.Unix(int64(boottime), 0), now, "ago", "from now"),
+	}
+
+	cctx, ccancel = context.WithTimeout(c.ctx, 10*time.Second)
+	allProcs, err := process.CountProcessesByStatus(cctx)
```
eahydra, line 195: we only need to statistic the data in the function.

components/os/component.go
```diff
@@ -3,140 +3,365 @@ package os
...
+	ccancel()
+	if err != nil {
+		d.err = fmt.Errorf("failed to get boot time: %w", err)
+		return
+	}
+
+	now := time.Now().UTC()
+	d.Uptimes = Uptimes{
+		Seconds:             uptime,
+		SecondsHumanized:    humanize.RelTime(now.Add(time.Duration(-int64(uptime))*time.Second), now, "ago", "from now"),
```
eahydra, line 189: We don't need to format the timestamp unless we have to do it..

components/pci/component.go
```diff
@@ -3,69 +3,268 @@ package pci
...
-	getDefaultPoller().Start(cctx, cfg.Query, id.Name)
-
 	return &component{
-		cfg:         cfg,
-		rootCtx:     ctx,
+		ctx:         cctx,
 		cancel:      ccancel,
-		poller:      getDefaultPoller(),
 		eventBucket: eventBucket,
+		listFunc:    func(ctx2 context.Context) ([]pci.Device, error) { return pci.List(ctx2) },
```
eahydra, line 50:
```suggestion
		listFunc:    pci.List
```

gyuho: @eahydra fixed

components/pci/component.go
```diff
@@ -3,69 +3,258 @@ package pci
...
+
+func (d *Data) getReason() string {
+	if d == nil {
+		return "no pci data"
+	}
+	if d.err != nil {
+		return fmt.Sprintf("failed to get pci data -- %s", d.err)
+	}
+
+	return fmt.Sprintf("found %d devices", len(d.Devices))
```
eahydra, line 191: found % devices, What happened to these devices?? healthy or unhealthy?

components/pci/component.go
```diff
@@ -3,69 +3,258 @@ package pci
...
+func createEvent(time time.Time, devices []pci.Device) *components.Event {
+	uuids := make([]string, 0)
+	for _, dev := range devices {
+		// check whether ACS is enabled on PCI bridges
+		if dev.AccessControlService == nil {
+			continue
+		}
+		if dev.AccessControlService.ACSCtl.SrcValid {
+			uuids = append(uuids, dev.ID)
+		}
```
eahydra, line 247:
```suggestion
		if dev.AccessControlService != nil  && dev.AccessControlService.ACSCtl.SrcValid {
			uuids = append(uuids, dev.ID)
		}
```

pkg/config/default.go
```diff
@@ -142,7 +142,7 @@ func DefaultConfig(ctx context.Context, opts ...OpOption) (*Config, error) {
 	cfg.Components[network_latency_id.Name] = nil
 
 	if runtime.GOOS == "linux" {
```
eahydra, line 144: We don't need to check if current os is linux.. let the component decide whether to enable..

gyuho: done

components/pci/component.go
gyuho: updated

eahydra (approved):

PR #571, opened 2025-03-26
feat(fd, cpu, memory, remapped-rows, info, containerd, kubelet, docker, tailscale): do not mark unhealthy in case of context errors, fix degraded status set for "fd"

gyuho:
- **fix(cpu): only log context errors in /states**
- **feat(memory): handle context errors**
- **fix(tailscale): handle context errors (only logs)**
- **feat(fd): handle context errors, fix gethealth**
- **feat(info): handle context errors**
- **fix(remapped-rows): handle context errors**
- **fix(kubelet/pod): handle context errors**
- **feat(containerd): handle context errors**
- **feat(kernel-module, docker): handle context errors**

### DIFF
--- a/components/accelerator/nvidia/remapped-rows/component.go
+++ b/components/accelerator/nvidia/remapped-rows/component.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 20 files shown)


components/accelerator/nvidia/remapped-rows/component.go
```diff
@@ -284,6 +291,13 @@ func (d *Data) getReason() string {
 func (d *Data) getHealth() (string, bool) {
 	healthy := d == nil || d.err == nil
 
+	// if the context is canceled or timeout, we assume the next course of action is unknown
+	// thus, treating it as healthy (or transient failure)
+	if d != nil && !healthy && (errors.Is(d.err, context.DeadlineExceeded) || errors.Is(d.err, context.Canceled)) {
+		log.Logger.Warnw("check canceled or timeout -- transient error, please retry", "error", d.err)
+		healthy = true
```
eahydra, line 298: I am worried that such a modification may cause something that has been unhealthy to suddenly become healthy.
It would be ignore if the error is context deadline or canceled.   Should we check the error here? https://github.com/leptonai/gpud/blob/3abba8476b77ca9b99c03333b609b9392ea2f57e/pkg/session/serve.go#L432

gyuho: +1. Let me create a new PR.

PR #572, opened 2025-03-26
feat(fd, cpu, memory, remapped-rows, info, containerd, kubelet, docker, tailscale): return data fetch error

gyuho:
Replace: https://github.com/leptonai/gpud/pull/571

- **feat(memory): return error from data fetch**
- **feat(cpu): return data fetch error**
- **feat(fd): return data fetch error**
- **feat(info): return data fetch error**
- **feat(containerd): return error if found**
- **feat(remapped-rows): return data fetch error**
- **tailscale**

### DIFF
--- a/components/accelerator/nvidia/remapped-rows/component.go
+++ b/components/accelerator/nvidia/remapped-rows/component.go
@@ -304,6 +304,13 @@ func (d *Data) getHealth() (string, bool) {
 	return health, healthy
 }
 
+func (d *Data) getError() string {
+	if d == nil || d.err == nil {
+		return ""
+	}
+	return d.err.Error()
+}
+
 func (d *Data) getStates() ([]components.State, error) {
 	if d == nil {
 		return []components.State{
@@ -319,6 +326,7 @@ func (d *Data) getStates() ([]components.State, error) {
 	state := components.State{
 		Name:   "row_remapping",
 		Reason: d.getReason(),
+		Error:  d.getError(),
 	}
 	state.Health, state.Healthy = d.getHealth()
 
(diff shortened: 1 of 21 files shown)


components/accelerator/nvidia/remapped-rows/component.go
```diff
@@ -284,7 +284,7 @@ func (d *Data) getReason() string {
 func (d *Data) getHealth() (string, bool) {
 	healthy := d == nil || d.err == nil
 
-	if healthy && d.MemoryErrorManagementCapabilities.RowRemapping {
+	if d != nil && healthy && d.MemoryErrorManagementCapabilities.RowRemapping {
```
eahydra, line 287: Good catch!

pkg/session/serve.go
```diff
@@ -454,6 +454,15 @@ func (s *Session) getStatesFromComponent(ctx context.Context, componentName stri
...
 				currState.States[i].Healthy = true
 			}
+
+			if componentState.Error != "" &&
+				(strings.Contains(componentState.Error, context.DeadlineExceeded.Error()) ||
+					strings.Contains(componentState.Error, context.Canceled.Error())) {
+				log.Logger.Warnw("set healthy state due to deadline exceeded or canceled error", "component", componentName, "error", componentState.Error)
+
+				currState.States[i].Health = components.StateHealthy
+				currState.States[i].Healthy = true
```
eahydra, line 464: we cannot change the healthy state if the error is deadline exceeded or canceled. We should just keep silence. We can log the information with error level..

gyuho: just logging now

eahydra (approved):
lgtm

PR #562, opened 2025-03-24
feat(cpu, disk, fd, fuse, memory, network-latency): simplify metrics, fix races in metrics registration

gyuho:
Address https://github.com/leptonai/gpud/issues/559.

- **feat(disk): simplify metrics, fix race**
- **feat(fd): simplify metrics, fix races in metrics registration**
- **feat(memory): simplify metrics, remove metrics registration races**
- **feat(fuse): simplify poller, metrics, fix races in metrics registration**
- **feat(latency): refactor metrics, fix races**

PR #576, opened 2025-03-27
feat(metrics, latency): fix metrics package races, improve latency component

gyuho:
Split from https://github.com/leptonai/gpud/pull/562.

- **feat(pkg/metrics): new metrics package to fix races**
- **feat(metrics, latency): fix metrics package races, improve latency component**


components/network/latency/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/leptonai/gpud/components"
+	components_metrics_state "github.com/leptonai/gpud/pkg/gpud-metrics/state"
+	"github.com/leptonai/gpud/pkg/log"
+	components_metrics "github.com/leptonai/gpud/pkg/metrics"
+)
+
+const SubSystem = "network_latency"
+
+var (
+	lastUpdateUnixSeconds = prometheus.NewGauge(
```
eahydra, line 20: Why do we need the metric?

components/network/latency/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		prometheus.GaugeOpts{
+			Namespace: "",
+			Subsystem: SubSystem,
+			Name:      "last_update_unix_seconds",
+			Help:      "tracks the last update time in unix seconds",
+		},
+	)
+
+	edgeInMilliseconds = prometheus.NewGaugeVec(
```
eahydra, line 29: It's better to use histogram to track latency.

components/network/latency/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+			Namespace: "",
+			Subsystem: SubSystem,
+			Name:      "edge_in_milliseconds",
+			Help:      "tracks the edge latency in milliseconds",
+		},
+		[]string{"provider_region"},
+	)
+)
+
+var _ components.PromRegisterer = (*component)(nil)
```
eahydra, line 40:
```suggestion
var _ components.PromRegisterer = &component{}
```

components/network/latency/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (c *component) RegisterCollectors(reg *prometheus.Registry, dbRW *sql.DB, dbRO *sql.DB, tableName string) error {
+	c.metricsMu.Lock()
+	defer c.metricsMu.Unlock()
+
+	c.edgeInMillisecondsMetricsStore = components_metrics.NewStore(dbRW, dbRO, tableName, SubSystem+"_edge_in_milliseconds")
```
eahydra, line 46: It's not reasonable for creating metric store for each metric..  We should invert the dependency.. We can implement a global metric store, then implement a scraper or collector to collect all these metrics from prometheus.Registry, so that we can save all these metrics to the store.. And we can easily change the underlay store to TSDB in future...

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/577.

PR #586, opened 2025-03-31
feat(host, pci): move reboot utils, simplify

gyuho:


pkg/os/reboot_test.go
```diff
@@ -2,11 +2,16 @@ package reboot
 
 import (
 	"context"
+	"os"
 	"testing"
 	"time"
 )
 
 func TestReboot(t *testing.T) {
+	if os.Geteuid() == 0 { // running as root
```
eahydra, line 11: This test is invalid.. we can mock a reboot command like `echo rebooted` to test the function.

pkg/host/virt.go
```diff
@@ -34,6 +35,22 @@ type VirtualizationEnvironment struct {
 	IsKVM bool `json:"is_kvm"`
 }
 
+var currentVirtEnv VirtualizationEnvironment
+
+func init() {
```
eahydra, line 40: oh.. there is a package named host here.. we can move the files in os package to here.

pkg/host/virt.go
```diff
@@ -34,6 +35,22 @@ type VirtualizationEnvironment struct {
 	IsKVM bool `json:"is_kvm"`
 }
 
+var currentVirtEnv VirtualizationEnvironment
```
eahydra, line 38: we should define constant values for the following code in os component:
```go
o.VirtualizationEnvironment = virtEnv

		// for some reason, init failed
		if currentSystemManufacturer == "" && runtime.GOOS == "linux" {
			cctx, ccancel = context.WithTimeout(ctx, 20*time.Second)
			currentSystemManufacturer, err = pkg_host.SystemManufacturer(cctx)
			ccancel()
			if err != nil {
				log.Logger.Warnw("failed to get system manufacturer", "error", err)
			}
		}
		o.SystemManufacturer = currentSystemManufacturer

		o.MachineMetadata = currentMachineMetadata

		if err = createRebootEvent(ctx, eventBucket, reboot.LastReboot); err != nil {
			log.Logger.Warnw("failed to create reboot event", "error", err)
		}

		hostID, err := host.HostID()
		if err != nil {
			return nil, err
		}
		o.Host = Host{ID: hostID}

		arch, err := host.KernelArch()
		if err != nil {
			return nil, err
		}
		kernelVer, err := host.KernelVersion()
		if err != nil {
			return nil, err
		}
		o.Kernel = Kernel{Arch: arch, Version: kernelVer}

		platform, family, version, err := host.PlatformInformation()
		if err != nil {
			return nil, err
		}
		o.Platform = Platform{Name: platform, Family: family, Version: version}

		cctx, ccancel = context.WithTimeout(ctx, 10*time.Second)
		uptime, err := host.UptimeWithContext(cctx)
		ccancel()
		if err != nil {
			return nil, err
		}

		cctx, ccancel = context.WithTimeout(ctx, 10*time.Second)
		boottime, err := host.BootTimeWithContext(cctx)
		ccancel()
		if err != nil {
			return nil, err
		}
```

pkg/eventstore/types.go
```diff
@@ -13,6 +13,10 @@ const (
 
 type Store interface {
 	Bucket(name string) (Bucket, error)
+
+	// LoadBucketWithNoPurge loads the bucket with no purge run.
+	// This is useful for loading the bucket for read-only operations.
+	LoadBucketWithNoPurge(name string) (Bucket, error)
```
eahydra, line 19: It's better to use options pattern for `Bucket` method, therefore we don't need to add the new method.. and it's unreasonable.

pkg/host/event.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/leptonai/gpud/components"
+	"github.com/leptonai/gpud/pkg/common"
+	"github.com/leptonai/gpud/pkg/eventstore"
+	"github.com/leptonai/gpud/pkg/log"
+)
+
+const defaultBucketName = "os"
+
+type EventStore interface {
```
eahydra, line 18:
```suggestion
type RebootEventStore interface {
```

components/os/component_output.go
```diff
@@ -231,25 +230,25 @@ func init() {
...
 
-	currentMachineMetadata.OSMachineID, err = pkg_host.GetOSMachineID()
+	currentMachineMetadata.OSMachineID, err = pkghost.ReadOSMachineID()
 	if err != nil {
 		log.Logger.Warnw("failed to get os machine id", "error", err)
 	}
 
 	cctx, ccancel := context.WithTimeout(context.Background(), 20*time.Second)
-	currentSystemManufacturer, err = pkg_host.SystemManufacturer(cctx)
+	currentSystemManufacturer, err = pkghost.SystemManufacturer(cctx)
```
eahydra, line 251: Why not use CurrentSystemManufacturer?

components/os/component_output.go
```diff
@@ -231,25 +230,25 @@ func init() {
 
 	// Get boot ID (Linux-specific)
 	var err error
-	currentMachineMetadata.BootID, err = pkg_host.GetBootID()
+	currentMachineMetadata.BootID, err = pkghost.GetBootID()
```
eahydra, line 233: We should not call these functions. just call those functions which return static variable values.

components/os/component_output.go
gyuho: This `os` component will be refactored in the following PR :)

let me change here anyways

pkg/host/init.go
```diff
@@ -0,0 +1,151 @@
...
+
+func MachineID() string {
+	return currentMachineID
+}
+
+func DmidecodeUUID() string {
+	return currentDmidecodeUUID
+}
+
+func VirtEnv() VirtualizationEnvironment {
```
eahydra, line 141:
```suggestion
func VirtualizationEnv() VirtualizationEnvironment {
```

eahydra (approved):
overall lgtm

PR #587, opened 2025-03-31
feat(systemd): remove (in favor of individual component checks)

gyuho:


components/accelerator/nvidia/fabric-manager/component.go
```diff
@@ -76,7 +75,12 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 		}, nil
 	}
 
-	defaultConn := systemd.GetDefaultDbusConn()
+	cctx, cancel := context.WithCancel(c.rootCtx)
+	defer cancel()
+	defaultConn, err := pkg_systemd.NewDbusConn(cctx)
```
eahydra, line 80: Eventually we should remove the dependency on dbus..

eahydra (approved):
lgtm

components/accelerator/nvidia/fabric-manager/component.go
gyuho: yes, https://github.com/leptonai/gpud/issues/594

PR #588, opened 2025-03-31
feat(systemd): remove global/shared dbus connection object

gyuho:


gyuho:
Yes, closing in favor of https://github.com/leptonai/gpud/pull/587 and other following PRs

PR #589, opened 2025-03-31
feat(kubelet): check port open, list pods with failure threshold

gyuho:

### DIFF
--- a/components/kubelet/pod/component.go
+++ b/components/kubelet/pod/component.go
@@ -40,6 +40,7 @@ type component struct {
 	lastMu      sync.RWMutex
 	lastData    *Data
 	lastHealthy bool
+	lastReason  string
 }
 
 func New(ctx context.Context, kubeletReadOnlyPort int, ignoreConnectionErrors bool) components.Component {
@@ -82,7 +83,7 @@ func (c *component) States(ctx context.Context) ([]components.State, error) {
 	c.lastMu.RLock()
 	lastData := c.lastData
 	c.lastMu.RUnlock()
-	return lastData.getStates(c.failedCount, c.lastHealthy)
+	return lastData.getStates(c.lastReason, c.lastHealthy)
 }
 
 func (c *component) Events(ctx context.Context, since time.Time) ([]components.Event, error) {
@@ -118,11 +119,15 @@ func (c *component) CheckOnce() {
 
 	if c.checkDependencyInstalled == nil || !c.checkDependencyInstalled() {
 		// "kubelet" is not installed, thus not needed to check its activeness
+		c.lastHealthy = true
+		c.lastReason = "kubelet is not installed"
 		return
 	}
 
 	if c.checkKubeletRunning == nil || !c.checkKubeletRunning() {
 		// "kubelet" is not running, thus not needed to check its activeness
+		c.lastHealthy = true
+		c.lastReason = "kubelet is installed but not running"
 		return
 	}
 
@@ -138,29 +143,19 @@ func (c *component) CheckOnce() {
 	}
 
 	c.lastHealthy = d.err == nil
+	if d.err == nil {
+		c.lastReason = fmt.Sprintf("total %d pods (node %s)", len(d.Pods), d.NodeName)
+	}
+
 	if isConnectionRefusedError(d.err) && c.ignoreConnectionErrors {
+		// e.g.,
+		// Get "http://localhost:10255/pods": dial tcp [::1]:10255: connect: connection refused
 		c.lastHealthy = true
+		c.lastReason = "connection error but ignored"
 	} else if c.failedCount >= c.failedCountThreshold {
 		c.lastHealthy = false
+		c.lastReason = fmt.Sprintf("list pods from kubelet read-only port failed %d time(s)", c.failedCount)
 	}
-}
-
-func (d *Data) getHealth(ignoreConnErr bool, failedCount int, failedCountThreshold int) (string, bool) {
-	healthy := d == nil || d.err == nil
-
-	if d != nil && d.err != nil && isConnectionRefusedError(d.err) && ignoreConnErr {
-		healthy = true
-	}
-
-	if d != nil && failedCount >= failedCountThreshold {
-		healthy = false
-	}
-
-	health := components.StateHealthy
-	if !healthy {
-		health = components.StateUnhealthy
-	}
-	return health, healthy
 }
 
 type Data struct {
@@ -177,24 +172,6 @@ type Data struct {
 	err error `json:"-"`
 }
 
-func (d *Data) getReason(failedCount int) string {
-	if d == nil || (d.err == nil && len(d.Pods) == 0) {
-		return "no pod found or kubelet is not running"
-	}
-
-	if d.err != nil {
-		if isConnectionRefusedError(d.err) {
-			// e.g.,
-			// Get "http://localhost:10255/pods": dial tcp [::1]:10255: connect: connection refused
-			return fmt.Sprintf("connection error to node %q -- %v", d.NodeName, d.err)
-		}
-
-		return fmt.Sprintf("list pods from kubelet read-only port failed %d time(s) (%v)", failedCount, d.err)
-	}
-
-	return fmt.Sprintf("total %d pods (node %s)", len(d.Pods), d.NodeName)
-}
-
 func (d *Data) getError() string {
 	if d == nil || d.err == nil {
 		return ""
@@ -202,7 +179,7 @@ func (d *Data) getError() string {
 	return d.err.Error()
 }
 
-func (d *Data) getStates(failedCount int, lastHealthy bool) ([]components.State, error) {
+func (d *Data) getStates(lastReason string, lastHealthy bool) ([]components.State, error) {
 	if d == nil {
 		return []components.State{
 			{
@@ -216,7 +193,7 @@ func (d *Data) getStates(failedCount int, lastHealthy bool) ([]components.State,
 
 	state := components.State{
 		Name:   Name,
-		Reason: d.getReason(failedCount),
+		Reason: lastReason,
 		Error:  d.getError(),
 
 		Healthy: lastHealthy,
--- a/components/kubelet/pod/component.go
+++ b/components/kubelet/pod/component.go
@@ -4,18 +4,21 @@ package pod
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/leptonai/gpud/components"
 	"github.com/leptonai/gpud/pkg/log"
-	"github.com/leptonai/gpud/pkg/systemd"
+	"github.com/leptonai/gpud/pkg/netutil"
 )
 
-// Name is the ID of the kubernetes pod component.
-const Name = "kubelet"
+const (
+	// Name is the ID of the kubernetes pod component.
+	Name = "kubelet"
+
+	defaultFailedCountThreshold = 5
+)
 
 var _ components.Component = &component{}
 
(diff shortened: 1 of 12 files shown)


components/kubelet/pod/component.go
```diff
@@ -113,12 +113,17 @@ func (c *component) CheckOnce() {
...
+	if !netutil.IsPortOpen(c.kubeletReadOnlyPort) {
+		d.err = fmt.Errorf("kubelet is installed but kubelet read-only port %d is not open", c.kubeletReadOnlyPort)
+		return
+	}
+
 	// below are the checks in case "kubelet" is installed, thus requires activeness checks
-	cctx, ccancel := context.WithTimeout(c.ctx, 15*time.Second)
-	running := checkKubeletReadOnlyPortListening(cctx, c.kubeletReadOnlyPort)
+	cctx, ccancel := context.WithTimeout(c.ctx, 30*time.Second)
+	err := checkKubeletReadOnlyPortHealthzWithRetries(cctx, c.kubeletReadOnlyPort)
```
eahydra, line 123: We need to regularly check and record the number of recent consecutive failures. Only when the number exceeds a certain number can it be unhealthy, similar to the liveness probe. Like this function, although it is retried, all the failures encountered during this period may be failures, but during this period, the kubelet may be restarted and will recover soon.

components/kubelet/pod/component.go
```diff
@@ -113,12 +113,17 @@ func (c *component) CheckOnce() {
 		return
 	}
 
+	if !netutil.IsPortOpen(c.kubeletReadOnlyPort) {
```
eahydra, line 116: No need to check if kubelet opens the readonly port individually

components/kubelet/pod/component.go
eahydra, line 123: IMO even we don't need to call the function, we can just check if the status returned from `listPodsFromKubeletReadOnlyPort` is ok.

gyuho: This PR now removes this healthz checks. Do we want to have failure thresholds for list pods?

components/kubelet/pod/component.go
```diff
@@ -42,8 +41,8 @@ func New(ctx context.Context, kubeletReadOnlyPort int, ignoreConnectionErrors bo
 		cancel:                   cancel,
 		checkDependencyInstalled: checkKubeletInstalled,
 		kubeletReadOnlyPort:      kubeletReadOnlyPort,
-		checkServiceActive: func(ctx context.Context) (bool, error) {
-			return systemd.CheckServiceActive(ctx, "kubelet")
+		checkServiceActive: func(context.Context) (bool, error) {
```
eahydra, line 44: We don't even need to check if kubelet service is active.

components/kubelet/pod/component.go
eahydra: We should have failure thresholds. We cannot assume that kubelet readonly service is available if it's running.

components/kubelet/pod/component.go
```diff
@@ -108,35 +115,27 @@ func (c *component) CheckOnce() {
...
-	}
-
-	cctx, ccancel = context.WithTimeout(c.ctx, 30*time.Second)
+	// below are the checks in case "kubelet" is installed and running, thus requires activeness checks
+	cctx, ccancel := context.WithTimeout(c.ctx, 30*time.Second)
 	d.NodeName, d.Pods, d.err = listPodsFromKubeletReadOnlyPort(cctx, c.kubeletReadOnlyPort)
 	ccancel()
 
+	if d.err != nil {
+		c.failedCount++
```
eahydra, line 134: it's better to update the count in CheckOnce function.

components/kubelet/pod/component.go
```diff
@@ -168,17 +167,23 @@ func (d *Data) getReason() string {
...
+func (d *Data) getHealth(ignoreConnErr bool, failedCount int, failedCountThreshold int) (string, bool) {
 	healthy := d == nil || d.err == nil
+
 	if d != nil && d.err != nil && d.connErr && ignoreConnErr {
 		healthy = true
 	}
+
+	if d != nil && failedCount >= failedCountThreshold {
+		healthy = false
+	}
```
eahydra, line 185: we should move the logical to checkonce function... the data is just used to trace the last state, but only checkonce function is responsible for determine whether the kubelet is healthy...

components/kubelet/pod/component.go
gyuho: @eahydra This is already in CheckOnce function? https://github.com/leptonai/gpud/pull/589/files#diff-2aff32cd20735f318eff3f7d05a30e909e35b7e8176380e14127fb4827b8b5c4R107

components/kubelet/pod/component.go
gyuho: @eahydra Done

components/kubelet/pod/component.go
```diff
@@ -108,36 +116,51 @@ func (c *component) CheckOnce() {
...
+		c.failedCount++
+	} else {
+		c.failedCount = 0
+	}
+
+	c.lastHealthy = d.err == nil
+	if isConnectionRefusedError(d.err) && c.ignoreConnectionErrors {
+		c.lastHealthy = true
+	} else if c.failedCount >= c.failedCountThreshold {
+		c.lastHealthy = false
```
eahydra, line 144: Should we set the data.healthy = false and data.reason = fmt.Sprintf("failed to query pods information from kubelet %d times", c.failedCount)?   It seems the c.lastHealthy is unnecessary.

components/kubelet/pod/component.go
```diff
@@ -108,36 +116,51 @@ func (c *component) CheckOnce() {
...
 
-	d.connErr = isConnectionRefusedError(d.err)
+	if d.err != nil {
+		c.failedCount++
+	} else {
+		c.failedCount = 0
+	}
+
+	c.lastHealthy = d.err == nil
+	if isConnectionRefusedError(d.err) && c.ignoreConnectionErrors {
```
eahydra, line 141: I'm wondering which case should set `c.ignoreConnectionErrors=false`..  we cannot ignore connection errors since we have checked if the kubelet opens readonly port..

components/kubelet/pod/component.go
```diff
@@ -152,48 +175,34 @@ type Data struct {
...
 	if d.err != nil {
-		if d.connErr {
+		if isConnectionRefusedError(d.err) {
 			// e.g.,
 			// Get "http://localhost:10255/pods": dial tcp [::1]:10255: connect: connection refused
 			return fmt.Sprintf("connection error to node %q -- %v", d.NodeName, d.err)
 		}
 
-		return fmt.Sprintf("failed to list pods from kubelet read-only port -- %v", d.err)
+		return fmt.Sprintf("list pods from kubelet read-only port failed %d time(s) (%v)", failedCount, d.err)
```
eahydra, line 192: 1. we should generate the reason in the checkOnce before setting up lastData.
2. we should not generate the reason with d.err since d.err will be passed to State.Error which will be showed as MoreInformation in alert..

components/kubelet/pod/component.go
gyuho: The idea was, we don't want to alert when the kubelet restarts, but now that we check port open, think we can safely assume we don't need this?

/cc @cardyok

eahydra: There's a scene we might miss as we rely on checking port..

components/kubelet/pod/component.go
```diff
@@ -108,36 +117,45 @@ func (c *component) CheckOnce() {
...
+	if d.err != nil {
+		c.failedCount++
+	} else {
+		c.failedCount = 0
+	}
+
+	c.lastHealthy = d.err == nil
+	if d.err == nil {
+		c.lastReason = fmt.Sprintf("total %d pods (node %s)", len(d.Pods), d.NodeName)
+	}
```
eahydra, line 148: shall we do as following? 

```suggestion
	d.healthy = d.err == nil
	if d.err == nil {
		d.reason = fmt.Sprintf("total %d pods (node %s)", len(d.Pods), d.NodeName)
	}
```

components/kubelet/pod/component.go
```diff
@@ -108,36 +117,45 @@ func (c *component) CheckOnce() {
...
+
+	if isConnectionRefusedError(d.err) && c.ignoreConnectionErrors {
+		// e.g.,
+		// Get "http://localhost:10255/pods": dial tcp [::1]:10255: connect: connection refused
+		c.lastHealthy = true
+		c.lastReason = "connection error but ignored"
```
eahydra, line 154:
```suggestion
		d.healthy = true
		d.reason = "connection error but ignored"
```

components/kubelet/pod/component.go
```diff
@@ -108,36 +117,45 @@ func (c *component) CheckOnce() {
...
+	} else if c.failedCount >= c.failedCountThreshold {
+		c.lastHealthy = false
+		c.lastReason = fmt.Sprintf("list pods from kubelet read-only port failed %d time(s)", c.failedCount)
```
eahydra, line 157:
```suggestion
		d.healthy = false
		d.reason = fmt.Sprintf("list pods from kubelet read-only port failed %d time(s)", c.failedCount)
```

components/kubelet/pod/component.go
```diff
@@ -193,7 +179,7 @@ func (d *Data) getError() string {
 	return d.err.Error()
 }
 
-func (d *Data) getStates(ignoreConnErr bool) ([]components.State, error) {
+func (d *Data) getStates(lastReason string, lastHealthy bool) ([]components.State, error) {
```
eahydra, line 182:
```suggestion
func (d *Data) getStates() ([]components.State, error) {
```

components/kubelet/pod/component.go
```diff
@@ -207,10 +193,15 @@ func (d *Data) getStates(ignoreConnErr bool) ([]components.State, error) {
 
 	state := components.State{
 		Name:   Name,
-		Reason: d.getReason(),
+		Reason: lastReason,
```
eahydra, line 196:
```suggestion
		Reason: d.reason,
```

components/kubelet/pod/component.go
```diff
@@ -152,38 +170,6 @@ type Data struct {
 	ts time.Time `json:"-"`
 	// error from the last check
 	err error `json:"-"`
```
eahydra, line 172:
```suggestion
	err error `json:"-"`
	healthy bool
	reason string
```

components/kubelet/pod/component.go
```diff
@@ -207,10 +193,15 @@ func (d *Data) getStates(ignoreConnErr bool) ([]components.State, error) {
 
 	state := components.State{
 		Name:   Name,
-		Reason: d.getReason(),
+		Reason: lastReason,
 		Error:  d.getError(),
+
+		Healthy: lastHealthy,
```
eahydra, line 199:
```suggestion
		Healthy: d.healthy,
```

components/kubelet/pod/component.go
```diff
@@ -207,10 +193,15 @@ func (d *Data) getStates(ignoreConnErr bool) ([]components.State, error) {
...
+		Health:  components.StateHealthy,
+	}
+	if !lastHealthy {
```
eahydra, line 202:
```suggestion
	if !d.healthy{
```

components/kubelet/pod/component.go
eahydra: I thought about it carefully, let's do check if the kubelet bin is there..

gyuho: addressed

eahydra (approved):

PR #590, opened 2025-04-01
feat(os): simplify poller, remove os component dependency from sxid/xid

gyuho:
Cherry-picked from https://github.com/leptonai/gpud/pull/570.


eahydra (approved):
Awesome, it's much clearer than before.

PR #597, opened 2025-04-01
feat(cpu): simplify metrics, simplify states output and checks

gyuho:


components/cpu/metrics.go
```diff
@@ -0,0 +1,55 @@
...
+			Namespace: "",
+			Subsystem: SubSystem,
+			Name:      "used_percent",
+			Help:      "tracks the current file descriptor usage percentage",
+		},
+		[]string{pkgmetrics.MetricComponentLabelKey},
+	).MustCurryWith(componentLabel)
+)
+
+var _ components.PromRegisterer = (*component)(nil)
```
eahydra, line 41:
```suggestion
var _ components.PromRegisterer = &component{}
```

components/cpu/cpuinfo.go
```diff
@@ -2,55 +2,32 @@ package cpu
...
 }
 
-func getCores() (Cores, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
-	defer cancel()
-	logicalCores, err := cpu.CountsWithContext(ctx, true)
-	if err != nil {
-		return Cores{}, err
-	}
+func getCores() Cores {
```
eahydra, line 27: Shall we remove the function?

gyuho: done

eahydra (approved):

PR #601, opened 2025-04-01
feat(nvml): add library check utils

gyuho:
Will be useful for coming nvidia component refactoring.


eahydra:
Let's hold the PR. We'd like to review all the nvml related code and find out approaches to ensure init nvml once, don't repeatedly init or shutdown..

Issue #512, opened 2025-03-10
Deprecate dmesg in favor of kmsg


eahydra:
If the kmsg package is stable, we should deprecate dmesg in the next release.

PR #602, opened 2025-04-02
feat(xid,sxid): use kmsg instead of dmesg

eahydra:
1. Fix kmsg package incorrectly calculate boot time.
2. Use kmsg instead of dmesg in xid and sxid components.
3. Improve logs

for #512


eahydra:
@gyuho @cardyok TBR, PTAL.

cardyok (approved):
lgtm

gyuho (approved):

PR #603, opened 2025-04-02
feat: use kmsg instead of dmesg

eahydra:


eahydra:
Need to do more tests and update UT

pkg/kmsg/syncer.go
```diff
@@ -0,0 +1,107 @@
...
+
+			name, message := w.matchFunc(kmsg.Message)
+			if name == "" {
+				continue
+			}
+			event := components.Event{
+				Time:    metav1.Time{Time: kmsg.Timestamp.UTC()},
+				Name:    name,
+				Message: message,
+				Type:    common.EventTypeWarning,
```
gyuho, line 74: So, all events are warning by default now?

pkg/diagnose/scan.go
```diff
@@ -217,21 +216,19 @@ func Scan(ctx context.Context, opts ...OpOption) error {
 
 	if op.dmesgCheck {
```
gyuho, line 217: We should rename this to `kmsgCheck` (in a separate PR :))

gyuho (approved):

PR #606, opened 2025-04-02
feat(gpud-manager/systemd): auto-add --log-file flag with default file path

gyuho:

### DIFF
--- a/pkg/gpud-manager/systemd/systemd.go
+++ b/pkg/gpud-manager/systemd/systemd.go
@@ -2,8 +2,12 @@
 package systemd
 
 import (
+	"bufio"
 	_ "embed"
+	"errors"
+	"fmt"
 	"os"
+	"strings"
 )
 
 //go:embed gpud.service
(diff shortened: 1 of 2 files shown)


pkg/gpud-manager/systemd/systemd.go
```diff
@@ -36,3 +46,48 @@ FLAGS="--log-level=info --log-file=/var/log/gpud.log"
 `)
 	return err
 }
+
+func addLogFileFlagIfExists(file string) error {
+	if _, err := os.Stat(file); errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
```
eahydra, line 53: It has been checked whether the file exists at L34.
```suggestion
```

pkg/gpud-manager/systemd/systemd.go
```diff
@@ -36,3 +46,48 @@ FLAGS="--log-level=info --log-file=/var/log/gpud.log"
...
+			lines = append(lines, line)
+			continue
+		}
+
+		// remove the trailing " character
+		line = strings.TrimSuffix(line, "\"")
+		line = fmt.Sprintf("%s --log-file=/var/log/gpud.log\"", line)
+
+		lines = append(lines, line)
+	}
```
eahydra, line 83: it's better to wrap the code as a function.

gyuho: done

eahydra (approved):

Issue #593, opened 2025-04-01
Remove Component.Metrics method


gyuho:
In favor of global Prometheus syncer (https://github.com/leptonai/gpud/pull/577).

We can do this after https://github.com/leptonai/gpud/issues/574.

PR #610, opened 2025-04-03
feat(nvidia/temperature, power, util, ecc, gpm, gsp-firmware-mode, processes, persistence-mode, nvlink): simplify metrics, remove shared poller

gyuho:


components/accelerator/nvidia/ecc/component.go
```diff
@@ -3,156 +3,183 @@ package ecc
...
+		metricAggregateTotalUncorrected.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(float64(eccErrors.Aggregate.Total.Uncorrected))
+		metricVolatileTotalCorrected.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(float64(eccErrors.Volatile.Total.Corrected))
+		metricVolatileTotalUncorrected.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(float64(eccErrors.Volatile.Total.Uncorrected))
+
+		d.ECCErrors = append(d.ECCErrors, eccErrors)
 	}
 
-	return ms, nil
+	d.healthy = true
+	d.reason = fmt.Sprintf("checked %d devices for ECC errors", len(devs))
```
eahydra, line 130: How about `All 8 GPUs were checked, and no ECC errors were found`?

components/accelerator/nvidia/ecc/component.go
```diff
@@ -3,156 +3,183 @@ package ecc
...
+	// timestamp of the last check
+	ts time.Time `json:"-"`
+	// error from the last check
+	err error `json:"-"`
 
-	return nil
+	// tracks the healthy evaluation result of the last check
+	healthy bool `json:"-"`
+	// tracks the reason of the last check
+	reason string `json:"-"`
```
eahydra, line 145: Actually we don't need to set json tags for these fields since they are private, json package cannot access them. IDE like goland will report it as warning.

components/accelerator/nvidia/memory/component.go
```diff
@@ -3,145 +3,180 @@ package memory
...
+			return
+		}
+		metricUsedPercent.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(usedPct)
+
+		d.Memories = append(d.Memories, mem)
 	}
 
-	return ms, nil
+	d.healthy = true
+	d.reason = fmt.Sprintf("checked %d devices for memory", len(devs))
```
eahydra, line 128: we should also improve the message.

components/accelerator/nvidia/power/component.go
```diff
@@ -3,144 +3,178 @@ package power
...
+			return
+		}
+		metricUsedPercent.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(usedPct)
+
+		d.Powers = append(d.Powers, power)
 	}
 
-	return ms, nil
+	d.healthy = true
+	d.reason = fmt.Sprintf("checked %d devices for power", len(devs))
```
eahydra, line 126: and here

components/accelerator/nvidia/memory/component.go
```diff
@@ -3,145 +3,180 @@ package memory
...
-	log.Logger.Debugw("closing component")
+type Data struct {
+	Memories []nvidianvml.Memory `json:"memories,omitempty"`
 
-	// safe to call stop multiple times
-	_ = c.poller.Stop(Name)
+	// timestamp of the last check
+	ts time.Time `json:"-"`
+	// error from the last check
+	err error `json:"-"`
```
eahydra, line 137:
```suggestion
	ts time.Time
	// error from the last check
	err error
```

components/accelerator/nvidia/memory/component.go
```diff
@@ -3,145 +3,180 @@ package memory
...
 
-	return nil
+	// tracks the healthy evaluation result of the last check
+	healthy bool `json:"-"`
+	// tracks the reason of the last check
+	reason string `json:"-"`
```
eahydra, line 142:
```suggestion
	healthy bool
	// tracks the reason of the last check
	reason string
```

pkg/diagnose/scan.go
```diff
@@ -165,31 +165,6 @@ func Scan(ctx context.Context, opts ...OpOption) error {
 					} else {
 						log.Logger.Warnw("failed to check gpm supported or not", "error", err)
 					}
-
-					if gpmSupported {
```
eahydra, line 169: if we delete the code, we'd remove all related code? from L158  to L167

gyuho: @eahydra Yes, this PR already deletes `RecvGPMEvents` and such.

eahydra (approved):

PR #213, opened 2024-11-29
feat(components/memory): track current jit alloc buffer size, vm alloc status

gyuho:
```
Events:
  Type     Reason                  Age                    From     Message
  ----     ------                  ----                   ----     -------
  Warning  FailedCreatePodSandBox  3m11s (x350 over 88m)  kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to init seccomp: error loading seccomp filter into kernel: err…
```

> FailedCreatePodSandBox  3m11s (x350 over 88m)  kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to init seccomp: error loading seccomp filter into kernel: err…

c.f., https://github.com/deckhouse/deckhouse/issues/7402

PR #611, opened 2025-04-03
feat(memory): simplify metrics using Prometheus-based ones

gyuho:


components/memory/component.go
```diff
@@ -148,62 +116,50 @@ func (c *component) CheckOnce() {
...
-	if err != nil {
-		d.err = err
-		return
-	}
-
-	metrics.SetFreeBytes(float64(vm.Free))
-
-	cctx, ccancel = context.WithTimeout(c.ctx, 5*time.Second)
-	bpfJITBufferBytes, err := getCurrentBPFJITBufferBytes(cctx)
+	bpfJITBufferBytes, err := c.getCurrentBPFJITBufferBytesFunc(cctx)
```
eahydra, line 143: What is bpfJITBufferBytes used for?

gyuho: @eahydra Some contexts https://github.com/awslabs/amazon-eks-ami/issues/1179 and https://github.com/leptonai/gpud/pull/213

eahydra (approved):

PR #612, opened 2025-04-04
feat(nvidia/remapped-rows): simplify metrics

gyuho:


components/accelerator/nvidia/remapped-rows/component.go
```diff
@@ -31,25 +30,30 @@ type component struct {
...
 
-	getRemappedRowsFunc func(uuid string, dev device.Device) (nvml.RemappedRows, error)
+	eventBucket eventstore.Bucket
 
 	lastMu   sync.RWMutex
 	lastData *Data
 }
 
-func New(ctx context.Context, nvmlInstance nvml.InstanceV2, eventBucket eventstore.Bucket) (components.Component, error) {
+func New(ctx context.Context, getDevicesFunc func() map[string]device.Device, getProductNameFunc func() string, getMemoryErrorManagementCapabilitiesFunc func() nvml.MemoryErrorManagementCapabilities, eventBucket eventstore.Bucket) components.Component {
```
eahydra, line 44: It's unnecessary that pass these function references from outside. actually these functions should be replaced by nvml.InstanceV2, and we can mock InstanceV2 to do more tests

eahydra (approved):

PR #614, opened 2025-04-06
feat(nvidia/xid): mark Xid 143 as fatal (gpu init failure)

gyuho:

### DIFF
--- a/pkg/nvidia-query/xid/xid.go
+++ b/pkg/nvidia-query/xid/xid.go
@@ -4922,11 +4922,24 @@ Report a GPU issue and reset GPU(s) reporting the XID (refer to GPU reset capabi
 		Name:            "GPU Initialization Failure",
 		Description:     "",
 
-		SuggestedActionsByGPUd:    nil,
-		CriticalErrorMarkedByGPUd: false,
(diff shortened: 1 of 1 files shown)


pkg/nvidia-query/xid/xid.go
```diff
@@ -4922,11 +4922,24 @@ Report a GPU issue and reset GPU(s) reporting the XID (refer to GPU reset capabi
...
+
+			RepairActions: []common.RepairActionType{
+				common.RepairActionTypeHardwareInspection,
+			},
+		},
+		CriticalErrorMarkedByGPUd: true,
 
 		// Xids whose GPUd.RepairActions is empty
-		EventType: common.EventTypeWarning,
+		EventType: common.EventTypeFatal,
```
eahydra, line 4942: We set the `PotentialFBCorruption:           true` for XID 143. IMO we should set false?

eahydra (approved):

pkg/nvidia-query/xid/xid.go
gyuho: discussed offline, we are using the value from https://docs.nvidia.com/deploy/xid-errors/

PR #615, opened 2025-04-07
feat(nvidia): remove shared poller package

gyuho:
Fix https://github.com/leptonai/gpud/issues/574.


pkg/server/server.go
```diff
@@ -94,13 +91,12 @@ type Server struct {
...
-	nvidiaComponentsExist bool
-	uid                   string
-	fifoPath              string
-	fifo                  *goOS.File
-	session               *session.Session
-	enableAutoUpdate      bool
-	autoUpdateExitCode    int
+	uid                string
+	fifoPath           string
+	fifo               *goOS.File
```
eahydra, line 96: `goOS` is weird.. should we keep the same style as others?

eahydra (approved):

Issue #472, opened 2025-02-27
[LEP-341] Increase GPUd test coverage ~85%


gyuho:
ref. https://app.codecov.io/gh/leptonai/gpud/tree/main

PR #616, opened 2025-04-07
feat(pkg/login): clean up, increase unit test coverage

gyuho:
from 40% to 89%

### DIFF
--- a/cmd/gpud/command/login.go
+++ b/cmd/gpud/command/login.go
@@ -82,7 +82,7 @@ func cmdLogin(cliContext *cli.Context) error {
 		if err != nil {
 			hostname = "UnknownName"
 		}
-		if err := login.Login(hostname, token, fmt.Sprintf("https://%s/api/v1/login", endpoint), components, uid); err != nil {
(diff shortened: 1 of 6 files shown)


cmd/gpud/command/login.go
```diff
@@ -82,7 +82,7 @@ func cmdLogin(cliContext *cli.Context) error {
 		if err != nil {
 			hostname = "UnknownName"
 		}
-		if err := login.Login(hostname, token, endpoint, components, uid); err != nil {
+		if err := login.Login(hostname, token, fmt.Sprintf("https://%s/api/v1/login", endpoint), components, uid); err != nil {
```
eahydra, line 85: It's better to hide the URL in login package. Other components should not care about what it is.

eahydra:
We should revisit the client code related to the gpud control plane.  There should be a package, which is responsibility for talk with control-plane. There should be a Client type, which can be reused and hide all internal detailed information.

cmd/gpud/command/login.go
gyuho: @eahydra Now it's all internal to login package.

eahydra (approved):

PR #617, opened 2025-04-07
feat(cpu, fd): clean up init, add unit tests

gyuho:


components/cpu/component.go
```diff
@@ -131,26 +145,30 @@ func (c *component) CheckOnce() {
 
 	curStat, usedPct, err := calculateCPUUsage(
 		c.ctx,
-		getPrevTimeStat(),
-		getTimeStatForAllCPUs,
-		getUsedPercentForAllCPUs,
+		c.getPrevTimeStatFunc(),
+		c.getTimeStatFunc,
```
eahydra, line 149: I'm curious that why you don't just pass the results of these functions..

eahydra (approved):

Issue #559, opened 2025-03-24
FUSE component has data race


eahydra:
https://github.com/leptonai/gpud/blob/45ac74a7ad6209f3ebd44dbb54c4142a982cd4e2/components/fuse/metrics/metrics.go#L50-L53

There is a race condition there that needs to be fixed.

PR #618, opened 2025-04-07
feat(server): deprecate per-component metrics backend for global prometheus registry, support component selection in metrics store

gyuho:

Fix https://github.com/leptonai/gpud/issues/559.


.github/workflows/tests-e2e.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - name: run e2e tests
         run: |
           KMSG_FILE_PATH=/dev/null ./scripts/tests-e2e.sh
-      - name: run e2e tests (experimental)
-        run: |
-          ENABLE_EXPERIMENTAL_FEATURES=true KMSG_FILE_PATH=/dev/null ./scripts/tests-e2e.sh
+
+      # - name: run e2e tests (experimental)
+      #   run: |
+      #     ENABLE_EXPERIMENTAL_FEATURES=true KMSG_FILE_PATH=/dev/null ./scripts/tests-e2e.sh
```
eahydra, line 36: should we remove these lines?

pkg/metrics/options.go
```diff
@@ -0,0 +1,41 @@
...
+func WithComponents(components ...string) OpOption {
+	return func(op *Op) {
+		for _, component := range components {
+			if len(component) == 0 {
+				continue
+			}
+
+			if op.SelectedComponents == nil {
+				op.SelectedComponents = make(map[string]struct{})
+			}
```
eahydra, line 37: should move these lines before the for loop

pkg/session/serve.go
```diff
@@ -103,39 +104,33 @@ func (s *Session) serve() {
 			for _, componentName := range payload.Components {
```
eahydra, line 104: we should refactor the for loop:
1. wrap it as a function
2. get component by name, and check if the component implements SetHealthy method, if so, call it, otherwise skip it.

gyuho: This requires `unwrap` method since `getComponent` itself returns interface... and `unwrap` is being removed as we remove watchable component interface (since it's too complicated). Do we want to add back?

components/accelerator/nvidia/clock-speed/metrics.go
```diff
@@ -40,22 +34,7 @@ var (
...
-}
-
-// TO BE DEPRECATED
-func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.Metric, error) {
-	log.Logger.Debugw("querying metrics", "since", since)
-
-	return nil, nil
+func init() {
+	prometheus.MustRegister(metricGraphicsMHz)
+	prometheus.MustRegister(metricMemoryMHz)
```
eahydra, line 39: We should define a prometheus registry in metric package, each component should register metrics into it.
```suggestion
	prometheus.MustRegister(
	         metricGraphicsMHz, 
	         metricMemoryMHz,
	)
```

components/components.go
```diff
@@ -47,19 +44,11 @@ type Component interface {
...
-	// Returns all the metrics from the component.
-	Metrics(ctx context.Context, since time.Time) ([]Metric, error)
-
 	// Called upon server close.
 	// Implements copmonent-specific poller cleanup logic.
 	Close() error
 }
 
-// Defines an optional component interface that supports Prometheus metrics.
-type PromRegisterer interface {
```
eahydra, line 59: WatchableComponent should be deleted together.

pkg/metrics/options.go
```diff
@@ -0,0 +1,40 @@
...
+func WithComponents(components ...string) OpOption {
+	return func(op *Op) {
+		if len(components) > 0 && op.SelectedComponents == nil {
+			op.SelectedComponents = make(map[string]struct{})
+		}
+		for _, component := range components {
+			if len(component) == 0 {
+				continue
+			}
+			op.SelectedComponents[component] = struct{}{}
```
eahydra, line 37:
```suggestion
			if component != "" {
				op.SelectedComponents[component] = struct{}{}
			}
```

pkg/session/serve.go
eahydra: I don't think we need the `Unwrap` method even `getComponent` returns interface since type assert checks if the underlay type supports target interface.

components/accelerator/nvidia/clock-speed/metrics.go
gyuho: done

pkg/session/serve.go
gyuho: Yeah understood your comment now, fixed.

eahydra (approved):

PR #622, opened 2025-04-07
feat(fabric-manager): simplify to remove dbus library dependency

gyuho:
Fix https://github.com/leptonai/gpud/pull/622.


components/accelerator/nvidia/fabric-manager/component.go
```diff
@@ -138,18 +113,127 @@ func (c *component) Close() error {
...
+		"encoding": "json",
+	}
+	return []components.State{state}, nil
+}
+
+// TO BE DEPRECATED
+func (c *component) Metrics(ctx context.Context, since time.Time) ([]components.Metric, error) {
+	log.Logger.Debugw("querying metrics", "since", since)
+
+	return nil, nil
```
eahydra, line 238: we can remove it now.

components/tailscale/component.go
```diff
@@ -35,8 +34,8 @@ func New(ctx context.Context) components.Component {
 		ctx:                      cctx,
 		cancel:                   cancel,
 		checkDependencyInstalled: checkTailscaledInstalled,
-		checkServiceActive: func(ctx context.Context) (bool, error) {
-			return systemd.CheckServiceActive(ctx, "tailscaled")
+		checkServiceActiveFunc: func() (bool, error) {
+			return systemd.IsActive("tailscaled")
```
eahydra, line 38: There is an implicit dependency on dbus in IsActive... We should find another way to detect if a component is alive. 
We can add a TODO to solve later.

pkg/systemd/systemd.go
```diff
@@ -108,6 +108,7 @@ func parseVersion(version string) (string, []string) {
 }
 
 // IsActive returns true if the systemd service is active.
+// TODO: deprecate this
```
eahydra, line 111: Should we delete the function GetLatestJournalctlOutput and its tests?

PR #624, opened 2025-04-07
feat(components/*): increase unit test coverage, make healthy/reason check consistent

gyuho:

### DIFF
--- a/components/kernel-module/component.go
+++ b/components/kernel-module/component.go
@@ -19,8 +19,8 @@ const Name = "kernel-module"
 var _ components.Component = &component{}
 
 type component struct {
-	getAllModules  func() ([]string, error)
-	modulesToCheck []string
+	getAllModulesFunc func() ([]string, error)
+	modulesToCheck    []string
 
 	lastMu   sync.RWMutex
 	lastData *Data
@@ -28,8 +28,8 @@ type component struct {
 
 func New(modulesToCheck []string) components.Component {
 	return &component{
-		getAllModules:  getAllModules,
-		modulesToCheck: modulesToCheck,
+		getAllModulesFunc: getAllModules,
+		modulesToCheck:    modulesToCheck,
 	}
 }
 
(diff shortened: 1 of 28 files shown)


components/kernel-module/component.go
```diff
@@ -67,71 +67,50 @@ func (c *component) CheckOnce() {
...
 
-func (d *Data) getHealth(modulesToCheck []string) (string, bool) {
-	healthy := d == nil || d.err == nil || len(modulesToCheck) == 0
-	health := components.StateHealthy
-	if !healthy {
-		health = components.StateUnhealthy
-	}
+type Data struct {
+	LoadedModules []string            `json:"loaded_modules"`
+	loadedModules map[string]struct{} `json:"-"`
```
eahydra, line 103:
```suggestion
	loadedModules map[string]struct{} 
```

components/network/latency/component.go
```diff
@@ -48,6 +50,10 @@ func New(ctx context.Context) components.Component {
 		ctx:    cctx,
 		cancel: ccancel,
 
+		getEgressLatenciesFunc: func(ctx2 context.Context) (latency.Latencies, error) {
+			return latency_edge.Measure(ctx2)
+		},
```
eahydra, line 55:
```suggestion
		getEgressLatenciesFunc: func(ctx context.Context) (latency.Latencies, error) {
			return latency_edge.Measure(ctx)
		},
```

components/network/latency/component.go
```diff
@@ -104,21 +110,34 @@ func (c *component) CheckOnce() {
 		c.lastMu.Unlock()
 	}()
 
-	cctx, ccancel := context.WithTimeout(c.ctx, 30*time.Second)
+	cctx, ccancel := context.WithTimeout(c.ctx, time.Minute)
```
eahydra, line 113: I'm curious why we change the timeout from 30s to 60s? 
```suggestion
	ctx, cancel := context.WithTimeout(c.ctx, time.Minute)
```

components/pci/component.go
```diff
@@ -256,16 +238,21 @@ func (d *Data) getStates() ([]components.State, error) {
 	return []components.State{state}, nil
 }
 
-func (d *Data) createEvent(time time.Time) *components.Event {
-	uuids := d.listACSEnabledDeviceUUIDs()
-	if len(uuids) == 0 {
+func (d *Data) listACSEnabledDeviceUUIDs() []string {
```
eahydra, line 241: It seems that we can make it as a freedom function, not necessary to make it as a method of Data.

PR #625, opened 2025-04-08
feat(server): remove local web UI

gyuho:
To prioritize other stability features


cmd/gpud/command/command.go
```diff
@@ -52,6 +48,8 @@ var (
 	ibstatCommand string
 
 	checkInfiniBand bool
+
+	deprecatedWebEnable bool
```
eahydra, line 52: No one references it, shall we remove it ?

eahydra (approved):
overall lgtm

cmd/gpud/command/command.go
gyuho: Some GPUd commands references it. After 0.5 release, we can drop this :)

Issue #568, opened 2025-03-25
Support external plugin management


eahydra:
We should support external plugin management to help users easily run their own scripts or bin to detect issues.

PR #627, opened 2025-04-09
feat(custom-plugins): implement server + client

gyuho:
Address https://github.com/leptonai/gpud/issues/568.

Some inspiration from https://docs.aws.amazon.com/systems-manager/latest/userguide/documents-schemas-features.html.

### DIFF
--- a/pkg/session/serve.go
+++ b/pkg/session/serve.go
@@ -15,6 +15,7 @@ import (
 	apiv1 "github.com/leptonai/gpud/api/v1"
 	"github.com/leptonai/gpud/components"
 	componentsnvidiainfiniband "github.com/leptonai/gpud/components/accelerator/nvidia/infiniband"
+	pkgcustomplugins "github.com/leptonai/gpud/pkg/custom-plugins"
 	"github.com/leptonai/gpud/pkg/errdefs"
 	pkghost "github.com/leptonai/gpud/pkg/host"
 	"github.com/leptonai/gpud/pkg/log"
@@ -38,6 +39,11 @@ type Request struct {
 	UpdateVersion string            `json:"update_version,omitempty"`
 	UpdateConfig  map[string]string `json:"update_config,omitempty"`
 	Bootstrap     *BootstrapRequest `json:"bootstrap,omitempty"`
+
+	// CustomPluginSpec is the spec for the custom plugin to register or update.
+	CustomPluginSpec *pkgcustomplugins.Spec `json:"custom_plugin_spec,omitempty"`
+	// CustomPluginName is the name of the custom plugin to deregister.
+	CustomPluginName string `json:"custom_plugin_name,omitempty"`
 }
 
 type Response struct {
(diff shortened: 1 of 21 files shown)


pkg/plugins/plugins.go
```diff
@@ -0,0 +1,124 @@
...
+	// If zero, it runs only once.
+	Interval metav1.Duration `json:"interval"`
+}
+
+type Job struct {
+	// Script is the script to run for this job.
+	// Assumed to be base64 encoded.
+	Script string `json:"script"`
+	// base64 decoded state script
+	scriptDecoded string
```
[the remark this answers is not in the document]
gyuho: removed

pkg/plugins/plugins.go
```diff
@@ -0,0 +1,124 @@
...
+	// Name is the name of the plugin.
+	// Does not allow whitespace characters.
+	Name string `json:"name"`
+	// GPUd canonical component name based on user-provided plugin name
+	componentName string
+
+	// StateJob represents the jobs to run for /states API.
+	StateJob *Job `json:"state_job,omitempty"`
+	// EventJob represents the jobs to run for /events API.
+	EventJob *Job `json:"event_job,omitempty"`
```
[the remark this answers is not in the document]
gyuho: Changed to `StatePlugin`. Want to have it as a `Plugin` struct for more advanced spec info (e.g., in the future, dependency installs, or something like `runtime: containerd`

pkg/session/serve.go
```diff
@@ -35,6 +37,7 @@ type Request struct {
 	Since         time.Duration     `json:"since"`
 	UpdateVersion string            `json:"update_version,omitempty"`
 	UpdateConfig  map[string]string `json:"update_config,omitempty"`
+	Plugin        *plugins.Plugin   `json:"plugin,omitempty"`
```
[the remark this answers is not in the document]
gyuho: Removing for now

pkg/session/serve.go
```diff
@@ -44,6 +47,8 @@ type Response struct {
 	States  v1.LeptonStates  `json:"states,omitempty"`
 	Events  v1.LeptonEvents  `json:"events,omitempty"`
 	Metrics v1.LeptonMetrics `json:"metrics,omitempty"`
+
+	Plugins plugins.Plugins `json:"plugins,omitempty"`
```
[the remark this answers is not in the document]
gyuho: Removing for now

pkg/custom-plugins/plugins.go
```diff
@@ -0,0 +1,117 @@
...
+	// If zero, it runs only once.
+	Interval metav1.Duration `json:"interval"`
+}
+
+// Plugin represents a plugin spec.
+type Plugin struct {
+	// Script is the script to run for this job.
+	// Assumed to be base64 encoded.
+	Script string `json:"script"`
+}
```
gyuho, line 44: Defining as a struct, in case we want to support container-based executions in the future:

```diff
--- a/pkg/custom-plugins/plugins.go
+++ b/pkg/custom-plugins/plugins.go
@@ -41,6 +41,10 @@ type Plugin struct {
        // Script is the script to run for this job.
        // Assumed to be base64 encoded.
        Script string `json:"script"`
+
+       Envs           map[string]string `json:"envs"`
+       ContainerImage string            `json:"container_image"`
+       ContainerArgs  []string          `json:"container_args"`
 }
```

pkg/custom-plugins/spec.go
```diff
@@ -0,0 +1,131 @@
...
+			return nil, err
+		}
+	}
+
+	return pluginSpecs, nil
+}
+
+// Spec is a plugin spec and configuration.
+// Each spec represents a single state or event, in the external-plugin component.
+type Spec struct {
```
eahydra, line 57: shall we move it into types.go and reorder the type and functions to keep a consistent view to help understand the relationship between them?

pkg/custom-plugins/plugin.go
```diff
@@ -0,0 +1,88 @@
...
+// Plugin represents a plugin spec.
+type Plugin struct {
+	// Steps is a sequence of steps to run for this plugin.
+	// The steps are executed in order.
+	// If a step fails, the execution stops and the error is returned.
+	Steps []PluginStep `json:"steps,omitempty"`
+}
+
+// Validate validates all the plugin steps.
+func (p Plugin) Validate() error {
```
eahydra, line 19: Let's use pointer receiver first.

```suggestion
func (p *Plugin) Validate() error {
```

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+	cr.Output, cr.ExitCode, cr.err = c.spec.StatePlugin.executeAllSteps(cctx)
+
+	if cr.err != nil {
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = fmt.Sprintf("error executing state plugin -- %s", cr.err)
+		return cr
+	}
+
+	cr.health = apiv1.HealthStateTypeHealthy
+	cr.reason = string(cr.Output)
```
eahydra, line 120: if it's healthy, we would like to not record the output unless we can limit the maximum length of the output.

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	cctx, ccancel := context.WithTimeout(c.ctx, c.spec.Timeout.Duration)
+	defer ccancel()
+
+	cr.Output, cr.ExitCode, cr.err = c.spec.StatePlugin.executeAllSteps(cctx)
+
+	if cr.err != nil {
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = fmt.Sprintf("error executing state plugin -- %s", cr.err)
```
eahydra, line 115: I remembered that we required plugin outputs some special key/value pairs to indicated the reason and health. But I didn't say the part.

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+				Name:      "custom-plugin",
+				Health:    apiv1.HealthStateTypeHealthy,
+				Reason:    "no data yet",
+			},
+		}
+	}
+
+	state := apiv1.HealthState{
+		Component: "custom-plugin",
+		Name:      cr.name,
```
eahydra, line 205: I didn't find where the field is populated

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+			cancel: ccancel,
+			spec:   spec,
+		}
+		return c, nil
+	}
+}
+
+var _ components.Component = &component{}
+
+type component struct {
```
eahydra, line 32: shall we rename the filename to component.go as others?

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	state := apiv1.HealthState{
+		Component: "custom-plugin",
+		Name:      cr.name,
+		Reason:    cr.reason,
+		Error:     cr.getError(),
+		Health:    cr.health,
+	}
+	return apiv1.HealthStates{state}
```
eahydra, line 210: Do users need to configure 3 specs if users have 3 scripts?

pkg/custom-plugins/spec_component.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return cr.err.Error()
+}
+
+func (cr *checkResult) getLastHealthStates() apiv1.HealthStates {
+	if cr == nil {
+		return apiv1.HealthStates{
+			{
+				Component: "custom-plugin",
+				Name:      "custom-plugin",
```
eahydra, line 196: We should keep the name as user defined and keep the same as L205 as well.

pkg/custom-plugins/spec_component.go
gyuho: Yes, one spec == one plugin.

One spec/plugin may have multiple steps (e.g., install dependency, run script)

pkg/custom-plugins/spec_component.go
gyuho: oh here, the `cr` is `nil`...

pkg/custom-plugins/spec_component.go
gyuho: done

pkg/custom-plugins/spec_component.go
gyuho: fixed

pkg/custom-plugins/spec_component.go
gyuho: done

pkg/custom-plugins/spec_component.go
gyuho: Meaning, default mode, is very simple. Just a script + exit code check. The user can define their own exit code. Once it's working, we can define GPUd specific output parsing logic.

eahydra (approved):

Issue #511, opened 2025-03-10
Improve initialization code path


eahydra:
1. Define and implement the GPUdHandle interface, which includes support for other components to obtain some shared objects, such as databases, eventstores, etc.
2. Define component factory protype.
3. Spilt the component interface
4. Remove unnecessary configuration
5. Define component registry

gyuho:
This should be resolved by https://github.com/leptonai/gpud/pull/651.

PR #628, opened 2025-04-09
feat(components): add unit tests, set component method, move common pkg to api/v1

gyuho:

c.f., https://github.com/leptonai/gpud/pull/627


components/components.go
```diff
@@ -2,144 +2,133 @@
...
+	defaultSetMu sync.RWMutex
+	defaultSet   = make(map[string]Component)
+)
 
-	SuggestedActions *common.SuggestedActions `json:"suggested_actions,omitempty"`
+// GetAllComponents returns all the components in the default set.
+func GetAllComponents() map[string]Component {
+	defaultSetMu.RLock()
+	defer defaultSetMu.RUnlock()
+	return defaultSet
```
eahydra, line 20: It's not better to directly return the map since it may bring potential races. shall we return a cloned map?

components/components.go
```diff
@@ -2,144 +2,133 @@
...
-type Metric struct {
-	UnixSeconds         int64   `json:"unix_seconds"`
-	MetricName          string  `json:"metric_name"`
-	MetricSecondaryName string  `json:"metric_secondary_name,omitempty"`
-	Value               float64 `json:"value"`
+// isComponentRegistered checks if a component is registered in the default set.
+func isComponentRegistered(set map[string]Component, name string) bool {
+	_, ok := set[name]
+	return ok
 }
```
eahydra, line 35: Why do we need the function? Shall we delete it?

components/components.go
```diff
@@ -2,144 +2,133 @@
...
-	defaultSet   = make(map[string]Component)
-)
-
-func IsComponentRegistered(name string) bool {
-	defaultSetMu.RLock()
-	defer defaultSetMu.RUnlock()
+// getComponent gets a component from the default set.
+// It returns an error if the component is not found.
+func getComponent(set map[string]Component, name string) (Component, error) {
+	if set == nil {
```
eahydra, line 49: nil map is read safe.

components/components.go
```diff
@@ -2,144 +2,133 @@
...
 	defaultSetMu.Lock()
 	defer defaultSetMu.Unlock()
 
-	if defaultSet == nil {
+	return registerComponent(defaultSet, comp)
+}
+
+// registerComponent registers a component in the default set.
+// It returns an error if the component is already registered.
+func registerComponent(set map[string]Component, comp Component) error {
```
eahydra, line 71: and we don't need it as well...

components/components.go
```diff
@@ -2,144 +2,133 @@
...
+
 	return nil
 }
 
-func GetComponent(name string) (Component, error) {
-	defaultSetMu.RLock()
-	defer defaultSetMu.RUnlock()
+// StopDeregisterComponent unregisters and stops a component from the default set.
+// It returns nil if the component is not found.
+func StopDeregisterComponent(name string) error {
```
eahydra, line 87: Why do we need this function to support unregister and stop operation? It's a bit weird.. We don't support dynamic component..

components/type.go
```diff
@@ -0,0 +1,90 @@
...
+
+type State struct {
+	Name      string            `json:"name,omitempty"`
+	Healthy   bool              `json:"healthy,omitempty"`
+	Health    string            `json:"health,omitempty"`     // Healthy, Degraded, Unhealthy
+	Reason    string            `json:"reason,omitempty"`     // a detailed and processed reason on why the component is not healthy
+	Error     string            `json:"error,omitempty"`      // the unprocessed error returned from the component
+	ExtraInfo map[string]string `json:"extra_info,omitempty"` // any extra information the component may want to expose
+
+	SuggestedActions *common.SuggestedActions `json:"suggested_actions,omitempty"`
```
eahydra, line 58: We should move the SuggestedActions type into the file in the future.

eahydra:
@gyuho  Shall we revisit the component management overall?  Who uses the GetComponent function?  Only the sesion manager?  Shall we consider improve it by referring to #511?

components/components.go
gyuho: Mainly to make these unit testable (vs. global map)

components/components.go
gyuho: Was thinking, the user may want to deregister the external plugin. Then this will come in handy? What do you think

components/components.go
gyuho: mainly, to make these unit testable, what do you think

components/components.go
gyuho: +1 for not creating new components for each plugin. Let me experiment more, so that we have 1 plugin component that can handle multiple runs

components/components.go
gyuho: removed

components/components.go
gyuho: addressed

components/type.go
gyuho: done

api/v1/types.go
```diff
@@ -1,36 +1,198 @@
...
+// Component represents an individual component of the system.
+//
+// Each component check is independent of each other.
+// But the underlying implementation may share the same data sources
+// in order to minimize the querying overhead (e.g., nvidia-smi calls).
+//
+// Each component implements its own output format inside the State struct.
+// And recommended to have a consistent name for its HTTP handler.
+// And recommended to define const keys for the State extra information field.
+type Component interface {
```
eahydra, line 20: it is reasonable that other than `Component` and `HealthSettable` are moved to this file..`Component` and `HealthSettable` should be moved to original file `components/components.go` or `components/types.go`

eahydra (approved):

PR #631, opened 2025-04-11
test(containerd/pod): add more unit tests

gyuho:

### DIFF
--- a/components/containerd/pod/cri.go
+++ b/components/containerd/pod/cri.go
@@ -9,13 +9,14 @@ import (
 	"sort"
 	"time"
 
-	pkg_file "github.com/leptonai/gpud/pkg/file"
-	"github.com/leptonai/gpud/pkg/log"
 	"google.golang.org/grpc"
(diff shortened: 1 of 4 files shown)


components/containerd/pod/cri.go
```diff
@@ -200,7 +201,12 @@ func checkContainerdRunning(ctx context.Context) bool {
 	return false
 }
 
-func listSandboxStatus(ctx context.Context, endpoint string) ([]PodSandbox, error) {
+func listAllSandboxes(
```
eahydra, line 204: Please find another way to do test..As I commented on others, if we want to mock some functions, we would define a function variable which can be replaced in tests and after test, we would reset the function variable to the real functuon.. Introduce these two new params  which lead the complex to the caller.. it is unreasonable..

gyuho: @eahydra Fixed.

eahydra (approved):

PR #633, opened 2025-04-11
feat(api/v1): add "LoginRequest" struct

gyuho:


api/v1/login.go
```diff
@@ -0,0 +1,55 @@
...
+// LoginRequest is the request for the login request.
+type LoginRequest struct {
+	Token       string             `json:"token"`
+	Network     MachineNetworkSpec `json:"network"`
+	Location    MachineLocation    `json:"location"`
+	CPUInfo     MachineCPUInfo     `json:"cpuInfo"`
+	GPUInfo     MachineGPUInfo     `json:"gpuInfo"`
+	Provider    string             `json:"provider"`
+	MachineInfo MachineInfo        `json:"machineInfo"`
+	MachineType string             `json:"machineType"`
```
eahydra, line 12: What's MachineType here? IMO gpud should not care about it.

api/v1/login.go
```diff
@@ -0,0 +1,55 @@
...
+	CUDAVersion             string `json:"cudaVersion,omitempty"`
+	// MachineID is the ID of the machine collected from GPUd
+	MachineID string `json:"machineID,omitempty"`
+}
+
+// LoginResponse is the response for the login request.
+type LoginResponse struct {
+	Status    string `json:"status"`
+	MachineID string `json:"machine_id,omitempty"`
+	Error     string `json:"error,omitempty"`
```
eahydra, line 54: What is the difference between status and error?

api/v1/login.go
```diff
@@ -0,0 +1,55 @@
+package v1
+
+// LoginRequest is the request for the login request.
+type LoginRequest struct {
```
eahydra, line 4: Shall we add the capacity information into the request, including how many cpus, memory in bytes, disks metadata, disks spaces in bytes, root filesystem in bytes and gpu count?

api/v1/login.go
gyuho: Documented

api/v1/login.go
gyuho: @cardyok ^ (removed in this PR)

api/v1/login.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// LoginResponse is the response for the login request.
+type LoginResponse struct {
+	// Status is the status of the login request.
+	Status string `json:"status"`
+
+	// MachineID is the ID of the machine generated by the login control plane.
+	// Only set upon successful login.
+	MachineID string `json:"machine_id,omitempty"`
```
eahydra, line 56: Let's keep the same style.. 
```suggestion
	MachineID string `json:"machineID,omitempty"`
```

gyuho: @cardyok ^

api/v1/login.go
eahydra: gpud does not know which type it is..

eahydra (approved):

PR #635, opened 2025-04-12
feat(process/scheduler): add exclusive scheduler, support "bootstrap" session message

gyuho:
- **feat(process/scheduler): add exclusive scheduler**
- **feat(pkg/session): add "bootstrap" endpoint**

### DIFF
--- a/pkg/session/serve.go
+++ b/pkg/session/serve.go
@@ -2,6 +2,7 @@ package session
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -35,6 +36,7 @@ type Request struct {
 	Since         time.Duration     `json:"since"`
 	UpdateVersion string            `json:"update_version,omitempty"`
 	UpdateConfig  map[string]string `json:"update_config,omitempty"`
+	Bootstrap     *BootstrapRequest `json:"bootstrap,omitempty"`
 }
 
 type Response struct {
@@ -44,6 +46,22 @@ type Response struct {
 	States  apiv1.GPUdComponentHealthStates `json:"states,omitempty"`
 	Events  apiv1.GPUdComponentEvents       `json:"events,omitempty"`
 	Metrics apiv1.GPUdComponentMetrics      `json:"metrics,omitempty"`
+
+	Bootstrap *BootstrapResponse `json:"bootstrap,omitempty"`
+}
+
+type BootstrapRequest struct {
+	// TimeoutInSeconds is the timeout for the bootstrap script.
+	// If not set, the default timeout is 10 seconds.
+	TimeoutInSeconds int `json:"timeout_in_seconds,omitempty"`
+
+	// ScriptBase64 is the base64 encoded script to run.
+	ScriptBase64 string `json:"script_base64,omitempty"`
+}
+
+type BootstrapResponse struct {
+	Output   string `json:"output,omitempty"`
+	ExitCode int32  `json:"exit_code,omitempty"`
 }
 
 func (s *Session) serve() {
@@ -171,6 +189,31 @@ func (s *Session) serve() {
 					default:
 						log.Logger.Warnw("unsupported component for updateConfig", "component", componentName)
 					}
+				}
+			}
+
+		case "bootstrap":
+			if payload.Bootstrap != nil {
+				script, err := base64.StdEncoding.DecodeString(payload.Bootstrap.ScriptBase64)
+				if err != nil {
+					response.Error = err.Error()
+					break
+				}
+
+				timeout := time.Duration(payload.Bootstrap.TimeoutInSeconds) * time.Second
+				if timeout == 0 {
+					timeout = 10 * time.Second
+				}
+
+				cctx, cancel := context.WithTimeout(ctx, timeout)
+				output, exitCode, err := s.processRunner.RunUntilCompletion(cctx, string(script))
+				cancel()
+				response.Bootstrap = &BootstrapResponse{
+					Output:   string(output),
+					ExitCode: exitCode,
+				}
+				if err != nil {
+					response.Error = err.Error()
 				}
 			}
 		}
--- a/pkg/process/runner.go
+++ b/pkg/process/runner.go
@@ -1,0 +1,19 @@
+package process
+
+import (
+	"context"
+	"errors"
+)
+
+var (
+	ErrProcessAlreadyRunning = errors.New("process already running")
+)
+
+// Runner defines the interface for a process runner.
+// It facillitates scheduling and running arbitrary bash scripts.
+type Runner interface {
+	// RunUntilCompletion starts a bash script, blocks until it finishes,
+	// and returns the output and the exit code.
+	// Whether to return an error when there is already a process running is up to the implementation.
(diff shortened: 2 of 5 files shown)


pkg/process/scheduler/exclusive.go
```diff
@@ -0,0 +1,107 @@
...
+
+// RunSyncMode starts a bash script, blocks until it finishes,
+// and returns the output and the exit code.
+// If there is already a process running, the previous process will be aborted,
+// and the new process will be started.
+func (s *exclusiveScheduler) RunSyncMode(ctx context.Context, script string) ([]byte, int32, error) {
+	s.stopRunningProcess(ctx)
+
+	// write all stderr + stdout to a temporary file
+	tmpFile, err := os.CreateTemp(os.TempDir(), "process-output-*.txt")
```
eahydra, line 43:
```suggestion
	tmpFile, err := os.CreateTemp(os.TempDir(), "lepton-process-output-*.txt")
```

pkg/process/scheduler/exclusive.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return 1, nil
+}
+
+// RunSyncMode starts a bash script, blocks until it finishes,
+// and returns the output and the exit code.
+// If there is already a process running, the previous process will be aborted,
+// and the new process will be started.
+func (s *exclusiveScheduler) RunSyncMode(ctx context.Context, script string) ([]byte, int32, error) {
+	s.stopRunningProcess(ctx)
```
eahydra, line 40: If we stop previsou porcess, it would not be a exclusive process. Shall we return an error if there is a process running?

pkg/process/scheduler/exclusive.go
```diff
@@ -0,0 +1,107 @@
...
+
+var _ Scheduler = &exclusiveScheduler{}
+
+func NewExclusiveScheduler() Scheduler {
+	return &exclusiveScheduler{}
+}
+
+// exclusiveScheduler is a scheduler that runs a single process at a time.
+// Does not support concurrent execution of multiple processes.
+type exclusiveScheduler struct {
```
eahydra, line 20: Scheduler is big concept, we should simplify the concept as much as possible. Here it's better to name it ExclusiveProcess, and if it's running, we can return an error. In other words, this control logic is the responsibility of the caller.

pkg/process/scheduler/exclusive.go
```diff
@@ -0,0 +1,107 @@
...
+	mu             sync.RWMutex
+	runningProcess process.Process
+}
+
+func (s *exclusiveScheduler) CountProcesses() (int, error) {
```
eahydra, line 25: We defined it but no one used it.. We should consider if it is necessary.. If not, we should remove it.

pkg/session/serve.go
```diff
@@ -44,6 +46,18 @@ type Response struct {
 	States  apiv1.LeptonStates  `json:"states,omitempty"`
 	Events  apiv1.LeptonEvents  `json:"events,omitempty"`
 	Metrics apiv1.LeptonMetrics `json:"metrics,omitempty"`
+
+	Bootstrap *BootstrapResponse `json:"bootstrap,omitempty"`
+}
+
+type BootstrapRequest struct {
+	// ScriptBase64 is the base64 encoded script to run.
+	ScriptBase64 string `json:"script_base64,omitempty"`
```
eahydra, line 55: Can we support timeout param?

pkg/process/scheduler/exclusive.go
gyuho: @cardyok ^ Our control plane needs to follow this semantics

pkg/process/runner/exclusive.go
```diff
@@ -0,0 +1,88 @@
...
+// RunUntilCompletion starts a bash script, blocks until it finishes,
+// and returns the output and the exit code.
+// If there is already a process running, it returns an error.
+func (er *exclusiveRunner) RunUntilCompletion(ctx context.Context, script string) ([]byte, int32, error) {
+	if er.alreadyRunning() {
+		return nil, 0, ErrProcessAlreadyRunning
+	}
+
+	// write all stderr + stdout to a temporary file
+	tmpFile, err := os.CreateTemp(os.TempDir(), "gpud-process-output-*.txt")
```
eahydra, line 34: I forgot that if there is an issue whether write the output to a bytes buffer..

eahydra (commented):
overall lgtm

pkg/process/runner/exclusive.go
gyuho: You mean, piping the output to memory and return? I think we can... but like you said, we may also hit the buffer limit, whereas writing to a file doesn't have that kind of issue

gyuho: When we want script output stream, maybe we can read line by line and send them back, but if we assume, we always run until completion, I think reading from the file is simpler. let me know if you have other preferences.

eahydra: OK. Let's output to a file. By the way, we should consider that if the GPU crashes, there may be no chance to delete the file and we should find a way to clean it.

gyuho: @eahydra done.

pkg/session/serve.go
```diff
@@ -173,6 +191,30 @@ func (s *Session) serve() {
 					}
 				}
 			}
+
+		case "bootstrap":
+			if payload.Bootstrap != nil {
+				script, err := base64.StdEncoding.DecodeString(payload.Bootstrap.ScriptBase64)
+				if err != nil {
+					response.Error = err.Error()
```
eahydra, line 199: Should return if we got an error.

pkg/process/runner/exclusive.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil, 0, fmt.Errorf("failed to create temp dir %s: %w", defaultScriptsDir, err)
+	}
+	tmpFile, err := os.CreateTemp(defaultScriptsDir, "gpud-process-output-*.txt")
+	if err != nil {
+		return nil, 0, err
+	}
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
+
+	proc, err := process.New(
```
eahydra, line 48: Don't use proc as a variable name. proc looks more like /sys/proc.

pkg/process/runner/exclusive.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// exclusiveRunner is a scheduler that runs a single process at a time.
+// Does not support concurrent execution of multiple processes.
+type exclusiveRunner struct {
+	mu      sync.RWMutex
+	running process.Process
+}
+
+var defaultScriptsDir = filepath.Join(os.TempDir(), "gpud-scripts-runner")
```
eahydra, line 27: shall we move it to the process pckage? IMO all these scripts and related files should be stored there. and I didn't see how to clean it.

gyuho: Sure, let me move

gyuho: @eahydra done

pkg/session/serve.go
gyuho: Done

PR #639, opened 2025-04-13
feat(components): rename state to health state

gyuho:
Makes more sense since each state represents the health states of a component.

### DIFF
--- a/components/accelerator/nvidia/ecc/component.go
+++ b/components/accelerator/nvidia/ecc/component.go
@@ -67,14 +67,14 @@ func (c *component) Start() error {
 	return nil
 }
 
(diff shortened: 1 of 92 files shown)


components/accelerator/nvidia/ecc/component.go
```diff
@@ -67,14 +67,14 @@ func (c *component) Start() error {
 	return nil
 }
 
-func (c *component) States(ctx context.Context) ([]apiv1.State, error) {
+func (c *component) HealthStates(ctx context.Context) (apiv1.HealthStates, error) {
```
eahydra, line 70: Some components return apiv1.HealthStates, others return []apiv1.HealthState. Should we keep the same style?

eahydra (approved):

PR #640, opened 2025-04-13
feat(os, nvidia, gpud scan): add "HealthCheckResult" interface to simplify scan per component

gyuho:
To get rid of a shared GPU fetch function in https://github.com/leptonai/gpud/blob/a37fbc23df80f827d4ef9d83206b6f580969182c/pkg/nvidia-query/query.go#L18-L24, and to support gossip API.

### DIFF
--- a/components/accelerator/nvidia/temperature/component.go
+++ b/components/accelerator/nvidia/temperature/component.go
@@ -2,6 +2,7 @@
 package temperature
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
+	"github.com/olekukonko/tablewriter"
 	"github.com/prometheus/client_golang/prometheus"
 
 	apiv1 "github.com/leptonai/gpud/api/v1"
@@ -88,25 +90,29 @@ func (c *component) Close() error {
 // run this periodically
 func (c *component) CheckOnce() {
 	log.Logger.Infow("checking temperature")
-	d := Data{
+
+	d := checkHealthState(c.nvmlInstance, c.getTemperatureFunc)
+
+	c.lastMu.Lock()
+	c.lastData = d
+	c.lastMu.Unlock()
+}
+
+func checkHealthState(nvmlInstance nvidianvml.InstanceV2, getTemperatureFunc func(uuid string, dev device.Device) (nvidianvml.Temperature, error)) *Data {
+	d := &Data{
 		ts: time.Now().UTC(),
 	}
-	defer func() {
-		c.lastMu.Lock()
-		c.lastData = &d
-		c.lastMu.Unlock()
-	}()
 
 	tempThresholdExceeded := make([]string, 0)
-	devs := c.nvmlInstance.Devices()
+	devs := nvmlInstance.Devices()
 	for uuid, dev := range devs {
-		temp, err := c.getTemperatureFunc(uuid, dev)
+		temp, err := getTemperatureFunc(uuid, dev)
 		if err != nil {
 			log.Logger.Errorw("error getting temperature for device", "uuid", uuid, "error", err)
 			d.err = err
 			d.healthy = false
 			d.reason = fmt.Sprintf("error getting temperature for device %s", uuid)
-			return
+			return d
 		}
 		d.Temperatures = append(d.Temperatures, temp)
 
@@ -131,7 +137,7 @@ func (c *component) CheckOnce() {
 			d.err = err
 			d.healthy = false
 			d.reason = fmt.Sprintf("error getting used percent for slowdown for device %s", uuid)
-			return
+			return d
 		}
 		metricSlowdownUsedPercent.With(prometheus.Labels{pkgmetrics.MetricLabelKey: uuid}).Set(slowdownPct)
 	}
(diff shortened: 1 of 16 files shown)


api/v1/types.go
```diff
@@ -16,6 +16,23 @@ const (
 	StateTypeInitializing StateType = "Initializing"
 )
 
+const (
+	checkMark   = "\033[32m✔\033[0m"
+	warningSign = "\033[31m✘\033[0m"
+)
+
+// TerminalHeader returns a terminal header based on its health state.
+func (st StateType) TerminalHeader() string {
```
eahydra, line 25: It's not recommended that define the method for StateType here.  We should define it close it caller and IMO there is only one package(scan) relied on it.

components/accelerator/nvidia/temperature/component.go
```diff
@@ -88,25 +90,29 @@ func (c *component) Close() error {
...
-	d := Data{
+
+	d := collectData(c.nvmlInstance, c.getTemperatureFunc)
+
+	c.lastMu.Lock()
+	c.lastData = d
+	c.lastMu.Unlock()
+}
+
+func collectData(nvmlInstance nvidianvml.InstanceV2, getTemperatureFunc func(uuid string, dev device.Device) (nvidianvml.Temperature, error)) *Data {
```
eahydra, line 101: Probably we don't need to pass the getTemperatureFunc here.. If we would like to mock it, we would mock the InstanceV2 interface, just override the GetTemeprature method and if there is no corresponding method, we would add it to support mock tests.

eahydra: Anyway, we should focus on improving the InstanceV2 which should support all NVIDIA GPU related oprations needed by us. So that we can inject failure for E2E tests...

components/accelerator/nvidia/temperature/component.go
```diff
@@ -143,8 +149,20 @@ func (c *component) CheckOnce() {
 		d.healthy = false
 		d.reason = fmt.Sprintf("exceeded HBM temperature thresholds: %s", strings.Join(tempThresholdExceeded, ", "))
 	}
+
+	return d
 }
 
+func CollectData(nvmlInstance nvidianvml.InstanceV2) (components.Data, error) {
+	d := collectData(nvmlInstance, nvidianvml.GetTemperature)
```
eahydra, line 157: This is what confuses me.. We pass a nvmlInstance, and then pass its method into the function.

components/type.go
```diff
@@ -43,3 +44,19 @@ type HealthSettable interface {
 	// SetHealthy sets the health state to healthy.
 	SetHealthy() error
 }
+
+// Data is the data type that can be checked by the component.
+type Data interface {
```
eahydra, line 49: A question: what's the purpose of Summary method? Is it used to generate message for events?

components/type.go
```diff
@@ -43,3 +44,19 @@ type HealthSettable interface {
...
+	// String returns a string representation of the data.
+	// Describes the data in a human-readable format.
+	fmt.Stringer
+
+	// Summary returns a summary of the data.
+	Summary() string
+
+	// JSON encodes the data into JSON format.
+	JSON() ([]byte, error)
```
eahydra, line 58: Is it better to use `json.Marshaler` instead of the JSON method? So that we can:
```go
result, err := componentA.CheckOnce()
if err != nil {
    // handle with error 
    return
}
data, err := json.Marshal(result)
if err != nil {
    return
}
```
and as you can see, the `Data` is not a good name.  It doesn't matter if it's defined in component internal, but if we define it as a public interface, shall we give it a resonable name?

pkg/diagnose/scan.go
```diff
@@ -44,43 +45,13 @@ func Scan(ctx context.Context, opts ...OpOption) error {
...
-		log.Logger.Warnw("error counting processes by status", "error", err)
-	} else {
-		fmt.Printf("%s counted processes by status\n", checkMark)
-		for status, count := range processCountsByStatus {
-			fmt.Printf("%s %q %d\n", checkMark, status, len(count))
-		}
+		return err
 	}
+	fmt.Printf("%s %s\n", osData.HealthState().TerminalHeader(), osData.Summary())
+	fmt.Println(osData.String())
```
eahydra, line 54: Shall we print the information without osData? The information is static almostly except that process counts.  And after gossip refactoring,  the os package only update some metrics.

pkg/nvidia-query/nvml/clock_events.go
```diff
@@ -13,19 +13,20 @@ import (
...
 )
 
 // Returns true if clock events is supported by all devices.
 // Returns false if any device does not support clock events.
 // ref. undefined symbol: nvmlDeviceGetCurrentClocksEventReasons for older nvidia drivers
 func ClockEventsSupported() (bool, error) {
-	nvmlLib := nvml_lib.NewDefault()
-	if ret := nvmlLib.NVML().Init(); ret != nvml.SUCCESS {
-		return false, fmt.Errorf("failed to initialize NVML: %v", nvml.ErrorString(ret))
+	nvmlLib, err := nvml_lib.New()
```
eahydra, line 23: As we talked before, we would like to initialize the NVML once.. For the case, shall we make the function as a method of InstanceV2? or provide a free function required passing the InstanceV2 into it?

components/type.go
gyuho: Each check in `gpud scan` output now has two values checkmark/warning sign + one-line summary. The `Summary` method can be used to print out that one-liner (e.g., `x this check has failed`).

components/accelerator/nvidia/temperature/component.go
gyuho: @eahydra  Yeah, only reason we do this, is not to regress on the unit test coverage. We will certainly clean up the NVML interface

components/accelerator/nvidia/temperature/component.go
gyuho: We need common function that can mock the get temp function, for the unit test coverage. In the future, I will clean up the nvml instance interface, so we don't need to pass this along.

api/v1/types.go
gyuho: removed

pkg/nvidia-query/nvml/clock_events.go
gyuho: Yeah I am also not a fan of current interface. Let us clean up in the following PR https://github.com/leptonai/gpud/issues/511

pkg/diagnose/scan.go
gyuho: The intention is, summary + Health state methods represent much condensed information about health state. And `String` just emits all data (not all, but important information), as it's used for "scanning"

eahydra (approved):
lgtm, let's merge it first, and we should improve these related code in the future.

PR #641, opened 2025-04-14
feat(gossip, login): separate gossip pkg, implement new login/gossip operations

gyuho:
@eahydra @cardyok Now implementing very basic login + gossip.

Needs a lot of cleanups later.


cmd/gpud/command/login.go
```diff
@@ -82,7 +82,7 @@ func cmdLogin(cliContext *cli.Context) error {
 		if err != nil {
 			hostname = "UnknownName"
 		}
-		if err := login.Login(hostname, token, endpoint, components, uid); err != nil {
+		if err := login.SendRequestToOldAPI(hostname, token, endpoint, components, uid); err != nil {
```
eahydra, line 85: We don't need it anymore

gyuho: @eahydra Removed.

And will implement the rest of the login workflow in the following PR

PR #643, opened 2025-04-14
feat(session): add cookiejar and do health check before starting session

cardyok:


gyuho:
lgtm, after fixing tests

pkg/server/server.go
```diff
@@ -648,7 +648,7 @@ func (s *Server) updateToken(ctx context.Context, db *sql.DB, uid string, endpoi
 		var err error
 		s.session, err = session.NewSession(
 			ctx,
-			fmt.Sprintf("https://%s/api/v1/session", endpoint),
+			fmt.Sprintf("https://%s", endpoint),
```
eahydra, line 651: Shall we make the https scheme as the part of endpoint?

pkg/session/session.go
```diff
@@ -172,8 +174,14 @@ func (s *Session) keepAlive() {
 			writerExit := make(chan any)
 			s.closer = &closeOnce{closer: make(chan any)}
 			ctx, cancel := context.WithCancel(context.Background()) // create local context for each session
-			go s.startReader(ctx, readerExit)
-			go s.startWriter(ctx, writerExit)
+			jar, _ := cookiejar.New(nil)
+			if err := s.checkServerHealth(ctx, jar); err != nil {
```
eahydra, line 178: I don't understand why we need the cookijar package..

eahydra (approved):

pkg/server/server.go
cardyok: updated

PR #644, opened 2025-04-14
feat(custom-plugins): support custom health state in custom plugin

gyuho:
Address https://github.com/leptonai/gpud/issues/568.

### DIFF
--- a/pkg/custom-plugins/output_state.go
+++ b/pkg/custom-plugins/output_state.go
@@ -1,0 +1,54 @@
+package customplugins
+
+import (
+	"fmt"
+	"strings"
+
+	apiv1 "github.com/leptonai/gpud/api/v1"
+)
+
+// HealthStateOutputPrefixType defines the prefix for the line,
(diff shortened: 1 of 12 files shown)


pkg/custom-plugins/output_state.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"strings"
+
+	apiv1 "github.com/leptonai/gpud/api/v1"
+)
+
+// StateOutputPrefix defines the prefix for the line,
+// where each plugin is expected to emit its state data in JSON format.
+// The JSON data schema for the state is defined in api/v1/types.go.
+const StateOutputPrefix = "GPUD_CUSTOM_PLUGIN_HEALTH_STATE:"
```
eahydra, line 15: we should make it shorter

eahydra (approved):

pkg/custom-plugins/output_state.go
gyuho: will update, and also add more e2e tests

PR #651, opened 2025-04-15
feat(components/*): refactor to add "Check" method for one-time operations, simplify Component interfaces

gyuho:
Will come in handy to dedup code base for `gpud scan` and gossip messages:

- **feat(components): add registry All method**
- **feat(components): rename HealthStateCheckResult, add "Check" to interface**
- **feat(os): refactor to use "Check" method**

Address https://github.com/leptonai/gpud/issues/511.

Also, address https://github.com/leptonai/gpud/issues/472.


pkg/scan/scan.go
```diff
@@ -50,48 +49,57 @@ func printSummary(result components.HealthStateCheckResult) {
...
+		oscomponent.New,
+	}
+
+	for _, initFunc := range componentInits {
+		c, err := initFunc(gpudInstance)
 		if err != nil {
 			return err
 		}
-		printSummary(nvidiaTempData)
+		defer c.Close()
```
eahydra, line 96: This is not a good practice. The for loop is short and we can close it after the check.

pkg/server/server.go
```diff
@@ -273,7 +273,14 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID
...
 
 		case os.Name:
-			allComponents = append(allComponents, os.New(ctx, rebootEventStore))
+			c, err := os.New(components.GPUdInstance{
+				RootCtx:          ctx,
+				RebootEventStore: rebootEventStore,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to create component %s: %w", k, err)
+			}
```
eahydra, line 282: we can init these components in a loop, so that we can remove all these switch.case eventually.

components/registry.go
```diff
@@ -14,17 +16,25 @@ import (
...
-	NVMLInstance     nvidianvml.InstanceV2
+	NVMLInstance         nvidianvml.InstanceV2
+	NVIDIAToolOverwrites nvidiacommon.ToolOverwrites
+
 	EventStore       eventstore.Store
 	RebootEventStore pkghost.RebootEventStore
 }
 
+// InitFunc is the function that initializes a component.
+type InitFunc func(GPUdInstance) (Component, error)
```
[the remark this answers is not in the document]
gyuho: let me merge draft PRs into this branch and fix this

pkg/server/server.go
gyuho: Yes, that's the plan. I will do this in this PR

components/os/component.go
```diff
@@ -162,17 +146,27 @@ func checkHealthState(ctx context.Context, countProcessesByStatusFunc func(ctx c
...
 	d.reason = fmt.Sprintf("os kernel version %s", d.Kernel.Version)
+
+	// iff reboot event store is available
+	// e.g., "gpud scan" is an one-time operation
+	// in such case, we may not set the db instance
+	if c.rebootEventStore != nil {
+		if err := c.rebootEventStore.RecordReboot(c.ctx); err != nil {
+			log.Logger.Warnw("error creating reboot event", "error", err)
+		}
+	}
```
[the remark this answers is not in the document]
gyuho: Now done in a separate goroutine.

pkg/server/server.go
gyuho: Fixed all.

pkg/server/server.go
```diff
@@ -203,186 +263,51 @@ func New(ctx context.Context, config *lepconfig.Config, endpoint string, cliUID
...
-			}
-			allComponents = append(allComponents, c)
+		MountPoints:  []string{"/"},
+		MountTargets: []string{"/var/lib/kubelet"},
+	}
 
-		case containerd_pod.Name:
-			allComponents = append(allComponents, containerd_pod.New(ctx))
+	allComponents := make([]components.Component, 0)
+	for _, initFunc := range componentInits {
```
eahydra, line 283: Shall we merge these three for loop together? L291, L297, L308

gyuho: done

components/accelerator/nvidia/nccl/component.go
```diff
@@ -4,59 +4,92 @@ package nccl
...
-func (c *component) Start() error { return nil }
+func (c *component) Start() error {
+	// do not need periodic kmsg checks since it already has a watcher
+	return nil
+}
 
-func (c *component) HealthStates(ctx context.Context) (apiv1.HealthStates, error) {
-	return []apiv1.HealthState{
+func (c *component) LastHealthStates() apiv1.HealthStates {
+	return apiv1.HealthStates{
```
eahydra, line 80: Do we need to return an empty state?  And there is no Name in the state, did we miss it?

eahydra (approved):

components/accelerator/nvidia/nccl/component.go
gyuho: oh good catch, fixing

pkg/scan/scan.go
```diff
@@ -2,238 +2,154 @@ package scan
...
 
-	for _, checkFunc := range defaultCheckHealthStateFuncs {
-		rs, err := checkFunc(ctx)
-		if err != nil {
-			return err
-		}
-		printSummary(rs)
-	}
-
 	nvidiaInstalled, err := nvidia_query.GPUsInstalled(ctx)
```
eahydra, line 118: rename to `nvidiaquery. GPUsInstalled(ctx)` ?

PR #663, opened 2025-04-16
COPY, DO NOT MERGE: feat(components, os): add one-time "Check" method for each component

gyuho:
Just for reference to https://github.com/leptonai/gpud/pull/651.


gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/651

PR #670, opened 2025-04-19
feat(reboot): reboot after 10s by default

eahydra:
To inform the control plane that the reboot request has been processed, reboot after 10 seconds.


gyuho (approved):
lgtm

PR #672, opened 2025-04-20
feat(components): add more register methods, add Deregisterable interface

gyuho:
Cherry-picked from https://github.com/leptonai/gpud/pull/627.

### DIFF
--- a/components/types.go
+++ b/components/types.go
@@ -47,6 +47,14 @@ type Component interface {
 	Close() error
 }
 
+// Deregisterable is an interface that allows a custom plugin to be deregistered.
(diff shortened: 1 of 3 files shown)


components/types.go
```diff
@@ -47,6 +47,14 @@ type Component interface {
 	Close() error
 }
 
+// Deregisterable is an interface that allows a custom plugin to be deregistered.
+// By default, the regular/built-in components are not allowed to be deregistered,
+// unless it implements this interface.
+type Deregisterable interface {
```
eahydra, line 53: IMO every component should be deregisterable.

eahydra (approved):

PR #706, opened 2025-04-25
feat(custom-plugins): jsonpath parser + regex to extract health info from output

gyuho:


pkg/custom-plugins/component.go
```diff
@@ -124,13 +124,35 @@ func (c *component) Check() components.CheckResult {
...
+			cr.reason = fmt.Sprintf("error extracting extra info -- %s (output: %s)", cr.err, string(cr.out))
+		}
+
+		if len(matchResults) > 0 {
+			for k, result := range matchResults {
+				cr.extraInfo[k] = result.fieldValue
+
+				if !result.matched {
+					cr.health = apiv1.HealthStateTypeUnhealthy
+					cr.reason = fmt.Sprintf("extra info %q has unexpected value %q (based on match rule %q)", k, result.fieldValue, result.rule)
```
eahydra: shall we hide the details? Just give the result: we cannot find target value by the rule(name or id, not the rule content)

pkg/custom-plugins/plugin_output.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if po.JSONPaths != nil {
+		return extractExtraInfoWithJSONPaths(input, po.JSONPaths)
+	}
+
+	return nil, nil
+}
+
+type extractedExtraInfo struct {
```
eahydra, line 48:
```suggestion
type extractedField struct {
```

pkg/custom-plugins/types.go
```diff
@@ -83,3 +87,35 @@ type RunBashScript struct {
...
+	// ref. https://goessner.net/articles/JsonPath/
+	QueryPath string `json:"query_path"`
+	// FieldName defines the field name to represent this query result.
+	FieldName string `json:"field_name"`
+
+	// MatchRule is the match rule for the field value.
+	// It not set, the field value is not checked.
+	// If set, the field value is checked against the match rule.
+	// If set but mismatched, the health state is set to "Unhealthy".
+	MatchRule
```
eahydra, line 113: Give it a field name and a json tag.

pkg/custom-plugins/plugin_output.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, err
+	}
+	return v, nil
+}
+
+// checkMatchRule checks if the input matches the match rule.
+// It returns true if the input matches the match rule, otherwise false.
+// It returns an error if the match rule is invalid.
+// The second return value is the match rule that was used to check the input.
+func (rule MatchRule) checkMatchRule(input string) (bool, string, error) {
```
eahydra, line 177: Please put the method close to where MatchRule is defined at. That can improve the readability.  BTW  for the receiver type, we should use pointer by default.

pkg/custom-plugins/types.go
```diff
@@ -58,6 +63,10 @@ type Plugin struct {
 	// If a step fails, the execution stops and the error is returned.
 	// Which means, the final success requires all steps to succeed.
 	Steps []Step `json:"steps,omitempty"`
+
+	// OutputParse is the parser for the plugin output.
+	// If not set, the default prefix parser is used.
+	OutputParse *PluginOutputParseConfig `json:"output_parse,omitempty"`
```
eahydra, line 69: Shall we give it a short name? Parser? Extractor?  Trait？

pkg/custom-plugins/types.go
```diff
@@ -83,3 +92,58 @@ type RunBashScript struct {
...
+	JSONPaths []JSONPath `json:"json_paths,omitempty"`
+}
+
+// JSONPath represents a JSON path to the output fields.
+type JSONPath struct {
+	// QueryPath defines the JSONPath query path to extract for.
+	// ref. https://pkg.go.dev/github.com/PaesslerAG/jsonpath#section-readme
+	// ref. https://en.wikipedia.org/wiki/JSONPath
+	// ref. https://goessner.net/articles/JsonPath/
+	QueryPath string `json:"query_path"`
```
eahydra, line 110:
```suggestion
	QueryPath string `json:"query"`
```

pkg/custom-plugins/types.go
```diff
@@ -83,3 +92,58 @@ type RunBashScript struct {
...
+	// FieldName defines the field name to represent this query result.
+	FieldName string `json:"field_name"`
```
eahydra, line 112:
```suggestion
	Field string `json:"field"`
```

pkg/custom-plugins/types.go
```diff
@@ -83,3 +92,58 @@ type RunBashScript struct {
...
+
+	// MatchRule is the match rule for the field value.
+	// It not set, the field value is not checked.
+	// If set, the field value is checked against the match rule.
+	// If set but mismatched, the health state is set to "Unhealthy".
+	MatchRule *MatchRule `json:"match_rule,omitempty"`
```
eahydra, line 118: Filter?

pkg/custom-plugins/component.go
```diff
@@ -124,13 +124,35 @@ func (c *component) Check() components.CheckResult {
 	cr.reason = "ok"
 	log.Logger.Debugw("successfully executed plugin", "exitCode", cr.exitCode, "output", string(cr.out))
 
+	// either custom parser (jsonpath) or default parser
+	if len(cr.out) > 0 && c.spec.HealthStatePlugin.Parser != nil {
+		var matchResults map[string]extractedField
+		matchResults, cr.err = c.spec.HealthStatePlugin.Parser.extractExtraInfo(cr.out)
+		if cr.err != nil {
+			log.Logger.Errorw("error extracting extra info", "error", cr.err)
+			cr.reason = fmt.Sprintf("error extracting extra info -- %s (output: %s)", cr.err, string(cr.out))
```
eahydra, line 133: Return the out to user? We should hide it...

pkg/custom-plugins/component.go
```diff
@@ -124,13 +124,35 @@ func (c *component) Check() components.CheckResult {
...
+		}
+
+		if len(matchResults) > 0 {
+			for k, result := range matchResults {
+				cr.extraInfo[k] = result.fieldValue
+
+				if !result.matched {
+					cr.health = apiv1.HealthStateTypeUnhealthy
+					cr.reason = fmt.Sprintf("cannot find the matching value for %q", k)
```
eahydra, line 144: we also need to hide this. we can add log for the case, but does not show user it.

pkg/custom-plugins/component.go
gyuho: fixed

eahydra (approved):

PR #714, opened 2025-04-27
fix(v1/metrics API): store all labels, remove "metric_name" and "metric_secondary_name"

gyuho:
e.g.,

/v1/metrics response:

> [{"component":"accelerator-nvidia-processes","metrics":[{"unix_seconds":1746523049574,"name":"accelerator_nvidia_processes_running_total","labels":{"uuid":"mock"},"value":0}]},{"component":"accelerator-nvidia-temperature","metrics":[{"unix_seconds":1746523049574,"name":"accelerator_nvidia_temperature_current_celsius","labels":{"uuid":"mock"},"value":1},


pkg/metrics/types.go
```diff
@@ -9,8 +9,9 @@ const (
 	// MetricComponentLabelKey is the key for the component of the metric.
 	MetricComponentLabelKey = "gpud_component"
 
-	// MetricLabelKey is the key for the label of the metric.
-	MetricLabelKey = "gpud_metric_label"
+	// MetricLabelNamePrefix is the key prefix for the label of the metric.
+	// The label key must be prefixed with this key.
+	MetricLabelNamePrefix = "label_"
```
eahydra, line 14: shall we remove the prefix?

pkg/metrics/scraper/prometheus.go
```diff
@@ -44,11 +45,14 @@ func (s *promScraper) Scrape(_ context.Context) (pkgmetrics.Metrics, error) {
...
+				if labelName == pkgmetrics.MetricComponentLabelKey {
 					m.Component = label.GetValue()
-				case pkgmetrics.MetricLabelKey:
-					m.Label = label.GetValue()
+					continue
+				}
+				if strings.HasPrefix(labelName, pkgmetrics.MetricLabelNamePrefix) {
+					m.LabelName = strings.TrimPrefix(labelName, pkgmetrics.MetricLabelNamePrefix)
+					m.LabelValue = label.GetValue()
 				}
```
eahydra, line 56: I don't think we need to do that. We should report all original labels.

pkg/metrics/store/sqlite.go
```diff
@@ -117,25 +126,26 @@ func insert(ctx context.Context, dbRW *sql.DB, table string, ms ...pkgmetrics.Me
...
-		"INSERT OR REPLACE INTO %s (%s, %s, %s, %s, %s) VALUES ",
+		"INSERT OR REPLACE INTO %s (%s, %s, %s, %s, %s, %s) VALUES ",
 		table,
 		ColumnUnixMilliseconds,
 		ColumnComponentName,
 		ColumnMetricName,
-		ColumnMetricLabel,
 		ColumnMetricValue,
+		ColumnMetricLabelName,
+		ColumnMetricLabelValue,
```
eahydra, line 136: I have always had a question that how we store these metrics if there are many label keys and values for the metrics.

gyuho: @eahydra  Our assumption is, we only allow one custom/secondary label (e.g., GPU UUID, mount point) to restrict the number of columns, and also support SQL selects based on the GPU UUID or mount points.

eahydra: Hmm.. I don't think the assumption is true.. It limits scalability.

eahydra (approved):

PR #715, opened 2025-04-27
feat(custom-plugins): support run mode, embed last run timestamp, fix e2e tests

gyuho:

### DIFF
--- a/pkg/custom-plugins/component.go
+++ b/pkg/custom-plugins/component.go
@@ -63,7 +63,7 @@ func (c *component) Name() string { return c.spec.ComponentName() }
 func (c *component) Start() error {
 	log.Logger.Infow("starting custom plugin", "type", c.spec.Type, "component", c.Name(), "plugin", c.spec.PluginName)
 
-	if c.spec.Mode == "manual" {
+	if c.spec.Mode == SpecModeManual {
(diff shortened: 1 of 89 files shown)


pkg/custom-plugins/component.go
```diff
@@ -63,6 +63,11 @@ func (c *component) Name() string { return c.spec.ComponentName() }
 func (c *component) Start() error {
 	log.Logger.Infow("starting custom plugin", "type", c.spec.Type, "component", c.Name(), "plugin", c.spec.PluginName)
 
+	if c.spec.Mode == "manual" {
```
eahydra, line 66: make the `manual` as a const value

pkg/custom-plugins/component.go
```diff
@@ -94,8 +99,11 @@ func (c *component) Check() components.CheckResult {
 		componentName: c.Name(),
 		pluginName:    c.spec.PluginName,
 		ts:            time.Now().UTC(),
-		extraInfo:     make(map[string]string),
 	}
+	cr.extraInfo = map[string]string{
+		"last_check_ts_unix_seconds": fmt.Sprintf("%d", cr.ts.Unix()),
+	}
```
eahydra, line 105: Don't add it in the extra_info, let's make it as a common field in HealthState

eahydra (commented):
overall lgtm

eahydra (approved):

PR #718, opened 2025-04-28
fix(custom-plugins): fix non-zero exit code commands output parsing, rename "mode" to "run_mode", add e2e tests for manual mode plugin

gyuho:
(to be more distinct from "type" field)

### DIFF
--- a/pkg/custom-plugins/component.go
+++ b/pkg/custom-plugins/component.go
@@ -154,6 +154,9 @@ func (c *component) Check() components.CheckResult {
 		}
 	}
 
+	// we still parsed the output above
+	// even when the command/script had failed
+	// e.g., command failed with non-zero exit code
 	if cr.err != nil {
 		cr.health = apiv1.HealthStateTypeUnhealthy
 		cr.reason = fmt.Sprintf("error executing state plugin -- %s (exit code: %d)", cr.err, cr.exitCode)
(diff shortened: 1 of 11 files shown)


pkg/process/runner_exclusive.go
```diff
@@ -77,7 +77,8 @@ func (er *exclusiveRunner) RunUntilCompletion(ctx context.Context, script string
 
 	case err := <-p.Wait():
 		if err != nil {
-			return nil, p.ExitCode(), err
+			output, _ := os.ReadFile(tmpFile.Name())
```
gyuho, line 80: this is the bug fix... we should still return the output

pkg/custom-plugins/component.go
```diff
@@ -162,6 +154,19 @@ func (c *component) Check() components.CheckResult {
 		}
 	}
 
+	if cr.err != nil {
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = fmt.Sprintf("error executing state plugin -- %s (output: %s)", cr.err, string(cr.out))
```
eahydra, line 159: don't populate the reason with the cr.out since the script may print lots of text.

gyuho: fixed

eahydra (approved):

PR #724, opened 2025-04-29
feat(components): make reason field message more static

gyuho:


gyuho:
Closing to split this into smaller ones

PR #730, opened 2025-04-29
feat(os, kernel, info): make reason field message more static

gyuho:

### DIFF
--- a/components/os/component.go
+++ b/components/os/component.go
@@ -123,7 +123,8 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = fmt.Sprintf("error getting uptime: %s", err)
+		cr.reason = "error getting uptime"
+		log.Logger.Errorw(cr.reason, "error", cr.err)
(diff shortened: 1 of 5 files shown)


components/os/component.go
```diff
@@ -123,7 +123,8 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = fmt.Sprintf("error getting uptime: %s", err)
+		cr.reason = "error getting uptime"
```
eahydra, line 126: We also need to consider deleting the component.  What's the `VirtualizationEnvironment`  used for?

gyuho: Yeah let us remove once machine info is working well :)

`VirtualizationEnvironment` is not really being used in the os component itself. Here just providing as an info.

eahydra: I mean that we would add it in the machine info if we need it.

eahydra (approved):

components/os/component.go
gyuho: Yes, let me know when we need it, we can add this to the machine info.

PR #741, opened 2025-04-29
feat(nvidia/info): make reason field message more static

gyuho:

### DIFF
--- a/components/accelerator/nvidia/info/component.go
+++ b/components/accelerator/nvidia/info/component.go
@@ -145,7 +145,8 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = fmt.Sprintf("error getting device count: %s", err)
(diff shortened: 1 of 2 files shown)


components/accelerator/nvidia/info/component.go
```diff
@@ -145,7 +145,8 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = fmt.Sprintf("error getting device count: %s", err)
+		cr.reason = "error getting device count"
```
eahydra, line 148: Shall we delete the component? There is same data in MachineInfo.

gyuho: Useful for `gpud scan` :)

I can create a separate PR to replace it with machine info.

eahydra: `gpud scan` should rely on MachineInfo.

eahydra (approved):

PR #742, opened 2025-04-29
feat(components): add component_type

gyuho:

### DIFF
--- a/api/v1/types.go
+++ b/api/v1/types.go
@@ -25,6 +25,16 @@ const (
 	HealthStateTypeInitializing HealthStateType = "Initializing"
 )
 
+// ComponentType defines the type of a component.
+type ComponentType string
+
+const (
+	// ComponentTypeComponent represents a native component of GPUd.
+	ComponentTypeComponent ComponentType = "component"
+	// ComponentTypeCustomPlugin represents a custom plugin of GPUd.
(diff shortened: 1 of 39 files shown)


api/v1/types.go
```diff
@@ -25,6 +25,16 @@ const (
 	HealthStateTypeInitializing HealthStateType = "Initializing"
 )
 
+// ComponentType defines the type of a component.
+type ComponentType string
+
+const (
+	// ComponentTypeComponent represents a native component of GPUd.
+	ComponentTypeComponent ComponentType = "component"
```
eahydra, line 33: All components are a component...We don't need to define the value

api/v1/types.go
```diff
@@ -33,10 +43,14 @@ type HealthState struct {
 	// Time represents when the event happened.
 	Time metav1.Time `json:"time"`
 
-	// Component represents which component generated the state.
+	// Component represents the component name.
 	Component string `json:"component,omitempty"`
+	// ComponentType represents the type of the component.
+	// It is either "component" or "custom-plugin".
+	ComponentType ComponentType `json:"component_type,omitempty"`
```
eahydra, line 50: If the field is empty, it is component.

gyuho: done

eahydra (approved):

PR #743, opened 2025-04-29
feat(components): remove "encoding: json" from extra info

gyuho:
Some of our control plane still uses `data` field... let's drop encoding field first.

### DIFF
--- a/components/accelerator/nvidia/bad-envs/component.go
+++ b/components/accelerator/nvidia/bad-envs/component.go
@@ -239,7 +239,9 @@ func (cr *checkResult) HealthStates() apiv1.HealthStates {
 		Health:    cr.health,
 	}
 
-	b, _ := json.Marshal(cr)
-	state.ExtraInfo = map[string]string{"data": string(b)}
+	if len(cr.FoundBadEnvsForCUDA) > 0 {
+		b, _ := json.Marshal(cr)
+		state.ExtraInfo = map[string]string{"data": string(b)}
+	}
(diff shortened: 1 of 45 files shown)


components/accelerator/nvidia/bad-envs/component.go
```diff
@@ -240,9 +240,6 @@ func (cr *checkResult) HealthStates() apiv1.HealthStates {
 	}
 
 	b, _ := json.Marshal(cr)
```
eahydra, line 242: Shall we check if the cr.FoundBadEnvsForCUDA is empty ? If so, we can skip marshal it and don't update the extra info.

components/accelerator/nvidia/remapped-rows/component.go
```diff
@@ -189,9 +189,8 @@ func (c *component) Check() components.CheckResult {
 					Type:    apiv1.EventTypeWarning,
 					Message: fmt.Sprintf("%s detected pending row remapping", uuid),
 					DeprecatedExtraInfo: map[string]string{
```
eahydra, line 191: we can remove it now

components/accelerator/nvidia/bad-envs/component.go
gyuho: done

components/accelerator/nvidia/remapped-rows/component.go
gyuho: removed

PR #751, opened 2025-05-01
fix(nvidia/xid): set unhealthy state if any critical errors are found in gpud scan

gyuho:


components/accelerator/nvidia/xid/component.go
```diff
@@ -281,9 +290,9 @@ func (cr *checkResult) String() string {
 			action = strings.Join(actions, ", ")
 		}
 
-		critical := false
+		criticality := "unknown"
```
eahydra, line 293: Should change the corresponding header type.

eahydra (approved):

PR #755, opened 2025-05-02
feat(eventstore): simplify event type, remove deprecated fields, increase unit test coverage

gyuho:
- **feat(eventstore): define its own row type**
- **feat(eventstore): simplify event type, remove deprecated fields**

### DIFF
--- a/components/accelerator/nvidia/fabric-manager/log_watcher.go
+++ b/components/accelerator/nvidia/fabric-manager/log_watcher.go
@@ -9,7 +9,6 @@ import (
 	"time"
 
 	"github.com/patrickmn/go-cache"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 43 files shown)


components/accelerator/nvidia/fabric-manager/log_watcher.go
```diff
@@ -53,30 +52,21 @@ func (llp *logLineProcessor) watch() {
...
-			ev := apiv1.Event{
-				Time: metav1.Time{Time: line.ts.UTC()},
-				Type: apiv1.EventTypeWarning,
-				DeprecatedExtraInfo: map[string]string{
+			ev := eventstore.Event{
+				Time: line.ts.UTC(),
+				Type: string(apiv1.EventTypeWarning),
+				ExtraInfo: map[string]string{
 					"log_line": line.content,
 				},
```
eahydra, line 60: we don't need the line content.

eahydra (approved):

PR #765, opened 2025-05-03
Add support for component_list in custom plugin

mikekg:
Add support for component_list in custom plugin


mikekg:
Operator error, not ready.

docs/PLUGIN.md
```diff
@@ -0,0 +1,314 @@
...
+   - Just the component name
+
+### Parameter Inheritance and Priority
+
+The plugin system supports parameter inheritance with the following priority order:
+
+1. **Run Mode Priority**:
+   - Highest priority: Component-specific run_mode (e.g., `name#run_mode`)
+   - Middle priority: Parent plugin's run_mode
+   - Lowest priority: Default run_mode (auto)
```
gyuho, line 81: nice! great idea

docs/PLUGIN.md
```diff
@@ -0,0 +1,314 @@
...
+- **Multi-Component**: Multiple health checks from a single configuration, either specified directly or loaded from a file
+
+## Configuration Format
+
+Plugins are configured using YAML files. Here's the basic structure:
+
+```yaml
+plugin_name: string  # Required, max 128 chars
+type: string        # Required, one of: "init", "component", "component_list"
+run_mode: string    # Optional, defaults to "manual"
```
gyuho, line 38:
```suggestion
run_mode: string    # Optional, defaults to "auto"
```

Do we want to treat empty `run_mode` as manual? The current logic is, we only check `manual` to skip periodic check calls. All other values including empty are NOT being used.

https://github.com/leptonai/gpud/blob/4cbd7434e79d26aade05d127562c9d425cb49cce/pkg/custom-plugins/component.go#L69-L72

docs/PLUGIN.md
```diff
@@ -0,0 +1,314 @@
...
+timeout: duration   # Optional, defaults to 1 minute
```
gyuho, line 39:
```suggestion
timeout: duration   # Optional, defaults to 1 minute (e.g., "1m")
```

docs/PLUGIN.md
```diff
@@ -0,0 +1,314 @@
...
+        script: string       # Required, bash script content
+```
+
+## Component List Format
+
+Each component in the list can be specified in one of four formats:
+
+1. **Full Format**: `name#run_mode:param`
+   - `name`: Component name
+   - `run_mode`: Optional run mode (auto, manual, once)
```
gyuho, line 60: +1 for `once`. Can we document `once` mode has not been implemented?

Something like:

```suggestion
   - `run_mode`: Optional run mode (possible values: auto, manual) -- "once" mode to be supported in the future
```

docs/PLUGIN.md
```diff
@@ -0,0 +1,314 @@
...
+- Their health states are tracked
+- They can be queried through the API
+- They participate in the overall system health status
+
+## API Access
+
+Plugins can be accessed through GPUd's API:
+- List all plugins: `GET /v1/components/custom-plugin`
+- Get plugin status: `GET /v1/states?component=<plugin_name>`
+- Trigger manual check: `POST /v1/components/check?componentName=<plugin_name>`
```
gyuho, line 314:
```suggestion
- Trigger manual check: `GET /v1/components/trigger-check?componentName=<plugin_name>` 
```

For example,

> curl -s -kL https://localhost:15132/v1/components/trigger-check?componentName=automatic-exit-1 | jq 

https://github.com/leptonai/gpud/blob/4cbd7434e79d26aade05d127562c9d425cb49cce/pkg/server/handlers_components.go#L134-L151

pkg/custom-plugins/spec.go
```diff
@@ -62,9 +57,144 @@ const (
...
+			}
+
+			// Copy and substitute each step
+			for i, step := range spec.HealthStatePlugin.Steps {
+				if step.RunBashScript != nil {
+					// Substitute parameters in the script
+					script := step.RunBashScript.Script
+					script = strings.ReplaceAll(script, "${NAME}", name)
+					script = strings.ReplaceAll(script, "${PAR}", param)
+					script = strings.ReplaceAll(script, "${PAR1}", param)
```
gyuho, line 175: What's the difference between `${PAR}` and `${PAR1}`?

gyuho:
overall lgtm

pkg/custom-plugins/spec.go
mikekg: No difference.  And maybe we clean it out.  I was thinking maybe we want eventually more than 1 PAR, so PAR wiuld be the sum, and then we're have PAR1, PAR2 like there's "$@" and $1 $2 $3.... but maybe that's over the top.  

Parameters really should come from a cluster config, not be strewn into the binding between two components.  (Although that's a practical and useful safety valve, and a good path to ramp up full plug-in integration).

If you think we should just make PAR1 go away for now (and bring it back if we ever need more than 1 PARameter) -- looking for opinions and perspectives.

mikekg: OK, I just removed support for ${PAR1}.  The fact that the explanation was longer than the code was a good indicator that the ability to use either name would be confusing to future users.  Thanks for your original question flagging this!

docs/PLUGIN.md
```diff
@@ -0,0 +1,325 @@
...
+type: "component"
+run_mode: "auto"
+interval: "1m"
+timeout: "30s"
+steps:
+  - name: "check-disk-space"
+    run_bash_script:
+      script: |
+        #!/bin/bash
+        df -h | grep ${NAME}
```
mikekg, line 103:
```suggestion
        df -h 
```

docs/PLUGIN.md
```diff
@@ -0,0 +1,325 @@
...
+        else
+          echo "Disk space check failed for ${NAME}"
+          exit 1
+        fi
+```
+
+#### Component List File
+```yaml
+plugin_name: "multi-disk-check"
+type: "component"
```
mikekg, line 135:
```suggestion
type: "component_list"
```

docs/PLUGIN.md
```diff
@@ -0,0 +1,325 @@
...
+run_mode: "manual"
+timeout: 10s
+health_state_plugin:
+  steps:
+    - name: "check-disk-space"
+      run_bash_script:
+        content_type: "plaintext"
+        script: |
+          #!/bin/bash
+          df -h | grep ${NAME}
```
mikekg, line 214:
```suggestion
          df -h | grep /var
```

docs/PLUGIN.md
```diff
@@ -0,0 +1,325 @@
...
+Plugins are configured using YAML files. Here's the basic structure:
+
+```yaml
+plugin_name: string  # Required, max 128 chars
+type: string        # Required, one of: "init", "component", "component_list"
+run_mode: string    # Optional, defaults to "auto"
+timeout: duration   # Optional, defaults to 1 minute (e.g., "1m")
+interval: duration  # Optional, must be >= 1 minute if specified
+
+# For component_list type
```
mikekg, line 42:
```suggestion
# For component_list type, specify exactly one of:
```

mikekg (commented):
Documentation tweaks and clarifications

gyuho (approved):
lgtm!

PR #768, opened 2025-05-03
feat(session): support logout session request

cardyok:


pkg/session/serve.go
```diff
@@ -126,7 +130,24 @@ func (s *Session) serve() {
...
+				log.Logger.Errorw("failed to open state file", "error", err)
+				break
+			}
+			if err := gpudstate.DeleteLoginInfo(ctx, dbRW); err != nil {
+				log.Logger.Errorw("failed to delete login info", "error", err)
+			}
+			dbRW.Close()
+			if _, err = exec.CommandContext(ctx, "systemctl", "stop", "gpud").Output(); err != nil {
+				log.Logger.Errorw("failed to stop gpud service", "error", err)
+			}
```
eahydra, line 150: shall we delay exec this command like reboot?

cardyok: yea we can do that..lemme update
initially thought since we are going to stop gpud anyway so prob do not need to wait for resp

eahydra (approved):

PR #774, opened 2025-05-06
feat(components): add "IsSupported" method to skip GPU components in CPU machines

gyuho:
If GPUd runs on CPU-only machines, we shouldn't report nvidia/accelerator components.


eahydra (approved):
lgtm

PR #776, opened 2025-05-06
feat(disk): track usage of NFS partitions

gyuho:


components/disk/component.go
```diff
@@ -49,29 +55,45 @@ func DefaultFsTypeFunc(fsType string) bool {
 		fsType == "raid0"
 }
 
+func DefaultExt4FsTypeFunc(fsType string) bool {
+	return fsType == "ext4"
+}
+
+func DefaultNFSFsTypeFunc(fsType string) bool {
+	// ref. https://www.weka.io/
+	return fsType == "wekafs"
```
eahydra, line 64: Not only wekafs, but also nfs

components/disk/component.go
```diff
@@ -204,13 +226,41 @@ func (c *component) Check() components.CheckResult {
...
+			prevFailed = true
+			continue
+		}
+
+		cr.NFSPartitions = parts
+		if prevFailed {
+			log.Logger.Infow("successfully got nfs partitions after retries", "num_partitions", len(parts))
+		}
+		break
+	}
```
eahydra, line 259: shall we wrap these lines to a function to improve the readability?

eahydra (approved):

PR #778, opened 2025-05-07
feat(components): add Tags method to interface

gyuho:
To support https://github.com/leptonai/gpud/pull/777 better, which needs to trigger components by tag matching.

### DIFF
--- a/components/accelerator/nvidia/bad-envs/component.go
+++ b/components/accelerator/nvidia/bad-envs/component.go
@@ -52,7 +52,7 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 
 func (c *component) Name() string { return Name }
 
-func (c *component) Labels() []string {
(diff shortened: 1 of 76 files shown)


components/accelerator/nvidia/bad-envs/component.go
```diff
@@ -52,6 +52,22 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 
 func (c *component) Name() string { return Name }
 
+func (c *component) Labels() []string {
+	return []string{
+		"accelerator",
+		"gpu",
+		"nvidia",
+		Name,
```
eahydra, line 60: It seems that these are tags not labels. Labels are composed of keys and values..

gyuho: Yes, actually, tags make more sense. Reverted

eahydra (approved):

PR #783, opened 2025-05-08
feat(machine-info): include private IP device info

gyuho:

### DIFF
--- a/api/v1/types.go
+++ b/api/v1/types.go
@@ -3,6 +3,7 @@ package v1
 import (
 	"fmt"
 	"io"
+	"net/netip"
 	"strings"
 	"time"
 
@@ -266,6 +267,8 @@ type MachineInfo struct {
 	GPUInfo *MachineGPUInfo `json:"gpuInfo,omitempty"`
 	// DiskInfo is the Disk info of the machine.
 	DiskInfo *MachineDiskInfo `json:"diskInfo,omitempty"`
+	// NetworkInfo is the network info of the machine.
+	NetworkInfo *MachineNetworkInfo `json:"networkInfo,omitempty"`
 }
 
 func (i *MachineInfo) RenderTable(wr io.Writer) {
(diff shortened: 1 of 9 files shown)


pkg/machine-info/machine_info.go
```diff
@@ -45,7 +46,8 @@ func GetMachineInfo(nvmlInstance nvidianvml.Instance) (*apiv1.MachineInfo, error
 		Hostname:                hostname,
 		Uptime:                  metav1.NewTime(time.Unix(int64(pkghost.BootTimeUnixSeconds()), 0)),
 
-		CPUInfo: GetMachineCPUInfo(),
+		CPUInfo:     GetMachineCPUInfo(),
+		NetworkInfo: GetMachineNetwork(),
```
eahydra, line 50: GetMachineNetwork only returns the public ip, we also need the private ip.
we should add a new structure to collect all physical nic devices including the IB devices.

eahydra (approved):

pkg/machine-info/machine_info.go
eahydra: Shall we report the private ip specified when login for now?

api/v1/types.go
```diff
@@ -395,11 +398,39 @@ func (di *MachineDiskInfo) RenderTable(wr io.Writer) {
...
+	Interface string `json:"interface,omitempty"`
+
+	// MAC is the MAC address of the machine.
+	MAC string `json:"mac,omitempty"`
+
+	// IP is the public IP address of the machine.
+	IP string `json:"ip,omitempty"`
+
+	// PrivateIP is the private IP address of the machine.
+	PrivateIP string `json:"privateIP,omitempty"`
```
eahydra, line 427: Shall we just keep one of them?

api/v1/types.go
```diff
@@ -395,11 +398,39 @@ func (di *MachineDiskInfo) RenderTable(wr io.Writer) {
...
+
+	// Addr is the netip.Addr of the machine.
+	Addr netip.Addr `json:"-"`
```
eahydra, line 430: What's it? Shall we remove it?

gyuho: Used to pick IPv4 private IP

eahydra (approved):

PR #785, opened 2025-05-08
feat(components/os): combine "file descriptor" into "os", remove "file-descriptor" component

gyuho:


components/fd/component.go
```diff
@@ -294,6 +326,9 @@ type checkResult struct {
 	// Set to true if the file descriptor limit is supported.
 	FDLimitSupported bool `json:"fd_limit_supported"`
 
+	// ZombieProcesses is the number of zombie processes.
+	ZombieProcesses int `json:"zombie_processes"`
```
eahydra, line 330: the fd component looks like an OS component..  shall we move the fd component to the os?

gyuho: sure

eahydra (approved):

PR #792, opened 2025-05-09
Add description of plugin output parsing to PLUGIN.md

mikekg:
Expand docs for handling error check plugins


docs/PLUGIN.md
```diff
@@ -30,6 +30,10 @@ Component plugins run periodically to monitor system health. They can be:
 - **Single Component**: One health check with one configuration
 - **Multi-Component**: Multiple health checks from a single configuration, either specified directly or loaded from a file
 
+GPUd uses the exit status of plugins to determine success or failure of a plugin.  Ensure that plugins return an error code
```
mikekg, line 33: how does this relate to the success field?

gyuho: GPUd has no concept of success, by default. It's either healthy/unhealthy/degraded.

In general, any plugin/component that fails the check with non-zero exit code is set to unhealthy.

But some exceptions are, the command exits with 1 but the "component" can decide to discard it. See the example below, where we discard exit code 255, but ignores when there's a partial output:

https://github.com/leptonai/gpud/blob/d08d7ebb572872e529f09f747932a1bd5a09dfad/components/accelerator/nvidia/infiniband/component.go#L241-L256

So it is possible for a custom plugin to discard non-zero exit code.

gyuho (approved):
lgtm!

PR #797, opened 2025-05-12
feat(gpud): remove unused tables, new metadata tables for gpud states/logins

gyuho:


pkg/machine-info/login_request.go
```diff
@@ -9,10 +9,9 @@ import (
 	nvidianvml "github.com/leptonai/gpud/pkg/nvidia-query/nvml"
 )
 
-func CreateLoginRequest(token string, nvmlInstance nvidianvml.Instance, machineID string, gpuCount string, privateIP string, publicIP string) (*apiv1.LoginRequest, error) {
+func CreateLoginRequest(token string, nvmlInstance nvidianvml.Instance, gpuCount string, privateIP string, publicIP string) (*apiv1.LoginRequest, error) {
```
eahydra, line 12: @cardyok PTAL. I think we still need to the machineID?

cardyok: yes we still need the machine ID, for manual case we do not need it, but other cases we still need machineID

gyuho: reverted

eahydra (approved):

PR #802, opened 2025-05-13
feat(nvidia/persitence-mode): unhealthy if supported but not enabled

gyuho:


components/accelerator/nvidia/persistence-mode/component.go
```diff
@@ -142,8 +143,24 @@ func (c *component) Check() components.CheckResult {
...
+		}
+	}
+
+	if len(notEnabled) > 0 {
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = strings.Join(notEnabled, ", ")
+
+		if len(notEnabled) == len(cr.PersistenceModes) {
+			cr.reason = fmt.Sprintf("all %d GPU(s) persistence mode supported but not enabled", len(devs))
+		}
```
eahydra, line 159:
```suggestion
		if len(notEnabled) == len(cr.PersistenceModes) {
			cr.reason = fmt.Sprintf("all %d GPU(s) disabled persistence mode", len(devs))
		} else {
		    cr.reason = strings.Join(notEnabled, ", ")
		}
```

gyuho: @eahydra Done

eahydra (approved):

PR #805, opened 2025-05-13
feat(gpud/cmd/common): define common terminal colors

gyuho:
to help refactor cmd/ directory


eahydra (approved):
BTW, we should polish the logs when running "gpud up/join/scan", unless it is error level log, many logs should be hidden. Like "gpud run", we don't need to hide the logs because it runs in the background.

PR #811, opened 2025-05-14
feat(gpud join): reduce noise in logs

gyuho:
- **feat(gpud up/join): reduce noise in logs**
- **remove default**

### DIFF
--- a/pkg/netutil/latency/edge/derp.go
+++ b/pkg/netutil/latency/edge/derp.go
@@ -30,7 +30,7 @@ func measureDERP(ctx context.Context, targetDerpServrs *tailcfg.DERPMap, opts ..
 		return nil, err
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/netutil/latency/edge/derp.go
```diff
@@ -30,7 +30,7 @@ func measureDERP(ctx context.Context, targetDerpServrs *tailcfg.DERPMap, opts ..
 		return nil, err
 	}
 
-	log.Logger.Debugw("measuring latencies to public tailscale DERP nodes")
+	log.Logger.Infow("measuring latencies to public tailscale DERP nodes")
```
eahydra, line 33: shall we print it just in up or join sub-command?

gyuho: done

eahydra:
@gyuho can you run these command to ensure the logs are hidden and the logs are readability?

PR #814, opened 2025-05-15
doc(README): explain how to install a specific version

gyuho:


mikekg (approved):
Looks good.  

Is there a way for users to discover what versions are supported?

gyuho:
@mikekg For now, user can search the github release page, but good idea to support these in GPUd like version lists

PR #819, opened 2025-05-16
fix(httputil): add "CreateURL", handle empty host in URL, fix e2e tests

gyuho:
For inputs like `:12345`

Fixed

> {"level":"info","ts":"2025-05-16T08:45:38Z","msg":"waiting for server to be ready","endpoint":"https://0.0.0.0:15132"}
{"level":"info","ts":"2025-05-16T08:45:39Z","msg":"/healthz","status":200,"method":"GET","path":"/healthz","query":"","ip":"127.0.0.1","user-agent":"Go-http-client/1.1","latency":0.000022086,"time":"2025-05-16T08:45:39Z"}
[GIN] 2025/05/16 - 08:45:39 | 200 |      60.042µs |       127.0.0.1 | GET      "/healthz"

### DIFF
--- a/pkg/session/session.go
+++ b/pkg/session/session.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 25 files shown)


pkg/session/session.go
```diff
@@ -190,21 +198,25 @@ func (s *Session) keepAlive() {
...
+			}
+
 			s.closer = &closeOnce{closer: make(chan any)}
+
 			ctx, cancel := context.WithCancel(context.Background()) // create local context for each session
+
+			readerExit := make(chan any)
+			writerExit := make(chan any)
 			jar, _ := cookiejar.New(nil)
-			if err := s.checkServerHealth(ctx, jar); err != nil {
```
eahydra, line 198: Why do you change these lines? we should use jar to receive the cookie to stick the session on one of our servers...

PR #833, opened 2025-05-16
contrib(helm): add example Helm chart, fix `gpud` install PATH

Lynnery:
## Summary

This PR includes two main improvements which are based on this [PR](https://github.com/leptonai/gpud/pull/123):

### 🛠️ 1. Fix install.sh compatibility with read-only filesystems and environment collisions

- Replaces the `VERSION` variable with `APP_VERSION` in `install.sh` to avoid conflicts with `/etc/os-release`, which commonly defines `VERSION` (e.g., "34 (Workstation Edition)").
- Changes the binary install path from `/usr/sbin` to `/usr/local/bin` to support immutable systems like Fedora CoreOS and OpenShift, where `/usr/sbin` is read-only.

### 🧩 2. Add gpud version configurability via Helm chart

- Introduces `.Values.gpud.version` in `values.yaml`, allowing users to specify which version of `gpud` to install.
- Updates the DaemonSet template to pass this version into the install script (`bash -s {{ .Values.gpud.version }}`).

## Future Suggestions

A Dockerfile would be a more suitable approach once there's a docker.io / registry which holds the image to encapsulate the installation command based on alpine.


install.sh
```diff
@@ -132,9 +132,9 @@ main() {
 
   tar xzf "$DLPATH" -C "$DIR"
 
-  $SUDO cp -f "$DIR"/gpud /usr/sbin
```
gyuho, line 135: Our default systemd relies on `/usr/sbin/gpud`. Any reason why we need `/usr/local/bin`?

gyuho: Oh

> Changes the binary install path from /usr/sbin to /usr/local/bin to support immutable systems like Fedora CoreOS and OpenShift, where /usr/sbin is read-only.

gyuho:
overall lgtm, let us do some testing with existing machines to ensure its compatibility!

Lynnery:
> overall lgtm, let us do some testing with existing machines to ensure its compatibility!

Thanks Gyuho, btw how is the deployment of the install script to the pkg endpoint done?
I was wondering if I can replicate that because I just tested it using the local script instead of the one hosted.

Or is it just a simple upload procedure?

gyuho:
> how is the deployment of the install script to the pkg endpoint done

Oh, should've been clearer. I was referring to our internal approach where we used 

https://github.com/leptonai/gpud/blob/d682b8c651e493ada07cdc2165c006ea0a2b9478/pkg/gpud-manager/systemd/gpud.service#L12

We want to change to `/usr/sbin/gpud`, but would like to make sure our internal deployment process works before we update https://pkg.gpud.dev/install.sh endpoint.

Lynnery:
> > how is the deployment of the install script to the pkg endpoint done
> 
> Oh, should've been clearer. I was referring to our internal approach where we used 
> 
> https://github.com/leptonai/gpud/blob/d682b8c651e493ada07cdc2165c006ea0a2b9478/pkg/gpud-manager/systemd/gpud.service#L12
> 
> We want to change to `/usr/sbin/gpud`, but would like to make sure our internal deployment process works before we update https://pkg.gpud.dev/install.sh endpoint.

I meant that I wonder how do you upload the install script to https://pkg.gpud.dev/install.sh in general (if not through some CI/CD).

Regarding the service Unit file, I guess it can stay in `/usr/sbin` since it's not used in the Kubernetes deployment? (Which needs support for immutable filesystems)

gyuho:
> do you upload the install script to https://pkg.gpud.dev/install.sh in general - but I see it was done through the workflow.

We upload manually. Please wait for us to test this, and update the link.

PR #837, opened 2025-05-17
fix(config/default): always use /var/lib dir for linux

gyuho:
If one forgets to run "gpud" without sudo, and later use sudo,
we have inconsistent behavior.


gyuho:
This breaks CI...

PR #843, opened 2025-05-19
feat(disk): match no space left via kmsg for events

gyuho:

### DIFF
--- a/components/disk/component.go
+++ b/components/disk/component.go
@@ -18,6 +18,8 @@ import (
 	apiv1 "github.com/leptonai/gpud/api/v1"
 	"github.com/leptonai/gpud/components"
 	"github.com/leptonai/gpud/pkg/disk"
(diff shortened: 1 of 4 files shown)


eahydra (approved):

components/disk/component.go
```diff
@@ -87,6 +92,23 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
...
 
+	if gpudInstance.EventStore != nil && runtime.GOOS == "linux" {
+		var err error
+		c.eventBucket, err = gpudInstance.EventStore.Bucket(Name)
+		if err != nil {
+			ccancel()
+			return nil, err
+		}
+
+		if os.Geteuid() == 0 {
```
eahydra, line 103: We have required running gpud with root permission, so shall we remove the condition?

gyuho: Oh let us keep it for now, to make ci pass

PR #846, opened 2025-05-20
feat(install.sh): default to /usr/local/bin (do not use /usr/sbin)

gyuho:
ref. https://github.com/leptonai/gpud/pull/833


cardyok:
still need a symlink at /usr/sbin/gpud for compatibility...once dependency is cleared we can fully remove it

eahydra (approved):

PR #851, opened 2025-05-21
feat(machine-info): detect aws/azure/gcp provider, add "gpud machine-info" command

gyuho:


cmd/gpud/command/command.go
```diff
@@ -558,6 +559,18 @@ sudo gpud join
 				},
 			},
 		},
+		{
+			Name:      "provider",
```
eahydra, line 563: Do we really need a new command?

pkg/providers/detect.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return "", err
+		}
+		if tk != "" {
+			return d.providerName, nil
+		}
+	}
+
+	if d.fetchZoneFunc != nil {
+		zone, err := d.fetchZoneFunc(ctx)
```
eahydra, line 49: these are provider special logic.. we should hide them.

gyuho: Yes, this is not implemented in detect package (each package implements its own). Otherwise, we need to copy and paste this detect.go to each provider, which is fine as well.

pkg/machine-info/asn.go
```diff
@@ -0,0 +1,24 @@
+package machineinfo
+
+import "strings"
+
+func NormalizeASNName(asnName string) string {
```
eahydra, line 5: We have a `asn` package, why not move the file to that?

eahydra (approved):

PR #852, opened 2025-05-21
feat(session): add "triggerComponentCheckByTag"

gyuho:
Signed-off-by: Gyuho Lee <gyuhox@gmail.com>

PR #853, opened 2025-05-21
feat(session): support tag_name in "triggerComponentCheck"

gyuho:
Replace https://github.com/leptonai/gpud/pull/852.

### DIFF
--- a/pkg/session/serve.go
+++ b/pkg/session/serve.go
@@ -268,47 +268,41 @@ func (s *Session) serve() {
 			}
 
 		case "triggerComponentCheck":
+			checkResults := make([]components.CheckResult, 0)
 			if payload.ComponentName != "" {
-				switch payload.TagName {
-				case "":
-					comp := s.componentsRegistry.Get(payload.ComponentName)
-					if comp == nil {
-						log.Logger.Warnw("component not found", "name", payload.ComponentName)
-						response.ErrorCode = http.StatusNotFound
-						break
-					}
-
-					rs := comp.Check()
-					response.States = apiv1.GPUdComponentHealthStates{
-						{
-							Component: payload.ComponentName,
-							States:    rs.HealthStates(),
-						},
-					}
-
-				default:
-					components := s.componentsRegistry.All()
-					for _, comp := range components {
-						matched := false
-						for _, tag := range comp.Tags() {
-							if tag == payload.TagName {
-								matched = true
-								break
-							}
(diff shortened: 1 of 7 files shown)


pkg/session/serve.go
```diff
@@ -48,6 +48,9 @@ type Request struct {
 	// ComponentName is the name of the component to query or deregister.
 	ComponentName string `json:"component_name,omitempty"`
 
+	// TagName is the tag of the component to trigger check.
+	TagName string `json:"tag_name,omitempty"`
```
eahydra, line 52: How to support the case where someone wants to trigger multiple components via a tag?

gyuho: @eahydra User can configure tags via this field

https://github.com/leptonai/gpud/blob/2bb317e8c0002549be6cbde1aa39a59f94df85cc/pkg/custom-plugins/types.go#L73-L77

And send the request with the matching tag

eahydra: I mean, does requests support triggering multiple components selected by tags for checking at once?

gyuho: Yes

pkg/session/serve.go
```diff
@@ -264,19 +269,45 @@ func (s *Session) serve() {
 
 		case "triggerComponentCheck":
 			if payload.ComponentName != "" {
-				comp := s.componentsRegistry.Get(payload.ComponentName)
-				if comp == nil {
-					log.Logger.Warnw("component not found", "name", payload.ComponentName)
-					response.ErrorCode = http.StatusNotFound
-					break
-				}
+				switch payload.TagName {
```
eahydra, line 272: we have to take some time to refactor the file... it's too long to maintenance..

pkg/session/serve.go
```diff
@@ -264,19 +269,45 @@ func (s *Session) serve() {
...
-						Component: payload.ComponentName,
-						States:    rs.HealthStates(),
-					},
+					rs := comp.Check()
+					response.States = apiv1.GPUdComponentHealthStates{
+						{
+							Component: payload.ComponentName,
+							States:    rs.HealthStates(),
+						},
+					}
```
eahydra, line 287: we should merge these lines with the default branch..
1. get components by component name or tags
2. for each component, run check and append the result.. 
3. return

eahydra (approved):
please fix the lint error

PR #854, opened 2025-05-22
feat(info): remove components in favor of `/machine-info`, remove gpud metrics

gyuho:


eahydra:
Shall we remove the component?

eahydra (approved):

PR #867, opened 2025-05-25
feat(session): support pacakge status invocation

cardyok:


gyuho (approved):

api/v1/types.go
```diff
@@ -157,6 +157,19 @@ type ComponentInfo struct {
 
 type GPUdComponentInfos []ComponentInfo
 
+type PackageStatus struct {
+	Name           string        `json:"name"`
+	IsInstalled    bool          `json:"is_installed"`
+	Installing     bool          `json:"installing"`
```
eahydra, line 163: regarding these two fields and status field, we should merge them into one

api/v1/types.go
```diff
@@ -157,6 +157,19 @@ type ComponentInfo struct {
...
+	Progress       int           `json:"progress"`
+	TotalTime      time.Duration `json:"total_time"`
+	Status         bool          `json:"status"`
+	TargetVersion  string        `json:"target_version"`
```
eahydra, line 167: and why do we report the target version?

cardyok: removed.

PR #868, opened 2025-05-25
fix(session): use raw ip instead of sticky session

cardyok:


pkg/session/session.go
```diff
@@ -195,17 +200,21 @@ func (s *Session) keepAlive() {
 			writerExit := make(chan any)
 			s.closer = &closeOnce{closer: make(chan any)}
 			ctx, cancel := context.WithCancel(context.Background()) // create local context for each session
-			jar, _ := cookiejar.New(nil)
 
 			log.Logger.Infow("session keep alive: checking server health")
-			if err := s.checkServerHealth(ctx, jar); err != nil {
+			if err := s.checkServerHealth(ctx); err != nil {
```
gyuho, line 205: +1

eahydra, line 205: The health check should be done with sessionAddr, otherwise the check is meaningless

pkg/session/session.go
```diff
@@ -195,17 +200,21 @@ func (s *Session) keepAlive() {
...
 				cancel()
 				continue
 			}
 
-			go s.startReader(ctx, readerExit, jar)
-			go s.startWriter(ctx, writerExit, jar)
+			sessionAddr, err := resolveIPFromEndpoint(s.epControlPlane)
+			if err != nil {
+				log.Logger.Errorf("session keep alive: error resolving endpoint %s: %v", s.epControlPlane, err)
+				sessionAddr = ""
```
eahydra, line 214: Is it reasonable to connect to the control plane without a specific address?

pkg/session/session.go
```diff
@@ -447,3 +466,85 @@ func (s *Session) Stop() {
...
 }
+
+func resolveIPFromEndpoint(origEndpoint string) (string, error) {
+	if origEndpoint == "" {
+		return "", nil
+	}
+	parsedURL, err := url.Parse(origEndpoint)
+	if err != nil || parsedURL.Host == "" {
+		if !strings.Contains(origEndpoint, "://") {
+			origEndpoint = "http://" + origEndpoint
```
eahydra, line 477: why not use https?

pkg/session/session.go
```diff
@@ -447,3 +466,85 @@ func (s *Session) Stop() {
...
+		if err == nil && resp.StatusCode == 200 {
+			return ipURL.String(), nil
+		}
+	}
+
+	return "", fmt.Errorf("no reachable IP found from DNS resolution")
+}
+
+func resolveIPsViaNslookup(domain string) ([]string, error) {
+	cmd := exec.Command("nslookup", domain)
```
eahydra, line 522: don't use commands to do that. Shall we use https://pkg.go.dev/net#Resolver?

pkg/session/session.go
cardyok: #817 changed endpoint in checkServerHealth to local gpud addr, I thought it is expected

cardyok: currently we can fully remove this function, shall we do it now?

pkg/session/session.go
cardyok: no specific reason, updated to https

eahydra (approved):

PR #877, opened 2025-05-27
Log plugin output with path substitution

mikekg:
Create ability to log plugin outputs via the new log_path option.

### DIFF
--- a/pkg/custom-plugins/plugin_output.go
+++ b/pkg/custom-plugins/plugin_output.go
@@ -66,25 +66,22 @@ func (po *PluginOutputParseConfig) extractExtraInfo(input []byte, pluginName str
 		return nil, nil
 	}
 
-	// Handle logging if LogPath is set
-	if po.LogPath != "" {
-		// Substitute variables in the log path
-		substitutedPath := substituteLogPath(po.LogPath, triggerName, pluginName)
-		if substitutedPath != "" {
-			// Open the log file in append mode, create if it doesn't exist
-			f, err := os.OpenFile(substitutedPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
-			if err != nil {
-				return nil, fmt.Errorf("failed to open log file: %w", err)
-			}
-			defer f.Close()
+	// Substitute variables in the log path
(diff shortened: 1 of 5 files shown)


gyuho:
Will be even better if we pass log file paths to the underlying process runner here:

https://github.com/leptonai/gpud/blob/d8d03e1b79ef397e465cb8d2ce58884cb414c3b4/pkg/process/runner_exclusive.go#L37-L47

but this requires passing the O_APPEND opened file object to the function. I can follow up with this approach in the separate PR.

pkg/custom-plugins/plugin_output.go
```diff
@@ -32,11 +34,57 @@ func (po *PluginOutputParseConfig) Validate() error {
...
+func (po *PluginOutputParseConfig) extractExtraInfo(input []byte, pluginName string, triggerName string) (map[string]extractedField, error) {
 	if po == nil {
 		return nil, nil
 	}
 
+	// Substitute variables in the log path
+	substitutedPath := substituteLogPath(po.LogPath, triggerName, pluginName)
+	if substitutedPath != "" {
+		// Open the log file in append mode, create if it doesn't exist
+		f, err := os.OpenFile(substitutedPath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
eahydra, line 73: here use append option, so how to limit or rotate the file?

mikekg: rotate file can be done by `mv /var/log/slurm-epilog.log /var/log/slurm-epilog-$(date +%Y%m%d).log`.  (or with whatever frequency you'd want to rotate, adjusting time stamps appropriately)

Advantageously, we create the file if it doesn't exist, so moving is a safe operation (and if we worry about the case where the file doesn't exist, rotation could be made to be well-behaved by adding ` || true` or checking for existence beforehand)

Open to other suggestions.

mikekg:
@gyuho I'm open to creating the file once and passing the descriptor, if you think that's better.  (That being said, for the ${PLUGIN} expansion, we'd have to create multiple files.  Or we can drop that expansion - the critical expansion  for now is trigger to capture slurm prolog and epilog)

PR #884, opened 2025-05-29
feat(login): add validation results to login response

gyuho:
- add ValidationResults field to LoginResponse
- print validation results in login command

### DIFF
--- a/api/v1/login.go
+++ b/api/v1/login.go
@@ -27,4 +27,16 @@ type LoginResponse struct {
 
 	// Token is the token used to report data from the machine.
 	Token string `json:"token,omitempty"`
+
+	// ValidationResults is the validation results for the login request.
+	// The validation results are done by the control plane.
+	ValidationResults []ValidationResult `json:"validationResults,omitempty"`
+}
+
+// ValidationResult is the validation result for the login request.
+type ValidationResult struct {
+	Name       string `json:"name"`
+	Valid      bool   `json:"valid"`
+	Reason     string `json:"reason"`
+	Suggestion string `json:"suggestion"`
 }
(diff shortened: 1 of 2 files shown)


api/v1/login.go
```diff
@@ -27,4 +27,10 @@ type LoginResponse struct {
 
 	// Token is the token used to report data from the machine.
 	Token string `json:"token,omitempty"`
+
+	// ValidationResults is the validation results for the login request.
+	// The validation results are done by the control plane.
+	// e.g., "kernel-version": "ok" means the kernel version is valid.
+	// e.g., "kernel-version": "invalid ..." means the kernel version is invalid.
+	ValidationResults map[string]string `json:"validationResults,omitempty"`
```
cardyok, line 35: seems now we need to make this something like
ValidationResults  []ValidationResult 
type ValidationResult struct {
  name string
  valid bool
  reason string
  suggestion string
}

cmd/gpud/login/command.go
```diff
@@ -204,6 +204,17 @@ func Command(cliContext *cli.Context) error {
 		}
 	}
 
+	if len(loginResp.ValidationResults) > 0 {
+		fmt.Printf("validation results:\n")
+		for _, result := range loginResp.ValidationResults {
+			if result.Valid {
+				fmt.Printf("%s %s: ok\n", cmdcommon.CheckMark, result.Name)
+			} else {
+				fmt.Printf("%s %s: %s (invalid)\n", cmdcommon.WarningSign, result.Name, result.Reason)
```
eahydra: missing suggestion.

gyuho: @eahydra Done

PR #892, opened 2025-05-29
feat(pkg/machine-info): measure machine location and public IP async

gyuho:


pkg/machine-info/login_request.go
```diff
@@ -40,24 +40,45 @@ func createLoginRequest(
...
-	req.Provider = detectedProvider.Provider
-	req.ProviderInstanceID = detectedProvider.InstanceID
-	req.Network.PublicIP = detectedProvider.PublicIP
+	// fetching public IP can take awhile
+	// depending on the network latency
+	// run async
+	publicIPCh := make(chan string, 1)
+	go func() {
+		pubIP, err := getPublicIPFunc()
+		if err != nil {
```
eahydra, line 71: we can revert it, it just initials a request to get the out ip..

pkg/machine-info/login_request.go
```diff
@@ -101,5 +122,15 @@ func createLoginRequest(
 		req.Resources["nvidia.com/gpu"] = gpuCnt
 	}
 
+	req.Location = <-machineLocationCh
```
eahydra, line 125: we should move it after line 133.

eahydra (approved):

PR #903, opened 2025-06-04
feat(fabric-manager): only check with >2 GPUs

gyuho:


eahydra:
perhaps we can determine whether there are nvswitch devices and VM env + < 8 GPUs, if so we can check the fm, otherwise we can skip it.

PR #905, opened 2025-06-05
fix(gpud join): use private-ip flag if provided

gyuho:

### DIFF
--- a/cmd/gpud/join/command.go
+++ b/cmd/gpud/join/command.go
@@ -89,12 +89,15 @@ func Command(cliContext *cli.Context) (retErr error) {
 	}
 
 	// assume if not empty, it should have been persisted by the "gpud login" command
-	log.Logger.Debugw("reading private IP from state file")
-	privateIP, err := pkgmetadata.ReadMetadata(rootCtx, dbRO, pkgmetadata.MetadataKeyPrivateIP)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/gpud/join/command.go
```diff
@@ -89,12 +89,15 @@ func Command(cliContext *cli.Context) (retErr error) {
 	}
 
 	// assume if not empty, it should have been persisted by the "gpud login" command
-	log.Logger.Debugw("reading private IP from state file")
-	privateIP, err := pkgmetadata.ReadMetadata(rootCtx, dbRO, pkgmetadata.MetadataKeyPrivateIP)
-	if err != nil {
-		return fmt.Errorf("failed to read private IP: %w", err)
+	privateIP := cliContext.String("private-ip")
+	if privateIP == "" {
```
eahydra, line 93: https://github.com/leptonai/gpud/blob/3ed069ff97520636248208da4f3ec10cb059567b/cmd/gpud/command/command.go#L511-L546
we provide these flags, are there similar issues?

gyuho: @eahydra Confirmed. Other flags do not have these issues:

https://github.com/leptonai/gpud/blob/1d4da70a163d104711f78d73ca81dc22fb296ee5/cmd/gpud/join/command.go#L108-L112

eahydra (approved):

PR #910, opened 2025-06-07
feat(gpud up): support "gpud join" in a single "up" command

gyuho:

### DIFF
--- a/cmd/gpud/up/command.go
+++ b/cmd/gpud/up/command.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/urfave/cli"
 
+	cmdjoin "github.com/leptonai/gpud/cmd/gpud/join"
 	cmdlogin "github.com/leptonai/gpud/cmd/gpud/login"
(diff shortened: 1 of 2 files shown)


cmd/gpud/up/command.go
```diff
@@ -30,14 +31,33 @@ func Command(cliContext *cli.Context) (retErr error) {
...
+	nodeGroup := cliContext.String("node-group")
+	if nodeGroup != "" {
+		log.Logger.Debugw("non-empty --node-group provided, joining")
+		if jerr := cmdjoin.Command(cliContext); jerr != nil {
+			return jerr
+		}
+		log.Logger.Debugw("successfully joined")
+	} else {
+		log.Logger.Infow("no --token provided, skipping join")
+	}
```
eahydra: if the nodegroup is specified, we should pass it in login request, so that we only need one request to login + join.

gyuho: Already done in login command flag parsing `nodeGroup := cliContext.String("node-group")`

eahydra (approved):

PR #911, opened 2025-06-07
feat(os): add suggested actions for "too many zombie processes", other file descriptor states

gyuho:
Add more unit tests.

### DIFF
--- a/components/os/component.go
+++ b/components/os/component.go
@@ -69,19 +69,19 @@ type component struct {
 	checkFileHandlesSupportedFunc func() bool
 	checkFDLimitSupportedFunc     func() bool
 
-	// thresholdAllocatedFileHandles is the number of file descriptors that are currently allocated,
+	// maxAllocatedFileHandles is the number of file descriptors that are currently allocated,
 	// at which we consider the system to be under high file descriptor usage.
-	thresholdAllocatedFileHandles                 uint64
-	thresholdAllocatedFileHandlesPercentDegraded  float64
-	thresholdAllocatedFileHandlesPercentUnhealthy float64
+	maxAllocatedFileHandles             uint64
+	maxAllocatedFileHandlesPctDegraded  float64
+	maxAllocatedFileHandlesPctUnhealthy float64
 
-	// thresholdRunningPIDs is the number of running pids at which
+	// maxRunningPIDs is the number of running pids at which
 	// we consider the system to be under high file descriptor usage.
 	// This is useful for triggering alerts when the system is under high load.
 	// And useful when the actual system fd-max is set to unlimited.
-	thresholdRunningPIDs                 uint64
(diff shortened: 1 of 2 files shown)


components/os/component.go
```diff
@@ -505,7 +537,10 @@ func init() {
 		limit, err := file.GetLimit()
 		if limit > 0 && err == nil {
 			// set to 20% of system limit
-			defaultZombieProcessCountThreshold = int(float64(limit) * 0.20)
+			defaultZombieProcessCountThresholdDegraded = int(float64(limit) * 0.20)
+
+			// set to 40% of system limit
+			defaultZombieProcessCountThresholdUnhealthy = int(float64(limit) * 0.40)
```
eahydra: how about 80%?

components/os/component.go
```diff
@@ -83,9 +85,11 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 
 		rebootEventStore: gpudInstance.RebootEventStore,
 
-		countProcessesByStatusFunc:  process.CountProcessesByStatus,
-		zombieProcessCountThreshold: defaultZombieProcessCountThreshold,
+		countProcessesByStatusFunc:           process.CountProcessesByStatus,
+		zombieProcessCountThresholdDegraded:  defaultZombieProcessCountThresholdDegraded,
+		zombieProcessCountThresholdUnhealthy: defaultZombieProcessCountThresholdUnhealthy,
```
eahydra, line 100: not only zombie process but also other resources such as fd count?

gyuho: done

components/os/component.go
gyuho: done

components/os/component.go
```diff
@@ -83,18 +95,25 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
...
 
-		thresholdAllocatedFileHandles: DefaultThresholdAllocatedFileHandles,
-		thresholdRunningPIDs:          DefaultThresholdRunningPIDs,
+		thresholdAllocatedFileHandles:                 DefaultThresholdAllocatedFileHandles,
+		thresholdAllocatedFileHandlesPercentDegraded:  defaultThresholdAllocatedFileHandlesPercentDegraded,
+		thresholdAllocatedFileHandlesPercentUnhealthy: defaultThresholdAllocatedFileHandlesPercentUnhealthy,
+
+		thresholdRunningPIDs:                 DefaultThresholdRunningPIDs,
+		thresholdRunningPIDsPercentDegraded:  defaultThresholdRunningPIDsPercentDegraded,
+		thresholdRunningPIDsPercentUnhealthy: defaultThresholdRunningPIDsPercentUnhealthy,
```
eahydra: shall we make these names shorter? loll..

gyuho: done

PR #922, opened 2025-06-12
doc(components/*): document /v1/states API health field behavior

gyuho:
Fix https://github.com/leptonai/gpud/issues/595.

PR #925, opened 2025-06-13
feat(network/latency): disable periodic check, set "degraded" health state on egress check issues

gyuho:
Need to update https://github.com/leptonai/gpud/pull/922.


eahydra:
Why do we need the latency component? Even the latency is slow, would it affect the machine?

gyuho:
@eahydra We had this since the beginning, maybe not needed anymore. How about we do this once a day?

eahydra:
> @eahydra We had this since the beginning, maybe not needed anymore. How about we do this once a day?

we would remove it if possible. or we can just run in `gpud scan`?

gyuho:
@eahydra Now the periodic check is disabled. User can only run manually (by clicking the manual trigger button), or `gpud scan`.

eahydra (approved):

PR #927, opened 2025-06-13
feat(server): retain only 24 hours of metrics/events data (retain xid/sxid/reboot as exceptions)

gyuho:

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -127,7 +127,7 @@ func New(ctx context.Context, auditLogger log.AuditLogger, config *lepconfig.Con
 		return nil, fmt.Errorf("failed to create metadata table: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -125,7 +125,7 @@ func New(ctx context.Context, config *lepconfig.Config, packageManager *gpudmana
 		return nil, fmt.Errorf("failed to create metadata table: %w", err)
 	}
 
-	eventStore, err := eventstore.New(dbRW, dbRO, 0)
+	eventStore, err := eventstore.New(dbRW, dbRO, 24*time.Hour)
```
gyuho, line 130: @eahydra `0` disables purge. We should set non-zero value here, so db can remain stably low (`gpud compact` will shrink the actual db size).

cardyok (approved):

PR #929, opened 2025-06-13
feat(database): support purge events/metrics data in "compact" command

gyuho:

### DIFF
--- a/cmd/gpud/compact/command.go
+++ b/cmd/gpud/compact/command.go
@@ -3,6 +3,7 @@ package compact
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/dustin/go-humanize"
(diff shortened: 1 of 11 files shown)


cmd/gpud/compact/command.go
```diff
@@ -69,6 +73,48 @@ func Command(cliContext *cli.Context) error {
 	}
 	log.Logger.Infow("state file size before compact", "size", humanize.Bytes(dbSize))
 
+	purgeEvents := cliContext.String("purge-events")
+	if purgeEvents != "" {
+		componentNames := strings.Split(purgeEvents, ",")
+		if purgeEvents == "*" {
+			componentNames = []string{}
+		}
```
eahydra, line 81: Why do we only purge by component for events?

PR #932, opened 2025-06-14
perf(pkg, server): pre-allocate buffer/slice, use atomic vs. mutex

gyuho:
Some basic performance improvements.


gyuho:
Probably, `buffer` 1024 bytes allocation in for-loop is most obvious one

eahydra:
> Probably, `buffer` 1024 bytes allocation in for-loop is most obvious one

I don't think it's an issue.

gyuho:
@eahydra Let's revisit this, when we observe any perf/resource usage issue.

gyuho:
-

PR #935, opened 2025-06-17
feat(pkg/log): add audit logger

gyuho:


pkg/log/audit.go
```diff
@@ -0,0 +1,156 @@
...
+		})
+	} else {
+		w = zapcore.AddSync(os.Stdout)
+	}
+
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.LevelKey = ""
+	encoderConfig.MessageKey = ""
+	encoderConfig.CallerKey = ""
+	encoderConfig.EncodeTime = zapcore.TimeEncoderOfLayout(time.RFC3339)
```
eahydra: shall we encode time with milliseconds?

pkg/server/handlers_components.go
```diff
@@ -255,6 +255,14 @@ const URLPathStates = "/states"
 // @Failure 500 {object} map[string]interface{} "Internal server error"
 // @Router /v1/states [get]
 func (g *globalHandler) getHealthStates(c *gin.Context) {
+	g.cfg.AuditLogger.Log(
```
eahydra: IMO we don't need to audit this kind of request.

pkg/server/handlers_components.go
```diff
@@ -338,6 +346,14 @@ const URLPathEvents = "/events"
 // @Failure 500 {object} map[string]interface{} "Internal server error"
 // @Router /v1/events [get]
 func (g *globalHandler) getEvents(c *gin.Context) {
+	g.cfg.AuditLogger.Log(
```
eahydra: we should skip all these kinds of read requests

pkg/config/config.go
```diff
@@ -10,10 +10,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	nvidia_common "github.com/leptonai/gpud/pkg/config/common"
+	"github.com/leptonai/gpud/pkg/log"
 )
 
 // Config provides gpud configuration data for the server
 type Config struct {
+	AuditLogger log.AuditLogger `json:"-"`
```
eahydra: it's a bit weird that put the logger interface in the structure. we should keep the same style as eventstore.

gyuho: @eahydra Done

pkg/log/audit.go
gyuho: Other GPUd logs use RFC3339, more readable. Any reason to do in ms?

c.f., looks like kube-apiserver uses micro seconds https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#MicroTime

eahydra: actually, I want to change the gpud logs timestamp with milliseconds as well..

pkg/log/audit.go
```diff
@@ -0,0 +1,142 @@
...
+		})
+	} else {
+		w = zapcore.AddSync(os.Stdout)
+	}
+
+	encoderConfig := zap.NewProductionEncoderConfig()
+	encoderConfig.LevelKey = ""
+	encoderConfig.MessageKey = ""
+	encoderConfig.CallerKey = ""
+	encoderConfig.EncodeTime = zapcore.TimeEncoderOfLayout(time.RFC3339Nano)
```
eahydra: ISO8601Time is enough for us. 
```suggestion
	encoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
```

eahydra (approved):

PR #936, opened 2025-06-17
feat(nvidia/gsp-firmware-mode): degraded when GSP firmware mode is supported/enabled

gyuho:


cardyok (approved):
lgtm

Issue #591, opened 2025-04-01
Checking number of GPUs


hmeScaler:
Hi,

We use your tool which is wonderful.
However, we can't find a way to notify us when a GPU disappears (after an error or whatever) and is no longer in the list.
Have you implemented a system for checking the number of GPUs?

Thanks for your help.

gyuho:
Great idea. We do similar checks for infiniband ports, by passing expected GPU counts.

Do you prefer us auto-detect the GPU counts and return an error if the count becomes below? Or expect to pass expected GPU counts yourself (this can be tricky when GPU fractions is involved as well).

hmeScaler:
I think auto-detect is a good idea this would make it possible not to maintain a list of GPUs.
And may be easier for you to implement.
Knowing that the number of GPUs is even may also help if the number of GPUs becomes odd.

gyuho:
Now we have https://github.com/leptonai/gpud/pull/798 (counting GPUs based on `lspci` output). Do you have any preference how to consume this data?

hmeScaler:
Oh great thanks for the work 😊

Now the idea would be to know how Gpud can guess if I lose a GPU during computation.
Let's say I start with 8 GPUs and I lose 1 (so I've got 7 left) how will it raise an alert?
Do you have any ideas or suggestions?

gyuho:
@hmeScaler https://github.com/leptonai/gpud/pull/944 will exactly address that :)

hmeScaler:
> [@hmeScaler](https://github.com/hmeScaler) [#944](https://github.com/leptonai/gpud/pull/944) will exactly address that :)

oh awesome guys !
this is great news and a great feature
thanks a lot 🙏

PR #944, opened 2025-06-20
feat(nvidia/gpu-counts): add gpu counts component

gyuho:
```
{"level":"info","ts":"2025-06-20T19:47:06.578+0530","caller":"gpu-counts/component.go:114","msg":"checking nvidia gpu counts"}

+--------------+-----------------------+
| Product Name | NVIDIA-H100-80GB-HBM3 |
| Count lspci  |           8           |
|  Count NVML  |           8           |
+--------------+-----------------------+
```

Or

```
{"level":"error","ts":"2025-06-20T19:55:30.809+0530","caller":"log/log.go:102","msg":"nvidia gpu count mismatch (lspci 8, expected 10)","count_lspci":8,"count_nvml":8,"expected":10,}
✘ nvidia gpu count mismatch (lspci 8, expected 10)
+--------------+-----------------------+
| Product Name | NVIDIA-H100-80GB-HBM3 |
| Count lspci  |           8           |
|  Count NVML  |           8           |
+--------------+-----------------------+
```

Close https://github.com/leptonai/gpud/issues/591.

### DIFF
--- a/components/accelerator/nvidia/gpu-counts/component.go
+++ b/components/accelerator/nvidia/gpu-counts/component.go
@@ -28,6 +28,8 @@ const (
 	EventNameMisMatch = "gpu-count-mismatch"
 )
 
+const defaultLookbackPeriod = 3 * 24 * time.Hour
+
 var _ components.Component = &component{}
 
 type component struct {
@@ -48,8 +50,6 @@ type component struct {
 	lastMu          sync.RWMutex
 	lastCheckResult *checkResult
 }
-
-const defaultLookbackPeriod = 3 * 24 * time.Hour
 
 func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 	cctx, ccancel := context.WithCancel(gpudInstance.RootCtx)
@@ -231,7 +231,6 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = "error getting reboot events"
 		log.Logger.Warnw(cr.reason, "error", cr.err)
 		return cr
 	}
@@ -239,7 +238,6 @@ func (c *component) Check() components.CheckResult {
 	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeUnhealthy
-		cr.reason = "error getting gpu count mismatch events"
 		log.Logger.Warnw(cr.reason, "error", cr.err)
 		return cr
 	}
--- a/components/accelerator/nvidia/gpu-counts/component_test.go
+++ b/components/accelerator/nvidia/gpu-counts/component_test.go
@@ -1585,7 +1585,7 @@ func TestComponent_Check_SuggestedActions_RebootEventsError(t *testing.T) {
 	require.True(t, ok)
 
 	assert.Equal(t, apiv1.HealthStateTypeUnhealthy, cr.health)
-	assert.Equal(t, "error getting reboot events", cr.reason)
+	assert.Equal(t, "nvidia gpu count mismatch (found 1, expected 2)", cr.reason)
 	assert.Equal(t, testErr, cr.err)
 	assert.Nil(t, cr.suggestedActions)
 }
@@ -1641,7 +1641,7 @@ func TestComponent_Check_SuggestedActions_GPUMismatchEventsError(t *testing.T) {
 	require.True(t, ok)
 
 	assert.Equal(t, apiv1.HealthStateTypeUnhealthy, cr.health)
-	assert.Equal(t, "error getting gpu count mismatch events", cr.reason)
+	assert.Equal(t, "nvidia gpu count mismatch (found 1, expected 2)", cr.reason)
 	assert.Equal(t, testErr, cr.err)
 	assert.Nil(t, cr.suggestedActions)
 }


components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func (c *component) Check() components.CheckResult {
+	log.Logger.Infow("checking nvidia gpu counts")
+
+	cr := &checkResult{
+		ts: time.Now().UTC(),
+	}
+
+	thresholds := c.getThresholdsFunc()
```
eahydra: it's better that check if threshold is zero first.

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,290 @@
...
+
+	if c.getCountLspci != nil {
+		cr.CountLspci, cr.err = c.getCountLspci(c.ctx)
+		if cr.err != nil {
+			cr.health = apiv1.HealthStateTypeUnhealthy
+			cr.reason = "error getting count of lspci"
+			log.Logger.Errorw(cr.reason, "error", cr.err)
+			return cr
+		}
+	}
```
eahydra: Is it necessary for this component to depend on lspci?

eahydra (commented):
overall lgtm

components/accelerator/nvidia/gpu-counts/component.go
gyuho: done

components/accelerator/nvidia/gpu-counts/component.go
gyuho: Yes, let's keep them both for now. When we know this is stable, we can drop one (lspci).

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,287 @@
...
+		Name:             Name,
+		Health:           cr.health,
+		Reason:           cr.reason,
+		Error:            cr.getError(),
+		SuggestedActions: cr.getSuggestedActions(),
+		ExtraInfo: map[string]string{
+			"product_name": cr.ProductName,
+			"count_lspci":  fmt.Sprintf("%d", cr.CountLspci),
+			"count_nvml":   fmt.Sprintf("%d", cr.CountNVML),
+		},
```
eahydra: BTW,  we should remove them.

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	state := apiv1.HealthState{
+		Time:             metav1.NewTime(cr.ts),
+		Component:        Name,
+		Name:             Name,
+		Health:           cr.health,
+		Reason:           cr.reason,
+		Error:            cr.getError(),
+		SuggestedActions: cr.getSuggestedActions(),
```
eahydra, line 514: What's the suggested action? In some cases, the GPU may be lost and can be recovered by rebooting. In some case, it cannot be recovered even after rebooting.. It is better to suggest restarting N times first, and then suggest hardware inspection

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	if cr.CountNVML == thresholds.Count {
+		cr.health = apiv1.HealthStateTypeHealthy
+		cr.reason = fmt.Sprintf("nvidia gpu count matching thresholds (%d)", thresholds.Count)
+		return cr
+	}
+
+	cr.health = apiv1.HealthStateTypeUnhealthy
+	cr.reason = fmt.Sprintf("nvidia gpu count mismatch (nvml %d, expected %d)", cr.CountNVML, thresholds.Count)
```
eahydra:
```suggestion
	cr.reason = fmt.Sprintf("nvidia gpu count mismatch (found %d, expected %d)", cr.CountNVML, thresholds.Count)
```

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,423 @@
...
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = "error getting gpu mismatch events"
+		log.Logger.Warnw(cr.reason, "error", cr.err)
+		return cr
+	}
+
+	// check
+	// case 2. gpu mismatch -> reboot -> gpu mismatch; suggest hw inspection
+	// we just inserted above, safe to assume reboot (if ever) happened before gpu mismatch
+	firstRebootTime := rebootEvents[0].Time
```
eahydra: is it the most recent?

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,423 @@
...
+	for _, mismatchEvent := range gpuMismatchEvents {
```
eahydra: Is the first one the most recent?

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,423 @@
...
+	if err != nil {
+		cr.err = err
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = "error getting gpu mismatch events"
+		log.Logger.Warnw(cr.reason, "error", cr.err)
+		return cr
+	}
+
+	// check
+	// case 2. gpu mismatch -> reboot -> gpu mismatch; suggest hw inspection
```
eahydra: we should support rebooting twice.

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,423 @@
...
+	getThresholdsFunc func() ExpectedGPUCounts
+
+	// lookback period to query the past reboot + gpu count mismatch events
+	lookbackPeriod time.Duration
+
+	lastMu          sync.RWMutex
+	lastCheckResult *checkResult
+}
+
+const defaultLookbackPeriod = time.Hour
```
eahydra: It's too short. It should be consistent with the strategy of XID and XSXID components and support at least 3 days.

components/accelerator/nvidia/gpu-counts/component.go
gyuho: Yes, the event store returns in reverse order of timestamp (latest first)

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,519 @@
...
+	getThresholdsFunc func() ExpectedGPUCounts
+
+	// lookback period to query the past reboot + gpu count mismatch events
+	lookbackPeriod time.Duration
+
+	lastMu          sync.RWMutex
+	lastCheckResult *checkResult
+}
+
+const defaultLookbackPeriod = 3 * 24 * time.Hour
```
eahydra: shall we move it after line 28

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,519 @@
...
+	if err := c.recordMismatchEvent(cr); err != nil {
+		return cr
+	}
+
+	// look up past events to derive the suggested actions
+	rebootEvents, err := c.rebootEventStore.GetRebootEvents(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+	if err != nil {
+		cr.err = err
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = "error getting reboot events"
```
eahydra: IMO we should not override the reason.

components/accelerator/nvidia/gpu-counts/component.go
```diff
@@ -0,0 +1,519 @@
...
+		log.Logger.Warnw(cr.reason, "error", cr.err)
+		return cr
+	}
+	gpuMismatchEvents, err := c.eventBucket.Get(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+	if err != nil {
+		cr.err = err
+		cr.health = apiv1.HealthStateTypeUnhealthy
+		cr.reason = "error getting gpu count mismatch events"
```
eahydra: same as above

eahydra (approved):
overall lgtm. We should do more testing in dev.

PR #951, opened 2025-06-24
fix(reboot): select only "reboot" events in "os" bucket

gyuho:
And support queries with other buckets.

And help with https://github.com/leptonai/gpud/pull/944 to support taking reboot events into account.

### DIFF
--- a/pkg/host/event.go
+++ b/pkg/host/event.go
@@ -27,7 +27,7 @@ type RebootEventStore interface {
 	// it will also query the events from the extra buckets, with the same since time.
 	// The returned events do NOT include other events from the "os" component (e.g., kmsg watcher).
 	// The returned events are in the descending order of timestamp (latest event first).
-	GetRebootEvents(ctx context.Context, since time.Time, otherBuckets ...eventstore.Bucket) (eventstore.Events, error)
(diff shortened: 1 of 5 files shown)


pkg/host/event.go
```diff
@@ -3,18 +3,31 @@ package host
...
+	// RecordReboot records a reboot event, with the event name [EventNameReboot],
+	// in the bucket [EventBucketName].
 	RecordReboot(ctx context.Context) error
-	GetRebootEvents(ctx context.Context, since time.Time) (eventstore.Events, error)
+
+	// GetRebootEvents queries all "reboot" events and if any extra buckets are provided,
+	// it will also query the events from the extra buckets, with the same since time.
+	// The returned events do NOT include other events from the "os" component (e.g., kmsg watcher).
+	// The returned events are in the descending order of timestamp (latest event first).
+	GetRebootEvents(ctx context.Context, since time.Time, otherBuckets ...eventstore.Bucket) (eventstore.Events, error)
```
eahydra: The method signature is not reasonable since the caller does not know what the otherBuckets are.

gyuho: removed

cardyok (approved):

PR #956, opened 2025-06-30
fix(cmd/status): do not parse login timestamp if empty

gyuho:


cmd/gpud/status/utils.go
```diff
@@ -0,0 +1,27 @@
...
+		ts, err := strconv.ParseInt(loginSuccess, 10, 64)
+		if err != nil {
+			return fmt.Errorf("failed to parse login success: %w", err)
+		}
+		loginTimeUTC := time.Unix(ts, 0)
+		nowUTC := time.Now().UTC()
+		loginTimeHumanized := humanize.RelTime(loginTimeUTC, nowUTC, "ago", "from now")
+		fmt.Printf("%s login success at %s (machine id: %s)\n", cmdcommon.CheckMark, loginTimeHumanized, machineID)
+	} else {
+		fmt.Printf("%s login information not found\n", cmdcommon.CheckMark)
```
eahydra: it's better to do nothing if there is no login info.

gyuho: done

PR #957, opened 2025-06-30
feat(nfs-checker): improve successful check message

gyuho:

### DIFF
--- a/pkg/nfs-checker/checker.go
+++ b/pkg/nfs-checker/checker.go
@@ -126,7 +126,7 @@ func (c *checker) Check() CheckResult {
 		return result
 	}
 
-	result.Message = fmt.Sprintf("successfully checked directory %q with %d files", c.cfg.VolumePath, len(matches))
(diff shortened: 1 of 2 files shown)


eahydra:
https://github.com/leptonai/gpud/pull/957/files#diff-1bf81447673432086f1501cbc9fdd9046e6906908ffff3d42c179cebd18025e1R103-R121

Regarding these lines, it may still fail if there are files with different contents.. IMO, it should check if the content of the corresponding file of the current machine is the same as the expected content.

gyuho:
> should check if the content of the corresponding file of the current machine is the same as the expected content.

You mean, only check the current machine file contents? It already does. Regardless of local machine or any other peer reads, we return an error if the file contents are different. Currently, no versioned check is supported for file contents. Assume, it will be rather static.

gyuho:
@eahydra Closing in favor of https://github.com/leptonai/gpud/pull/963.

PR #963, opened 2025-07-02
fix(nfs-checker): add missing checker.Clean, simplify the logic to only check the self-machine

gyuho:
This fixes the issue where the NFS checker would not clean up old files
after a successful check.


components/nfs/component.go
```diff
@@ -181,6 +181,15 @@ func (c *component) Check() components.CheckResult {
 			return cr
 		}
 
+		// safe to clean/delete old files
+		if err := checker.Clean(); err != nil {
```
eahydra, line 184: the clean function delete all files, how do you guarantee if the deletion operation does not affect other gpud health check?

eahydra: checker.Write just writes a file with the current machine id, but checker.Clean deletes files in the target dir... it does not make sense

gyuho: https://github.com/leptonai/gpud/blob/efe31602804c5137d9e72191ebf4c6979dec5481/pkg/nfs-checker/checker.go#L133-L145 checks the file creation time

And the TTL is set by the control plane for our case. We can set >10-min, so that we only delete files that are >10 minutes old. And if there's no write in the last 10-minute, other peers won't be able to read the disconnected nodes (which is expected, as a health checker).

eahydra: the nfs volume is shared for multiple GPUd instances.. all these GPUd instances are trying to delete these files, it's not a good practice.

gyuho: We can launch a pruning job, with the interval of TTLToDelete, with some jitter, and just log errors (if any), in warn level. This is required to prevent leaky files for this component.

gyuho: Discussed offline, will make this much simpler.

pkg/nfs-checker/checker.go
```diff
@@ -83,64 +73,32 @@ func (c *checker) Write() error {
...
-
-	// remove all files that are older than the TTL
-	// in order to make sure the checker only checks
-	// the latest file writes/reads from other members in the group
-	now := time.Now().UTC()
-	before := now.Add(-c.cfg.TTLToDelete.Duration)
-
-	return pkgfilecleaner.Clean(pattern, before)
+	file := c.cfg.fileSelf()
+	return os.RemoveAll(file)
```
eahydra, line 103: we still need to check if the return error is equal to `NotExist`, if so, the deletion is success.

gyuho: @eahydra https://pkg.go.dev/os#RemoveAll

> If the path does not exist, RemoveAll returns nil (no error).

eahydra (approved):

Issue #375, opened 2025-02-08
rate limit events ingestion


xiang90:
we need to protect gpud from receiving too many events from the checked component. 

as a general rule, we should:
1. rate limit at minute granularity. less than 15 events per minutes. 
2. dedup similar events in the past 1 minute. this is low priority.

PR #964, opened 2025-07-02
feat(eventstore): support ingest rate limit (no limit for now)

gyuho:
Address https://github.com/leptonai/gpud/issues/375. And we can use this for rate-limiting the system reboot as well.


eahydra (commented):
This PR is not yet completed. Who will call WithIngestRateLimit to set RateLimiter? I do not recommend directly modifying a currently stable package(eventstore). 

If you want to support current limiting, make a layer of encapsulation outside, and use it separately if you need to use the encapsulation. This ensures that this change will not affect so much code.

In addition, have you analyzed how many events we currently have? Currently, only IB Flapping and GPU HW Slowdown should generate a large number of events. You are improving IB Flapping. After the modification, theoretically there will not be many events. The same logic can be applied to HW Slowdown.

gyuho:
Let's revisit this

PR #966, opened 2025-07-03
bus id pci id

gyuho:
- **1**
- **feat(nvidia): new device interface to get PCI Bus ID**


components/accelerator/nvidia/clock-speed/component.go
```diff
@@ -183,10 +183,10 @@ func (cr *checkResult) String() string {
 	table := tablewriter.NewWriter(buf)
 	table.SetAlignment(tablewriter.ALIGN_CENTER)
 
-	table.SetHeader([]string{"GPU UUID", "Graphics MHz", "Memory MHz", "Graphics Supported", "Memory Supported"})
```
eahydra, line 186: Why remove the UUID? We still need it.

gyuho: @eahydra Now added back

PR #968, opened 2025-07-04
[LEP-1173] perf(eventstore/reboot): support event name column filter, optimize reboot events store reads

gyuho:
- Don't select all events + if-statement to filter/select reboots ("os" bucket"), instead, just`select WHERE...`
- Don't create the SQL table for every single reboot event insert/reads (very inefficient!)

Combining all these, reboot events store read performance 3x.

And we will need the same eventstore helper for the future Infiniband flap detection.

```
                                  │ baseline.txt │             new.txt             │
                                  │    sec/op    │    sec/op     vs base           │
EventStore_GetEvents_Filtering-10   3.752m ± ∞ ¹   2.787m ± ∞ ¹  ~ (p=1.000 n=1) ²

                                  │ baseline.txt │             new.txt             │
                                  │    sec/op    │    sec/op     vs base           │
EventStore_GetEvents_Filtering-10   3.542m ± ∞ ¹   2.762m ± ∞ ¹  ~ (p=1.000 n=1) ²

                                  │ baseline.txt  │             new.txt              │
                                  │     B/op      │     B/op       vs base           │
EventStore_GetEvents_Filtering-10   3.674Mi ± ∞ ¹   2.375Mi ± ∞ ¹  ~ (p=1.000 n=1) ²

                                  │ baseline.txt │             new.txt             │
                                  │  allocs/op   │  allocs/op    vs base           │
EventStore_GetEvents_Filtering-10   62.52k ± ∞ ¹   50.03k ± ∞ ¹  ~ (p=1.000 n=1) ²
```

```
                                        │ /tmp/baseline1.txt │          /tmp/new1.txt          │
                                        │       sec/op       │    sec/op     vs base           │
RebootsStore_Record_NoExistingEvents-10         32.05µ ± ∞ ¹   10.99µ ± ∞ ¹  ~ (p=1.000 n=1) ²

                                        │ /tmp/baseline1.txt │          /tmp/new1.txt           │
                                        │        B/op        │     B/op       vs base           │
RebootsStore_Record_NoExistingEvents-10        6.252Ki ± ∞ ¹   2.846Ki ± ∞ ¹  ~ (p=1.000 n=1) ²

                                        │ /tmp/baseline1.txt │         /tmp/new1.txt          │
                                        │     allocs/op      │  allocs/op   vs base           │
RebootsStore_Record_NoExistingEvents-10         160.00 ± ∞ ¹   73.00 ± ∞ ¹  ~ (p=1.000 n=1) ²
```

```
                              │ /tmp/baseline2.txt │          /tmp/new2.txt          │
                              │       sec/op       │    sec/op     vs base           │
RebootsStore_Get_SmallData-10        4980.5µ ± ∞ ¹   812.7µ ± ∞ ¹  ~ (p=1.000 n=1) ²

                              │ /tmp/baseline2.txt │          /tmp/new2.txt           │
                              │        B/op        │     B/op       vs base           │
RebootsStore_Get_SmallData-10       4060.2Ki ± ∞ ¹   483.3Ki ± ∞ ¹  ~ (p=1.000 n=1) ²

                              │ /tmp/baseline2.txt │          /tmp/new2.txt          │
                              │     allocs/op      │  allocs/op    vs base           │
RebootsStore_Get_SmallData-10         66.16k ± ∞ ¹   11.06k ± ∞ ¹  ~ (p=1.000 n=1) ²
```

```
                              │ /tmp/baseline3.txt │          /tmp/new3.txt          │
                              │       sec/op       │    sec/op     vs base           │
RebootsStore_Get_LargeData-10        25.103m ± ∞ ¹   4.088m ± ∞ ¹  ~ (p=1.000 n=1) ²

                              │ /tmp/baseline3.txt │          /tmp/new3.txt           │
                              │        B/op        │     B/op       vs base           │
RebootsStore_Get_LargeData-10       23.098Mi ± ∞ ¹   2.903Mi ± ∞ ¹  ~ (p=1.000 n=1) ²

                              │ /tmp/baseline3.txt │          /tmp/new3.txt          │
                              │     allocs/op      │  allocs/op    vs base           │
RebootsStore_Get_LargeData-10        330.17k ± ∞ ¹   55.07k ± ∞ ¹  ~ (p=1.000 n=1) ²
```

```
                                      │ /tmp/baseline4.txt │          /tmp/new4.txt          │
                                      │       sec/op       │    sec/op     vs base           │
RebootsStore_Record_ExistingEvents-10         30.79µ ± ∞ ¹   10.90µ ± ∞ ¹  ~ (p=1.000 n=1) ²

                                      │ /tmp/baseline4.txt │          /tmp/new4.txt           │
                                      │        B/op        │     B/op       vs base           │
RebootsStore_Record_ExistingEvents-10        6.253Ki ± ∞ ¹   2.846Ki ± ∞ ¹  ~ (p=1.000 n=1) ²

                                      │ /tmp/baseline4.txt │         /tmp/new4.txt          │
                                      │     allocs/op      │  allocs/op   vs base           │
RebootsStore_Record_ExistingEvents-10         160.00 ± ∞ ¹   73.00 ± ∞ ¹  ~ (p=1.000 n=1) ²
```

c.f., https://github.com/leptonai/gpud/pull/967


eahydra:
This is a case with more reads than writes. From your benchmark results, there is actually no essential change from baseline to new. They are all very slow because they are still SQL operations. Should we build a cache directly based on the latest DefaultRetentionPeriod event? Add a goroutine to refresh the cache periodically in the background. Is this more efficient?

gyuho:
@eahydra We can do that as well, in a separate PR? But we also need to remove `CREATE TABLE` runs on every single insert.

gyuho:
Since reboot bucket/table will remain pretty small, this is not the critical bottleneck, this is to mainly remove unnecessary filtering. (you don't need if-statement, when you can just use `WHERE` statement).

components/os/component.go
```diff
@@ -173,55 +169,15 @@ func (c *component) LastHealthStates() apiv1.HealthStates {
...
-			return nil, err
-		}
-		if len(rebootEvents) > 0 {
-			events = append(events, rebootEvents.Events()...)
-		}
-	}
-
-	if len(events) == 0 {
-		return nil, nil
+	evs, err := c.eventBucket.Get(ctx, since)
```
gyuho, line 176: @eahydra @cardyok Since os + reboot package share the same bucket, we don't need these two get calls... no idea why I added this duplicate code :/

eahydra:
> @eahydra We can do that as well, in a separate PR? But we also need to remove `CREATE TABLE` runs on every single insert.

If we support cache, we don't need to do this PR.

PR #972, opened 2025-07-06
perf(pkg/kmsg): drop kmsg messages in channel sends if channel is full

gyuho:

### DIFF
--- a/pkg/kmsg/watcher.go
+++ b/pkg/kmsg/watcher.go
@@ -280,7 +280,17 @@ func readFollow(kmsgFile *os.File, bootTime time.Time, msgs chan<- Message, dedu
 			}
 		}
 
-		msgs <- *msg
+		// even with channel buffer + deduplication
+		// it is still possible that the channel is full (flooding kmsg messages)
+		// if the database operations are slow on the receiver side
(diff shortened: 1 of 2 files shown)


pkg/kmsg/watcher.go
```diff
@@ -280,7 +280,17 @@ func readFollow(kmsgFile *os.File, bootTime time.Time, msgs chan<- Message, dedu
...
+		// if the database operations are slow on the receiver side
+		// rather than blocking the send here, we should just drop
+		// otherwise, go scheduler spins exhausting CPU cores with EpollWait syscalls
+		// ref. https://github.com/leptonai/gpud/pull/967
+		select {
+		case msgs <- *msg:
+		case <-time.After(time.Second):
+			log.Logger.Warnw("kmsg channel is full, dropping message", "timestamp", msg.Timestamp, "message", msg.Message)
+		}
 	}
```
eahydra, line 294: First, I don't think we should drop the message if the channel is full, since we may drop a critical message; Second, could you give more details about the EpollWait consume high cpu?   This is the most common practice in golang and I don't think it is the root cause.

PR #980, opened 2025-07-08
feat(memory): make oom events human readable using cadvisor code

gyuho:
ref.
https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser_test.go
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/oom/oom_watcher_linux.go

### DIFF
--- a/components/memory/kmsg_matcher.go
+++ b/components/memory/kmsg_matcher.go
@@ -1,123 +1,205 @@
+// Copyright 2014 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ref. https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser.go
+
 package memory
 
 import (
+	"fmt"
+	"path"
 	"regexp"
+	"strconv"
 )
 
-const (
-	// e.g.,
-	// Out of memory: Killed process 123, UID 48, (httpd).
-	// [...] Out of memory: Killed process 123, UID 48, (httpd).
-	//
-	// NOTE: this is often followed by a line like:
-	// [Sun Dec  8 09:23:39 2024] oom_reaper: reaped process 345646 (vector), now anon-rss:0kB, file-rss:0kB, shmem-rss:0
-	// (to reap the memory used by the OOM victim)
-	eventOOM   = "memory_oom"
-	regexOOM   = `Out of memory:`
-	messageOOM = `oom detected`
+const OOMEventName = "OOM"
 
+func createMatchFunc() func(line string) (eventName string, message string) {
+	// below match function is called for each incoming kmsg message
+	// so we track the progress outside of the function
+	readingOOMMessages := false
+	var oomCurrentInstance *OOMInstance
+
+	// this match function is to be run in sequence for each incoming message
+	return func(line string) (eventName string, message string) {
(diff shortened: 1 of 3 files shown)


components/memory/kmsg_matcher.go
```diff
@@ -1,123 +1,203 @@
...
+			// If there's an error, reset and skip
+			readingOOMMessages = false
+			oomCurrentInstance = nil
+			return "", ""
+		}
+
+		// Modern kernel format: containerRegexp extracts all info in one match
+		if containerFound && oomCurrentInstance.Pid != 0 {
+			// Generate the event message
+			eventName = "oom"
```
eahydra:
```suggestion
			eventName = "OOM"
```

components/memory/kmsg_matcher.go
```diff
@@ -1,123 +1,203 @@
...
+				// If there's an error, reset and skip
+				readingOOMMessages = false
+				oomCurrentInstance = nil
+				return "", ""
+			}
+
+			// If we found process info, the OOM event is complete
+			if processFound {
+				// Generate the event message
+				eventName = "oom"
```
eahydra:
```suggestion
				eventName = "OOM"
```

components/memory/kmsg_matcher.go
```diff
@@ -1,123 +1,203 @@
...
-	return false
 }
 
-func HasOOMKillConstraint(line string) bool {
-	if match := compiledOOMKillConstraint.FindStringSubmatch(line); match != nil {
-		return true
-	}
-	return false
+// struct that contains information related to an OOM kill instance
+type OomInstance struct {
```
eahydra:
```suggestion
type OOMInstance struct {
```

components/memory/kmsg_matcher.go
```diff
@@ -1,123 +1,203 @@
...
-		return true
+func (o *OomInstance) Summary() string {
+	if o == nil {
+		return ""
 	}
-	return false
+	eventMsg := "System OOM encountered"
+	if o.ProcessName != "" && o.Pid != 0 {
+		eventMsg = fmt.Sprintf("%s, victim process: %s, pid: %d", eventMsg, o.ProcessName, o.Pid)
+	}
```
eahydra: Can we determine if OOM is at the container level or at the pod level?

components/memory/kmsg_matcher_test.go
```diff
@@ -1,305 +1,843 @@
...
+			expectedEvents: []struct {
+				eventName string
+				message   string
+			}{
+				{
+					eventName: "oom",
+					message:   "System OOM encountered, victim process: evil-program2, pid: 19636",
+				},
+			},
+		},
```
eahydra: it seems that the OOM is at cgroup level?

components/memory/kmsg_matcher.go
gyuho: Even kubelet knows only at the container level https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/oom/oom_watcher_linux.go (unless we have pod to containers mapping), something we can do at control plane side?

components/memory/kmsg_matcher_test.go
gyuho: Yes, cgroup level

eahydra: but the expected message is System OOM

gyuho: updated to `OOM`

components/memory/kmsg_matcher.go
```diff
@@ -1,123 +1,205 @@
...
 }
 
-func HasOOMKiller(line string) bool {
-	if match := compiledOOMKiller.FindStringSubmatch(line); match != nil {
-		return true
+func (o *OOMInstance) Summary() string {
+	if o == nil {
+		return ""
+	}
+	eventMsg := "OOM encountered"
```
eahydra, line 130: https://github.com/kubernetes/kubernetes/blob/9e204ba03c2c95fd20dd18fcc6c40c3b16576592/pkg/kubelet/oom/oom_watcher_linux.go#L84

kubelet checks if the VictimContainerName is `/`, if so, it's system OOM.. IMO we should follow the same way, and we should display the `System OOM encountered`, accurately displaying System OOM information helps us troubleshoot problems

PR #981, opened 2025-07-09
feat(containerd): extend dangling pod count check in containerd

cardyok:


components/containerd/component.go
```diff
@@ -171,10 +180,19 @@ func (c *component) Check() components.CheckResult {
 			}
 			return cr
 		}
+		danglingCount := cr.danglingPodCount(c.exportResult)
+		if danglingCount > DanglingThreshold {
+			cr.health = apiv1.HealthStateTypeUnhealthy
+			cr.reason = fmt.Sprintf("node has  %v dangling pods, threshold %v", danglingCount, DanglingThreshold)
```
gyuho:
```suggestion
			cr.reason = fmt.Sprintf("node has %v dangling pods, threshold %v", danglingCount, DanglingThreshold)
```

components/containerd/component.go
```diff
@@ -171,10 +180,19 @@ func (c *component) Check() components.CheckResult {
...
+		danglingCount := cr.danglingPodCount(c.exportResult)
+		if danglingCount > DanglingThreshold {
+			cr.health = apiv1.HealthStateTypeUnhealthy
+			cr.reason = fmt.Sprintf("node has %v dangling pods, threshold %v", danglingCount, DanglingThreshold)
+			cr.suggestedAction = &apiv1.SuggestedActions{
+				Description:   "too many dangling pod",
+				RepairActions: []apiv1.RepairActionType{apiv1.RepairActionTypeRebootSystem},
+			}
+			return cr
+		} else if danglingCount != 0 {
```
eahydra, line 204: shall we mark node degraded if the danglingCount is greater than 5 and less than threshold?

components/containerd/component.go
```diff
@@ -277,3 +297,28 @@ func (cr *checkResult) HealthStates() apiv1.HealthStates {
...
 	return apiv1.HealthStates{state}
 }
+
+func (cr *checkResult) danglingPodCount(kubeletPods *kubelet.ExportPods) int {
+	var danglingCount int
+	if kubeletPods == nil {
+		return danglingCount
+	}
+	podMap := make(map[string]struct{})
+	kubeletPods.Mu.RLock()
```
eahydra: it's better not to do this.. just return a copy of kubeletPods.

components/containerd/component.go
```diff
@@ -277,3 +297,28 @@ func (cr *checkResult) HealthStates() apiv1.HealthStates {
 	}
 	return apiv1.HealthStates{state}
 }
+
+func (cr *checkResult) danglingPodCount(kubeletPods *kubelet.ExportPods) int {
```
eahydra: it's better to make the function as a freedom function, it's not a part of checkResult

pkg/server/server.go
```diff
@@ -205,6 +206,10 @@ func New(ctx context.Context, auditLogger log.AuditLogger, config *lepconfig.Con
 
 		EventStore:       eventStore,
 		RebootEventStore: rebootEventStore,
+		KubeletPods: &kubelet.ExportPods{
```
eahydra: it's better to setup a function...

eahydra (commented):
lgtm, defer to @gyuho

PR #984, opened 2025-07-11
fix(infiniband): improve reason message when ib flap/drop events are present

gyuho:
> ok; no infiniband port issue; mlx5_1 port 1 down since 2025-07-10T15:16:35Z, mlx5_2 port 1 down since 2025-07-10T15:16:35Z, mlx5_7 port 1 down since 2025-07-10T15:16:35Z, mlx5_8 port 1 down since 2025-07-10T15:16:35Z

should be

> mlx5_1 port 1 down since 2025-07-10T15:16:35Z, mlx5_2 port 1 down since 2025-07-10T15:16:35Z, mlx5_7 port 1 down since 2025-07-10T15:16:35Z, mlx5_8 port 1 down since 2025-07-10T15:16:35Z


gyuho:
@eahydra Then the user/customer wouldn't know "since when" a specific port is down. The `event_reason` column in our db stores that information. The goal is to show when a specific port became "drop" state

eahydra:
> @eahydra Then the user/customer wouldn't know "since when" a specific port is down. The `event_reason` column in our db stores that information. The goal is to show when a specific port became "drop" state

Can we just write the message to event, not in the state?  I want to make the state as simple as possible. If possible, simplify the message in state.

components/accelerator/nvidia/infiniband/component_read_class_test.go
```diff
@@ -1309,7 +1308,7 @@ func TestComponentReadClass_EventReporting(t *testing.T) {
 
 		// Should only include the known event type (drop), not the unknown one
 		summary := cr.Summary()
-		assert.Contains(t, summary, "mlx5_1 port 1 drop since")
+		assert.Contains(t, summary, "port drop mlx5_1")
```
eahydra: is it better that `port dropped of mlx5_1`?  or shall we make it easier to read or understand?  include all other messages in the component.

eahydra (approved):

PR #985, opened 2025-07-13
feat(xid, row remapping): override suggested actions in case of ongoing hw inspection

gyuho:
**feat(xid, row remapping): override suggested actions in case of ongoing hw inspection**


gyuho:
To be replaced by https://github.com/leptonai/gpud/pull/987

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/987

PR #986, opened 2025-07-14
fix(pkg/disk): set timeout for os.Stat in case of malfunctioning NFS

gyuho:


eahydra:
@cardyok PTAL. Thanks

pkg/disk/disk_partition_usage.go
```diff
@@ -162,3 +171,32 @@ type Usage struct {
...
+		case <-ctx.Done():
+			// Context cancelled, result will be discarded
+		}
+	}()
+
+	select {
+	case res := <-resultCh:
+		return res.info, res.err
+	case <-ctx.Done():
+		return nil, fmt.Errorf("stat %s: operation timed out after %v", path, timeout)
```
cardyok: can introduce a error type for this

cardyok (approved):

PR #987, opened 2025-07-14
fix(nvidia/xid): skip Xid 63/64 in favor of row remapping checks

gyuho:


components/accelerator/nvidia/xid/component.go
```diff
@@ -247,6 +247,22 @@ func (c *component) Check() components.CheckResult {
...
+		// can also be detected by NVML API (vs. kmsg scanning)
+		// thus we discard Xid 63/64 in favor of row remapping checks
+		// especially, NVIDIA row remapping pending can happen >3 times
+		// which warrants system reboots, before reaching its row remapping
+		// failures threshold which requires hardware inspection
+		// in other words, we do not want to blindly suggest hw inspection
+		// from Xid 63/64, while row remmaping pending may self-resolve
+		// after >3 times of system reboots
+		// this is why we here discard Xid 63/64 in favor of row remapping checks
+		if xidErr.Xid == 63 || xidErr.Xid == 64 {
```
eahydra: Maybe it's better that we can consider if there is row-remapping capability in the machine. if so, we can ignore the 63 and 64, otherwise, we can keep them.

gyuho: @eahydra Done

PR #988, opened 2025-07-15
fix(nfs): make disk stat/read checks on NFS more reliable with timeouts

gyuho:
```
goroutine 444 [syscall, 73 minutes]:
syscall.Syscall6(0x106, 0xffffffffffffff9c, 0xc001dd2140, 0xc001de6448, 0x0, 0x0, 0x0)
	/opt/hostedtoolcache/go/1.24.2/x64/src/syscall/syscall_linux.go:95 +0x39
syscall.fstatat(0xffffffffffffff9c, {0xc0019a4a40?, 0x0?}, 0xc001de6448, 0x0)
	/opt/hostedtoolcache/go/1.24.2/x64/src/syscall/zsyscall_linux_amd64.go:1437 +0x96
syscall.Stat(...)
	/opt/hostedtoolcache/go/1.24.2/x64/src/syscall/syscall_linux_amd64.go:61
os.statNolog.func1(...)
	/opt/hostedtoolcache/go/1.24.2/x64/src/os/stat_unix.go:32
os.ignoringEINTR(...)
	/opt/hostedtoolcache/go/1.24.2/x64/src/os/file_posix.go:251
os.statNolog({0xc0019a4a40, 0xe})
	/opt/hostedtoolcache/go/1.24.2/x64/src/os/stat_unix.go:31 +0x52
os.Stat({0xc0019a4a40, 0xe})
	/opt/hostedtoolcache/go/1.24.2/x64/src/os/stat.go:13 +0x2c
github.com/leptonai/gpud/pkg/nfs-checker.(*Config).ValidateAndMkdir(0xc0016e9b28)
	/home/runner/work/gpud/gpud/pkg/nfs-checker/group_config.go:75 +0x3d
github.com/leptonai/gpud/pkg/nfs-checker.(*MemberConfig).Validate(...)
	/home/runner/work/gpud/gpud/pkg/nfs-checker/member_config.go:47
github.com/leptonai/gpud/pkg/nfs-checker.MemberConfigs.Validate({0xc001058980?, 0x6?, 0x6?})
	/home/runner/work/gpud/gpud/pkg/nfs-checker/member_config.go:36 +0xb1
github.com/leptonai/gpud/components/nfs.(*component).Check(0xc000e12540)
	/home/runner/work/gpud/gpud/components/nfs/component.go:148 +0x99e
github.com/leptonai/gpud/components/nfs.(*component).Start.func1()
	/home/runner/work/gpud/gpud/components/nfs/component.go:82 +0x79
created by github.com/leptonai/gpud/components/nfs.(*component).Start in goroutine 1
	/home/runner/work/gpud/gpud/components/nfs/component.go:71 +0x4f
```


pkg/nfs-checker/checker.go
```diff
@@ -34,11 +37,15 @@ type CheckResult struct {
 	// Set to an empty string, if there was no error, and
 	// validation succeeded.
 	Error string `json:"error,omitempty"`
+
+	// TimeoutError indicates that if the [CheckResult].Error is not empty,
+	// the error was due to operation timeout (e.g., [context.DeadlineExceeded]).
+	TimeoutError bool `json:"timeout_error,omitempty"`
```
cardyok, line 43: do we really need this bool..

cardyok (dismissed):

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+	timeoutCtx, cancel := context.WithTimeout(c.ctx, 5*time.Second)
+	err := c.validateMemberConfigs(timeoutCtx, memberConfigs)
+	cancel()
+	if err != nil {
 		cr.err = err
 		cr.health = apiv1.HealthStateTypeDegraded
-		cr.reason = "invalid nfs group configs"
+
+		if errors.Is(err, context.DeadlineExceeded) {
+			cr.reason = "NFS validation timed out - filesystem may be unresponsive"
```
eahydra:
```suggestion
			cr.reason = "NFS validation timed out - server may be unresponsive"
```

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+		} else {
+			cr.reason = "invalid nfs group configs"
```
eahydra, line 180: if it's invalid, shall we skip it?

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+		timeoutCtx, cancel = context.WithTimeout(c.ctx, 5*time.Second)
+		checker, err := c.newChecker(timeoutCtx, &memberConfig)
+		cancel()
 		if err != nil {
 			cr.err = err
 			cr.health = apiv1.HealthStateTypeDegraded
-			cr.reason = "failed to create nfs checker for " + memberConfig.VolumePath
+
+			if errors.Is(err, context.DeadlineExceeded) {
+				cr.reason = "NFS checker creation timed out for " + memberConfig.VolumePath + " - filesystem may be unresponsive"
```
eahydra:
```suggestion
				cr.reason = "NFS checker creation timed out for " + memberConfig.VolumePath + " - server may be unresponsive"
```

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+		// Write with timeout
+		timeoutCtx, cancel = context.WithTimeout(c.ctx, 5*time.Second)
+		err = c.writeChecker(timeoutCtx, checker)
+		cancel()
+		if err != nil {
 			cr.err = err
 			cr.health = apiv1.HealthStateTypeDegraded
-			cr.reason = "failed to write to nfs checker for " + memberConfig.VolumePath
+			if errors.Is(err, context.DeadlineExceeded) {
+				cr.reason = "NFS write timed out for " + memberConfig.VolumePath + " - filesystem may be unresponsive"
```
eahydra:
```suggestion
				cr.reason = "NFS write timed out for " + memberConfig.VolumePath + " - server may be unresponsive"
```

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+		nfsResult := c.checkChecker(timeoutCtx, checker)
+		cancel()
 		if len(nfsResult.Error) > 0 {
 			cr.err = errors.New(nfsResult.Error)
 			cr.health = apiv1.HealthStateTypeDegraded
-			cr.reason = "failed to check nfs checker for " + memberConfig.VolumePath
-			log.Logger.Warnw(cr.reason, "error", err)
+
+			if nfsResult.TimeoutError {
+				cr.reason = "NFS check timed out for " + memberConfig.VolumePath + " - filesystem may be unresponsive"
```
eahydra:
```suggestion
				cr.reason = "NFS check timed out for " + memberConfig.VolumePath + " - server may be unresponsive"
```

pkg/file/timeout_ops.go
```diff
@@ -0,0 +1,94 @@
+package file
```
eahydra: it's better to rename the file to timeout_io.go? or event timeout_file?

pkg/file/timeout_ops.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	resultCh := make(chan result, 1)
+
+	go func() {
+		info, err := os.Stat(path)
+		select {
+		case resultCh <- result{info: info, err: err}:
+		case <-ctx.Done():
+		}
+	}()
```
eahydra: May it leak goroutines?

pkg/file/timeout_ops.go
```diff
@@ -0,0 +1,94 @@
...
+func MkdirAllWithTimeout(ctx context.Context, path string, perm os.FileMode) error {
+	errCh := make(chan error, 1)
+
+	go func() {
+		err := os.MkdirAll(path, perm)
+		select {
+		case errCh <- err:
+		case <-ctx.Done():
+		}
+	}()
```
eahydra: and here

pkg/nfs-checker/checker.go
eahydra: +1, there is Error, why we need the bool field?

gyuho: Otherwise, we need to manually check the `CheckResult.Error` sub-string. If we have this, it's easier to check

pkg/file/timeout_ops.go
gyuho: We always pass the root context to these functions, so, on server shutdown, we should exit this here

pkg/file/timeout_ops.go
gyuho: We pass the root context from the server here, so we can exit this goroutine in case of server shutdown

components/nfs/component.go
```diff
@@ -145,43 +167,75 @@ func (c *component) Check() components.CheckResult {
...
+		timeoutCtx, cancel = context.WithTimeout(c.ctx, 5*time.Second)
+		nfsResult := c.checkChecker(timeoutCtx, checker)
+		cancel()
 		if len(nfsResult.Error) > 0 {
 			cr.err = errors.New(nfsResult.Error)
 			cr.health = apiv1.HealthStateTypeDegraded
-			cr.reason = "failed to check nfs checker for " + memberConfig.VolumePath
-			log.Logger.Warnw(cr.reason, "error", err)
+
+			if nfsResult.TimeoutError {
```
gyuho, line 229: @eahydra @cardyok This is the use case for `TimeoutError bool`. Simpler to do this way. Otherwise, need sub-string match, which is hacky.

components/nfs/component.go
gyuho: Yes, we do skip here by returning the `cr` (so we don't proceed to the next steps)

pkg/nfs-checker/checker.go
eahydra: make sense

pkg/file/timeout_ops.go
eahydra: but if the server is not shutdown, we leaked goroutines, right?

gyuho: Oh we also pass the context with timeouts to these functions, e.g., `ctx, cancel := context.WithTimeout(rootCtx, 5-second)`

eahydra: although the context timeout, the goroutine is stuck in IO path, so that we leak it.

components/disk/component.go
```diff
@@ -244,13 +252,20 @@ func (c *component) Check() components.CheckResult {
 	}
 
 	for target := range c.mountPointsToTrackUsage {
-		if _, err := os.Stat(target); err != nil {
+		timeoutCtx, cancel := context.WithTimeout(c.ctx, 5*time.Second)
```
gyuho, line 255: @eahydra This is where we pass root context + timeouts, so we don't leak goroutines, in case of blocking stat operations.

pkg/file/timeout_ops.go
gyuho: Discussed offline. There are multiple ways to work around this. Will work on this separately

PR #990, opened 2025-07-15
[LEP-895] feat(disk): catch raid failures, read-only degraded fs, nvme path failure

gyuho:


components/disk/kmsg_matcher.go
```diff
@@ -10,10 +10,31 @@ const (
...
+	// e.g.,
+	// md/raid0md0: Disk failure on nvme0n1p1 detected, failing array.
+	eventRAIDArrayFailure   = "raid_array_failure"
+	regexRAIDArrayFailure   = `md/raid.*: Disk failure on .* detected, failing array`
+	messageRAIDArrayFailure = `RAID array has failed due to disk failure`
+
+	// e.g.,
+	// EXT4-fs (md0): Remounting filesystem read-only
+	eventFilesystemReadOnly   = "filesystem_read_only"
+	regexFilesystemReadOnly   = `EXT4-fs.*Remounting filesystem read-only`
```
eahydra: if the fs is not EXT4, the regex match will fail. IMO, we can just match `.*Remounting filesystem read-only`

gyuho: @eahydra Fixed

components/disk/component.go
```diff
@@ -340,6 +502,83 @@ func (c *component) Check() components.CheckResult {
...
+		recentEvents, err := c.eventBucket.Get(c.ctx, cr.ts.Add(-5*time.Minute))
+		if err != nil {
+			log.Logger.Warnw("failed to get recent events", "error", err)
+		} else {
+			// Check if we have any of the three critical disk failure events
+			var foundFailure bool
+			var failureEventName string
+			var failureMessage string
+
+			for _, ev := range recentEvents {
```
eahydra: This implementation will miss some failure events if eventRAIDArrayFailure, eventFilesystemReadOnly, and eventNVMePathFailure have occurred within the last 5 minutes.  Another solution would be to reference the xid component, receive kmsg events, and log all of these events into a database.

eahydra:
GPUd should generate health state with suggestion for these cases.

components/disk/component.go
gyuho: @eahydra *~That should be fine, since next Check call in the iteration will find those events. If we want to have a shared kmsg poller, it's more complicated and not every machine will enable xid component.~* Actually, we need a safer way to query all historical events and purge safely, for time series events. Currently only supported for infiniband store.

gyuho: I am going to increase the lookback period for recentEvents, and will create a new issue to handle this more fundamentally

eahydra: > @eahydra That should be fine, since next Check call in the iteration will find those events. If we want to have a shared kmsg poller, it's more complicated and not every machine will enable xid component.

I don't want to have a shared ksmg poller for now. Just recommended that refer to xid component, watch the kmsg, handle the kmsg event async.

gyuho: I created LEP-1175 to track this. If we have a safe way to purge all unimportant data in events store on SetHealthy call, we can safely query all historical events without some arbitrary lookback period (3-day)

gyuho:
@eahydra Done.

> Just recommended that refer to xid component, watch the kmsg, handle the kmsg event async.

This `disk` component already has its own kmsg watcher, so we get these stream of events almost real-time. PTAL.

eahydra:
> @eahydra Done.
> 
> > Just recommended that refer to xid component, watch the kmsg, handle the kmsg event async.
> 
> This `disk` component already has its own kmsg watcher, so we get these stream of events almost real-time. PTAL.

That also should be improved. Like the NoSpaceLeft event, there are so many events, which is not reasonable.

components/disk/component.go
```diff
@@ -184,6 +196,121 @@ func (c *component) Close() error {
...
 }
 
+// evaluateSuggestedActions determines repair actions based on reboot history and disk failure events.
+// Suggests up to 2 reboots, otherwise suggests hardware inspection.
+//
+// Logic mirrors GPU counts component:
+// - Case 1: First disk failure (no previous reboots) → suggest reboot
+// - Case 2: Disk failure after 1 reboot → suggest second reboot
+// - Case 3: Disk failure after 2+ reboots → suggest hardware inspection
+func evaluateSuggestedActions(cr *checkResult, rebootEvents eventstore.Events, diskFailureEvents eventstore.Events) {
```
eahydra: there are some similar functions in GPud, could you abstract them to implement a generic one?

components/disk/component.go
```diff
@@ -340,6 +467,63 @@ func (c *component) Check() components.CheckResult {
 
 	log.Logger.Debugw(cr.reason, "extPartitions", len(cr.ExtPartitions), "nfsPartitions", len(cr.NFSPartitions), "blockDevices", len(cr.BlockDevices))
 
+	// Check for disk failure events from kmsg (only if event store is available)
+	if c.eventBucket != nil && c.rebootEventStore != nil {
+		// Query recent events to check for disk failures
+		recentEvents, err := c.eventBucket.Get(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+		if err != nil {
+			log.Logger.Warnw("failed to get recent events", "error", err)
```
eahydra: it should fail fast and return from here so that we can remove the following else branch.

components/disk/component.go
```diff
@@ -340,6 +467,63 @@ func (c *component) Check() components.CheckResult {
...
+				rebootEvents, err := c.rebootEventStore.GetRebootEvents(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+				if err != nil {
+					cr.err = err
+					cr.health = apiv1.HealthStateTypeUnhealthy
+					log.Logger.Warnw("failed to get reboot events", "error", cr.err)
+					return cr
+				}
+
+				// Evaluate the suggested actions (along with the reboot history)
+				evaluateSuggestedActions(cr, rebootEvents, diskFailureEvents)
```
eahydra: diskFailureEvents contains three kinds of event, and it's not reasonable to evaluate suggested action with it..

gyuho:
@eahydra All addressed. PTAL

eahydra (approved):

PR #992, opened 2025-07-16
[LEP-1109] feat(gpud.service): add CPU/memory resource limits to prevent runaway processes

gyuho:
In systemd:
  - CPUQuota=100% = 1 full core
  - CPUQuota=200% = 2 full cores
  - CPUQuota=500% = 5 full cores

ref. https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html

### DIFF
--- a/pkg/gpud-manager/systemd/gpud.service
+++ b/pkg/gpud-manager/systemd/gpud.service
@@ -27,6 +27,43 @@ KillMode=control-group
 TimeoutStartSec=300
 CPUAccounting=true
 MemoryAccounting=true
+
+# Resource limits to prevent runaway processes
+# Baseline usage in production (128-core system): ~0.3% CPU (of total system), ~150MB RAM
+# CPU usage context: 0.3% in top = 0.3% of all cores combined
+# On a 128-core system, 0.3% total = ~38% of one core
+# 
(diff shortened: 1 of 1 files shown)


pkg/gpud-manager/systemd/gpud.service
```diff
@@ -27,6 +27,43 @@ KillMode=control-group
...
+# - Initialization spikes
+# - Periodic intensive operations (scanning, health checks)
+# - Multiple component polling cycles overlapping
+# On a 128-core system: 100% of one core = 0.78% of total capacity
+# This is ~2.6x the baseline usage, providing reasonable headroom
+#
+# CPUQuota=100% = 1 full core
+# CPUQuota=200% = 2 full cores
+# CPUQuota=500% = 5 full cores
+CPUQuota=100%
```
eahydra, line 47: I don't recommend that we merge it right now, we can try to update some of our nodes. We must be very careful with rolling upgrades to see if these configurations meet the requirements.

gyuho: Yeah it's being QA-ed right now

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/1049

PR #993, opened 2025-07-16
[LEP-1108] feat(os): track current number of goroutines as metrics

gyuho:
for future monitoring

### DIFF
--- a/components/os/metrics.go
+++ b/components/os/metrics.go
@@ -6,17 +6,27 @@ import (
 	pkgmetrics "github.com/leptonai/gpud/pkg/metrics"
 )
 
-const SubSystem = "os_fd"
+const metricSubSystem = "os"
 
 var (
 	componentLabel = prometheus.Labels{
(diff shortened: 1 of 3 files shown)


gyuho:
Discussed with @eahydra. We will expose this in `os` component, as metrics.

components/os/metrics.go
```diff
@@ -6,17 +6,30 @@ import (
 	pkgmetrics "github.com/leptonai/gpud/pkg/metrics"
 )
 
-const SubSystem = "os_fd"
+const (
+	subSystemGo = "os_go"
+	subSystemFD = "os_fd"
```
eahydra: shall we just use `os` as subsystem? and the metric metricGoroutines should be renamed to `self_goroutines`. Some metrics such as metricThresholdRunningPIDs, metricZombieProcesses do not belong to subSystemFD..

gyuho: @eahydra Done

eahydra (approved):

PR #999, opened 2025-07-19
[LEP-770] feat(disk): detect nvme issues (e.g., reset controllers)

gyuho:
Relies on https://github.com/leptonai/gpud/pull/990

### DIFF
--- a/components/disk/component.go
+++ b/components/disk/component.go
@@ -370,8 +370,11 @@ func (c *component) Check() components.CheckResult {
 		raidFailureEvents := make(eventstore.Events, 0)
 		fsReadOnlyEvents := make(eventstore.Events, 0)
 		nvmePathFailureEvents := make(eventstore.Events, 0)
+		nvmeTimeoutEvents := make(eventstore.Events, 0)
+		nvmeDeviceDisabledEvents := make(eventstore.Events, 0)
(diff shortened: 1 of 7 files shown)


components/disk/component.go
```diff
@@ -385,6 +388,22 @@ func (c *component) Check() components.CheckResult {
...
+				cr.health = apiv1.HealthStateTypeUnhealthy
+				failureReasons = append(failureReasons, "NVME device disabled after reset failure")
+			case eventBeyondEndOfDevice:
+				beyondEndOfDeviceEvents = append(beyondEndOfDeviceEvents, ev)
+				cr.health = apiv1.HealthStateTypeUnhealthy
+				failureReasons = append(failureReasons, "I/O beyond device boundaries detected")
+			case eventBufferIOError:
+				bufferIOErrorEvents = append(bufferIOErrorEvents, ev)
+				cr.health = apiv1.HealthStateTypeUnhealthy
+				failureReasons = append(failureReasons, "Buffer I/O error detected on device")
```
eahydra, line 407: is it possible that merge these cases into one?

eahydra (approved):

components/disk/component.go
gyuho: discussed offline, for now, we keep them separate and revisit

PR #1002, opened 2025-07-20
[LEP-765] feat(reboot): track reboot reason (if requested in session)

gyuho:


gyuho:
Discussed offline with @eahydra -- we will refer to other syslogs to derive the reboot reasons.

PR #1007, opened 2025-07-23
[LEP-1218] feat(row-remapping): inject failure by uuid (optional)

gyuho:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -56,7 +56,7 @@ type Config struct {
 	disabledComponents map[string]any `json:"-"`
 
 	// FailureInjector is the failure injector.
-	FailureInjector *components.FailureInjector `json:"-"`
+	FailureInjector *components.FailureInjector `json:"failure_injector,omitempty"`
 }
 
 func (config *Config) Validate() error {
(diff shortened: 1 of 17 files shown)


pkg/config/config.go
```diff
@@ -53,6 +54,9 @@ type Config struct {
 	Components         []string       `json:"components"`
 	selectedComponents map[string]any `json:"-"`
 	disabledComponents map[string]any `json:"-"`
+
+	// FailureInjector is the failure injector.
+	FailureInjector *components.FailureInjector `json:"-"`
```
eahydra: You have defined FailureInjector in components.GPUdInstance, why do we re-define it here?

eahydra (approved):

pkg/config/config.go
gyuho: Inside gpud server `func New(ctx context.Context, auditLogger log.AuditLogger, config *lepconfig.Config` is where we create GPUd instance, and we pas all the raw values from the flags via `*lepconfig.Config` to `gpudserver.New`

gyuho: Discussed offline, also added json tag.

PR #1027, opened 2025-08-02
[LEP-1385] fix(disk): skip nfs partitions with stat timeout

gyuho:
Possible that a partition was just un-mounted,
thus stat call times out.

### DIFF
--- a/components/disk/component.go
+++ b/components/disk/component.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"os"
 	"runtime"
 	"sort"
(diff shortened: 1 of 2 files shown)


components/disk/component.go
```diff
@@ -682,7 +665,16 @@ func (c *component) fetchNFSPartitions(cr *checkResult) bool {
 			continue
 		}
 
-		cr.NFSPartitions = parts
+		// possible that some partitions were just un-mounted
+		// thus stat call times out
+		for _, p := range parts {
+			if p.StatTimedOut {
+				log.Logger.Warnw("stat timed out for nfs partition", "device", p.Device, "mount_point", p.MountPoint)
+				continue
```
eahydra, line 673: Why do you hide the the error?  GPUd really detected the error, and we should report it, just need to 
1. make the error human-readable to let users know what the issue is.
2. currently, we can skip checking if the mountpoint is a subpath of a container.

gyuho: @eahydra See https://github.com/leptonai/gpud/pull/1029 instead that addresses these.

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/1029

PR #1038, opened 2025-08-11
[LEP-1564] feat(machine-info): log ASN lookup failures, raw ASN name for provider, retries in case of empty ASN name, add OCI as provider

gyuho:

### DIFF
--- a/pkg/machine-info/machine_info.go
+++ b/pkg/machine-info/machine_info.go
@@ -192,10 +192,11 @@ func GetProvider(publicIP string) *providers.Info {
 	if err != nil {
 		log.Logger.Warnw("failed to detect provider", "error", err)
 	} else {
-		log.Logger.Debugw("provider info", "provider", providerInfo, "error", err)
+		log.Logger.Debugw("provider detection result", "provider", providerInfo)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/machine-info/machine_info.go
```diff
@@ -216,11 +221,22 @@ func GetProvider(publicIP string) *providers.Info {
...
-		providerInfo.Provider = asn.NormalizeASNName(asnResult.AsnName)
+		log.Logger.Debugw("ASN lookup result", "asnResult", asnResult, "asnName", asnResult.AsnName)
+		normalizedProvider := asn.NormalizeASNName(asnResult.AsnName)
+		log.Logger.Debugw("normalized provider name", "normalizedProvider", normalizedProvider)
+
+		// Ensure we don't set an empty provider
+		if normalizedProvider != "" {
+			providerInfo.Provider = normalizedProvider
+		} else {
+			log.Logger.Warnw("normalized provider is empty, keeping unknown")
```
eahydra: still don't know the reason...

gyuho: logging errors in above lines

eahydra (approved):

PR #1039, opened 2025-08-13
feat: add dockerfile, update and reorganize helm chart

pteranodan:
### Changes

**Introduced a Dockerfile to:**
- Build a versioned, multi-architecture gpud image.
- Require the GPUd version as a mandatory build-time argument to ensure predictable, versioned builds.

**Refactored the Helm Chart to:**
- Deploy a container image specified in values.yaml instead of downloading the binary at runtime.
- Add a default nodeAffinity to ensure pods are scheduled only on nodes with NVIDIA GPUs.
- Abstract away internal details by using `gpud.telemetry.enabled` instead of the confusing double-negative `gpud.GPUD_NO_USAGE_STATS`.

### Benefits
- **Faster Startup**: Pods start instantly without needing to download and unpack the binary at runtime.
- **Improved Reliability**: The deployment no longer depends on network connectivity to an external release server during pod startup.
- **Enhanced Security**: The `curl | bash` pattern is removed, and the deployment uses an immutable, versioned image.
- **Better User Experience**: The chart now follows standard Helm conventions and uses a clear, intuitive configuration.


deployments/helm/gpud/values.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+# Liveness and Readiness probes.
+livenessProbe:
+  httpGet:
+    path: /healthz
+    port: http
+readinessProbe:
+  httpGet:
+    path: /ready
+    port: http
```
gyuho: Remove this? We don't have /ready endpoint

pteranodan: Removed. Left a comment indicating it is not defined.

PR #1045, opened 2025-08-20
[LEP-1325] [LEP-1736] feat(disk): remove "no space left" event, add health state checks based on free space left

gyuho:

### DIFF
--- a/components/disk/component.go
+++ b/components/disk/component.go
@@ -14,6 +14,7 @@ import (
 	"sync"
 	"time"
 
+	humanize "github.com/dustin/go-humanize"
 	"github.com/olekukonko/tablewriter"
 	"github.com/prometheus/client_golang/prometheus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -57,6 +58,11 @@ type component struct {
 
 	mountPointsToTrackUsage map[string]struct{}
 
+	// freeSpaceThresholdBytesDegraded is the threshold for the free space in bytes
+	// when the system is considered degraded.
+	// Default is 500MB.
+	freeSpaceThresholdBytesDegraded uint64
+
 	rebootEventStore pkghost.RebootEventStore
 	eventBucket      eventstore.Bucket
 	kmsgSyncer       *kmsg.Syncer
@@ -72,6 +78,8 @@ type component struct {
 const (
 	defaultRetryInterval  = 5 * time.Second
 	defaultLookbackPeriod = 14 * 24 * time.Hour // 14 days
+
+	defaultFreeSpaceThresholdBytesDegraded = 500 * 1024 * 1024 // 500 MB
 )
 
 func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
@@ -104,6 +112,8 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 
 		// Initialize file operation function field with real implementation
 		statWithTimeoutFunc: pkgfile.StatWithTimeout,
+
+		freeSpaceThresholdBytesDegraded: defaultFreeSpaceThresholdBytesDegraded,
 	}
 
 	if runtime.GOOS == "linux" {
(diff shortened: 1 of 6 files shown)


eahydra:
let's put it in next release

components/disk/component.go
```diff
@@ -57,6 +57,15 @@ type component struct {
...
 	mountPointsToTrackUsage map[string]struct{}
 
+	// usedSpaceThresholdPctDegraded is the threshold for the used space percentage
+	// when the system is considered degraded.
+	// Default is 0.99 (99%).
+	usedSpaceThresholdPctDegraded float64
+	// usedSpaceThresholdPctUnhealthy is the threshold for the used space percentage
+	// when the system is considered unhealthy.
+	// Default is 0.995 (99.5%).
+	usedSpaceThresholdPctUnhealthy float64
```
eahydra: We should account for differences in storage capacity, as some system disks are very large—exceeding 1TB—while others are quite small, typically only 100-200GB. Applying the same percentage threshold across all cases is unreasonable. There's no need to define two percentages; one is sufficient. If it exceeds the threshold, set it to Degraded.

components/disk/component.go
```diff
@@ -568,6 +583,84 @@ func (c *component) Check() components.CheckResult {
 		}
 	}
 
+	// Check disk usage thresholds
+	var degradedPartitions []string
+	var unhealthyPartitions []string
+
+	// Check ExtPartitions
+	for _, p := range cr.ExtPartitions {
```
eahydra, line 582: We should not check the usage of each partition.. We should check 1) root disk; 2) data disk if it exists

components/disk/component.go
```diff
@@ -568,6 +583,84 @@ func (c *component) Check() components.CheckResult {
...
+		if p.Usage == nil || p.Usage.TotalBytes == 0 {
+			continue
+		}
+
+		usagePercent := float64(p.Usage.UsedBytes) / float64(p.Usage.TotalBytes)
+
+		if usagePercent >= c.usedSpaceThresholdPctUnhealthy {
+			reason := fmt.Sprintf("%s (%s): %.1f%% used exceeds %.1f%% threshold (%s total)",
+				p.Device, p.MountPoint,
```
eahydra: Perhaps we can remove the device info, just keep the MountPoint in the reason.

components/disk/component.go
```diff
@@ -568,6 +583,84 @@ func (c *component) Check() components.CheckResult {
...
+	for _, p := range cr.NFSPartitions {
+		if p.Usage == nil || p.Usage.TotalBytes == 0 {
+			continue
+		}
+
+		usagePercent := float64(p.Usage.UsedBytes) / float64(p.Usage.TotalBytes)
+
+		if usagePercent >= c.usedSpaceThresholdPctUnhealthy {
+			reason := fmt.Sprintf("%s (%s): %.1f%% used exceeds %.1f%% threshold (%s total)",
+				p.Device, p.MountPoint,
```
eahydra: and here, remove p.Device

components/disk/component.go
gyuho: now only checking root

components/disk/component.go
gyuho: now only check low threshold (500MB)

components/disk/component.go
```diff
@@ -568,6 +578,58 @@ func (c *component) Check() components.CheckResult {
...
+		if p.Usage.FreeBytes < c.freeSpaceThresholdBytesDegraded {
+			reason := fmt.Sprintf("%s: free space %s is below %s threshold (%s total)",
+				p.MountPoint,
+				humanize.IBytes(p.Usage.FreeBytes),
+				humanize.IBytes(c.freeSpaceThresholdBytesDegraded),
+				humanize.IBytes(p.Usage.TotalBytes))
+			log.Logger.Debugw("reason", "device", p.Device)
+			degradedPartitionsDueToThresholdExceeded = append(degradedPartitionsDueToThresholdExceeded, reason)
+		}
+	}
```
eahydra: shall we skip it? unless someone requires we detect it.

eahydra (approved):

PR #1047, opened 2025-08-21
fix(containerd): fix dangling pod check when kubelet gives empty pods

cardyok:


components/containerd/component.go
```diff
@@ -403,7 +403,7 @@ func (cr *checkResult) HealthStates() apiv1.HealthStates {
 
 func danglingPodCount(containerdPods []pkgcontainerd.PodSandbox, kubeletPods []kubelet.PodStatus) int {
 	var danglingCount int
-	if kubeletPods == nil {
+	if kubeletPods == nil || len(kubeletPods) == 0 {
```
eahydra:
```suggestion
	if len(kubeletPods) == 0 {
```

gyuho (approved):

PR #1057, opened 2025-08-28
[LEP-1876] feat(nvidia/hw-slowdown): retain threshold exceeded events until set healthy

gyuho:


components/accelerator/nvidia/hw-slowdown/component.go
```diff
@@ -92,8 +96,11 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 	}
 
 	if gpudInstance.EventStore != nil {
+		// do not purge unless set healthy is called
+		// retain all hw slowdown events in order to evaluate hw inspection
+		// hw inspection should NOT be self-resolved unless set healthy is explicitly called
 		var err error
-		c.eventBucket, err = gpudInstance.EventStore.Bucket(Name)
+		c.eventBucket, err = gpudInstance.EventStore.Bucket(Name, eventstore.WithDisablePurge())
```
cardyok, line 103: if we fully disable purge hwslowdown will this cause gpud db explode? since for some instances we do notice hwslowdown flooding

cardyok (approved):

components/accelerator/nvidia/hw-slowdown/component.go
gyuho: Discussed offline, we will revisit this later to upper limit the storage size.

PR #1058, opened 2025-08-29
[LEP-1874] [LEP-2000] feat(components, disk): retain/lookback events 2 weeks

gyuho:

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -129,7 +129,7 @@ func New(ctx context.Context, auditLogger log.AuditLogger, config *lepconfig.Con
 	}
 
 	// by default, we only retain past 24 hours of events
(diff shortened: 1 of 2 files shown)


eahydra (approved):

pkg/server/server.go
```diff
@@ -129,7 +129,7 @@ func New(ctx context.Context, auditLogger log.AuditLogger, config *lepconfig.Con
 	}
 
 	// by default, we only retain past 24 hours of events
-	eventStore, err := eventstore.New(dbRW, dbRO, 24*time.Hour)
+	eventStore, err := eventstore.New(dbRW, dbRO, 14*24*time.Hour)
```
gyuho, line 132: @eahydra This is the shared event store instance that is also used in xid/sxid

PR #1060, opened 2025-08-29
feat(nvidia/infiniband, gpu-counts): do not filter out "SetHealthy" events

gyuho:


eahydra (commented):
we don't need it since we generate event in control-plane.

gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/1070.

PR #1065, opened 2025-09-03
[LEP-2083] fix(session): prevent deadlock waits on broken connection

gyuho:


eahydra:
@cardyok PTAL

cardyok (approved):

PR #1068, opened 2025-09-08
feat: refactor dockerfile and expose additional helm chart variables

### LINKED ISSUE #1067: Dockerfile and helm chart improvements
Hi,

While running a POC of gpud in a Kubernetes environment, we encountered a few issues:
1.	Dockerfile limitations
 - The Dockerfile always pulls the gpud binary from GitHub releases, making it impossible to test local changes to gpud.
 - The resulting Docker image doesn't have a `sudo` installed, that causes some checks to fail
2.	Missing required volumes
Some local host folders required by gpud components are not mounted as volumes in the gpud DaemonSet:
- `/var/log` - required for the `accelerator-nvidia-error-xid` component
- `/var/lib/systemd/pstore` - required for the `os` component

Additionally, we’d like to propose a few enhancements to the Helm chart that would simplify working with custom plugins:
- `extraVolumes` — allow specifying additional volumes for the gpud DaemonSet
- `extraVolumeMounts` — allow specifying additional volume mounts for the gpud DaemonSet
- `initContainers` — support adding initContainers to handle initialization tasks

Thanks!

MaxFedotov:
fixes #1067

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,9 +38,10 @@ export IMAGE_NAME="gpud"
 export OS_NAME="ubuntu"
 export OS_VERSION="22.04"
 export CUDA_VERSION="12.4.1"
+export GIT_COMMIT_HASH=$(git rev-parse --short HEAD)
 
 # Create the full tag and build the image
-export FULL_TAG="${IMAGE_NAME}:dev-cuda${CUDA_VERSION}-${OS_NAME}${OS_VERSION}"
+export FULL_TAG="${IMAGE_NAME}:${GIT_COMMIT_HASH}-cuda${CUDA_VERSION}-${OS_NAME}${OS_VERSION}"
 
 docker buildx build \
   --build-arg OS_NAME=${OS_NAME} \
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,9 @@ WORKDIR /workspace
 COPY go.mod go.mod
 COPY go.sum go.sum
 
+# Download dependencies as a separate step to take advantage of Docker's caching
+RUN go mod download
+
 COPY api/ api/
 COPY client/ client/
 COPY cmd/ cmd/


gyuho:
We will take a look!

/cc @pteranodan

Dockerfile
```diff
@@ -6,34 +6,31 @@ ARG OS_NAME="ubuntu"
...
 
-RUN if [ -z "$GPUD_VERSION" ]; then \
-        echo "\n\nERROR: The build argument 'GPUD_VERSION' is required.\n" >&2; \
-        echo "Please provide it using: --build-arg GPUD_VERSION=<version>\n" >&2; \
-        exit 1; \
-    fi
+WORKDIR /workspace
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
```
pteranodan, line 17: Thanks for the contribution, @MaxFedotov!

I have a small suggestion. If we copy over the `go.mod` and `go.sum` files and run `go mod download` before copying the rest of the source code, Docker will only re-download dependencies when those module files actually change. What are your thoughts?

Dockerfile
```diff
@@ -6,34 +6,31 @@ ARG OS_NAME="ubuntu"
...
-    URL="https://github.com/leptonai/gpud/releases/download/${GPUD_VERSION}/gpud_${GPUD_VERSION}_linux_${TARGETARCH}_${OS_NAME}${OS_VERSION}.tgz" && \
-    curl -L "${URL}" | tar -xz -C /usr/local/bin/ gpud
+COPY api/ api/
+COPY client/ client/
+COPY cmd/ cmd/
+COPY components/ components/
+COPY docs/ docs/
+COPY pkg/ pkg/
+COPY version/ version/
+COPY Makefile Makefile
```
pteranodan, line 29: Thoughts on simplifying this to a single `COPY . .` command and managing exclusions in a `.dockerignore` file?

CONTRIBUTING.md
```diff
@@ -35,16 +35,14 @@ This is the fastest way to build an image for your local machine's architecture
...
 # Set local build variables
 export IMAGE_NAME="gpud"
-export GIT_TAG="v0.6.0"
 export OS_NAME="ubuntu"
 export OS_VERSION="22.04"
 export CUDA_VERSION="12.4.1"
 
 # Create the full tag and build the image
-export FULL_TAG="${IMAGE_NAME}:${GIT_TAG#v}-cuda${CUDA_VERSION}-${OS_NAME}${OS_VERSION}"
+export FULL_TAG="${IMAGE_NAME}:dev-cuda${CUDA_VERSION}-${OS_NAME}${OS_VERSION}"
```
pteranodan: Instead of using a generic prefix like `dev-`, can we use the short git commit hash for the image tag prefix? Makes every development build unique and easy to trace back to the exact version of the code it was created from.

Dockerfile
MaxFedotov: The only thing that worries me with adding them to `.dockerignore` is that we are abstracting some build details from the `Dockerfile` and that can make troubleshooting a bit harder. It is much easier when everything is in the single file, even if this makes it a bit bigger

Dockerfile
MaxFedotov: hey @pteranodan! 100% agree, will update the PR

CONTRIBUTING.md
MaxFedotov: Sure, will update the PR

gyuho:
@MaxFedotov Thanks for the contribution. We have many others that look to support GPUd in container mode. Please let us know how it goes, or feel free to email us at gyuhol@nvidia.com!

Dockerfile
pteranodan: Fair point. The main reason I suggested the `COPY . .` command together with a `.dockerignore` file, is that this pattern is recommended by Docker. It avoids having to update the Dockerfile when new directories are added, it prevents sensitive or unnecessary files from ever being sent to the Docker daemon (keeping the build context small and is a crucial security best practice), and makes exclusions explicit.

/cc @gyuho

PR #1069, opened 2025-09-10
feat(components/library): match more nvidia lib paths

gyuho:
- Add more lib paths to the search dirs.


cardyok:
how about components/library/component.go..

cardyok (approved):

Issue #1072, opened 2025-09-16
Reseting Unhealthy HARDWARE_INSPECTION  repair action for accelerator-nvidia-error-xid component


MaxFedotov:
Hey @gyuho!

I'm trying to figure out the correct way to set `accelerator-nvidia-error-xid` component healthy after injecting a few Xid errors and performing the reboots.

The current state that i'm observing is
```
{
   "component":"accelerator-nvidia-error-xid",
   "states":[
      {
         "time":null,
         "name":"error_xid",
         "health":"Unhealthy",
         "reason":"XID 48 (Double Bit ECC Error) detected on GPU PCI:0000:6b:00 UUID:GPU-c44c6e5f-72cf-8200-97c2-f9b9ebcf2dbb",
         "suggested_actions":{
            "description":"",
            "repair_actions":[
               "HARDWARE_INSPECTION"
            ]
         }
      }
   ]
},

```
Can you please recommend the correct way to reset this state?

Thanks!

MaxFedotov:
I've seen that there is a `SetHealthy` event, that can be used to reset the state, but seems like it can be send only by gpud controller in lepton cloud. May be it is worth exposing additional API in order to be able to send this event locally?

MaxFedotov:
@gyuho , i've went through the code one more time and have one idea - what do you think about adding a new `set-healthy` command for gpud and also exposing a new api endpoint `/v1/set-healthy` that will accept the component name and will set the component healthy, basically implement the same logic as here but for users that are not using lepton cloud: https://github.com/leptonai/gpud/blob/fc398aa0c9bb6a7d8e6ea41e6d08f29734c1a41e/pkg/session/serve.go#L184-L200)

If you are ok with this approach, i will be glad to make a PR :)

gyuho:
@MaxFedotov You are right. This is a new API so we are slowly introducing component by component. But for the local API, I quickly put up a PR to support it. Please review/try and provide any feedback you may have: https://github.com/leptonai/gpud/pull/1073

> curl -s -kL https://localhost:15132/v1/health-states/set-healthy?components=accelerator-nvidia-error-xid

Thanks!

MaxFedotov:
Wow, that was very fast, thanks a lot @gyuho!

What do you think about also adding a  new `set-healthy` command to gpud binary? API can be used for various automations and this new command can be used by operators to manually set components healthy.

gyuho:
[...]

PR #1073, opened 2025-09-17
feat(server,cmd): add set healthy states local API endpoint, add "gpud set-healthy" command

gyuho:
Address https://github.com/leptonai/gpud/issues/1072.

> curl -s -kL https://localhost:15132/v1/health-states/set-healthy?components=accelerator-nvidia-error-xid


MaxFedotov (approved):
Thanks a lot @gyuho! That's what we really need

eahydra (approved):

PR #1074, opened 2025-09-17
[LEP-2340] feat(nvidia/gsp-firmware-mode): fallback using kernel config file

gyuho:
NVML may report GSP as enabled even when it's disabled at the kernel level.
Fallback mechanism to validate the GSP firmware mode against kernel module configuration.

```
$ cat /etc/modprobe.d/nvidia.conf
options nvidia NVreg_EnableGpuFirmware=0

$ nvidia-smi -q | grep "GSP"
    GSP Firmware Version                  : N/A
```


gyuho:
Closing in favor of https://github.com/leptonai/gpud/pull/1083.

PR #1082, opened 2025-09-23
[LEP-2083] fix(session): async triggerComponent response writes to avoid blocking the main serve loop

gyuho:
A manual `triggerComponent`/`triggerComponentCheck` request against the disk component blocked GPUd's control-plane session loop. While the disk health probe sat in a long retry cycle, no responses were written back to the control plane, so GPUd appeared healthy locally but the control plane stayed stuck on stale health data.

**Issue**
- Control-plane machine object continued to show `nfs` health at `2025-09-18T05:48:16Z`, even though GPUd's local disk state had fresh timestamps (for example `2025-09-18T12:07:16Z`).
- GPUd logs repeated:
  - `session reader: error decoding response: EOF`
  - `session writer: error making request ... context canceled`
  - `drained stale messages from reader channel` during keep-alive resets.

**Root Cause**
1. `Session.serve` processes every control-plane message in a single goroutine. Before the fix it handled `triggerComponent` synchronously, so it waited for `comp.Check()` to finish before writing any response.
2. `components/disk/component.go` retries `disk.GetPartitions` up to **five** times per tracked mount, each wrapped in a **one-minute** timeout. A flaky `/mnt/nfs-share` can therefore block `Check()` for roughly five minutes per mount.
3. While `Check()` blocks, the writer goroutine cannot drain the 20-item `s.writer` channel. The control plane eventually cancels the streaming HTTP request, triggering the `session reader`/`session writer` error messages and the `drained stale messages` warning when the keep-alive loop tears everything down.
4. The disk component's background ticker keeps updating `lastCheckResult`, so GPUd's local state reflects the latest degraded result. However, the control plane never receives that update because the synchronous response never flushes.

**One** manual `triggerComponent` aimed at the disk component is sufficient to deadlock the session loop. The request is enqueued, `Session.serve` synchronously calls the long-running `disk.Check()`, and the writer never sends a response. When the control plane cancels the stuck request, GPUd restarts the session and the control plane still holds the last good health snapshot.

**Fix**
- `triggerComponent` will be processed asynchronously in a background goroutine.
- The response will be written back to the control plane once complete, with the same ReqID.

c.f., https://github.com/leptonai/gpud/pull/1078


pkg/session/serve.go
```diff
@@ -454,6 +429,114 @@ func (s *Session) serve() {
 	}
 }
 
+// processTriggerComponentRequest runs entirely in a background goroutine.
+// The worker reproduces the synchronous logic we historically ran in the main serve loop, but keeps
+// all potentially blocking `Check()` calls off the hot path so other control-plane requests can be
+// decoded and handled immediately. The ReqID travels with the request so the control plane can
+// correlate the eventually-delivered response.
+func (s *Session) processTriggerComponentRequest(reqID, method string, payload Request) {
```
eahydra: we have method name and payload, we can do much more like make the implementation more general.

PR #1089, opened 2025-10-01
[LEP-2600] fix(peermem,disk,*): run bash inline (without writing to disk)

gyuho:
fix

> open /tmp/gpud-453112547.bash: no space left on device
>
> write /tmp/gpud-453112547.bash: no space left on device


pkg/disk/lsblk.go
```diff
@@ -236,6 +238,7 @@ func fillFstypeFromFindmnt(ctx context.Context, dev *BlockDevice, cache map[stri
 
 // parseLsblkJSON parses the "lsblk --json" output.
 func parseLsblkJSON(ctx context.Context, b []byte, opts ...OpOption) (BlockDevices, error) {
+	fmt.Println("parseLsblkJSON", string(b))
```
eahydra: should revert it?

PR #1093, opened 2025-10-01
fix(set-healthy): allow to provide empty components list, fix incorrect response struct in the client

MaxFedotov:
**Summary of Changes**
These modifications fix the set-healthy API behavior to allow setting all health-settable components to healthy when no specific components are specified:

**Key Changes:**
1. Server Handler ( pkg/server/handlers_components.go)
- Removed requirement for  components parameter - now optional instead of mandatory
- When no components specified, only reports failures for components that were explicitly requested (avoids noise from non-health-settable components)
2. Client API ( client/v1/v1_set_healthy.go)
- Returns list of successfully set components instead of just error/nil
- Fixed JSON field name: success → successful (matches server response)
- Fixed type: Code field changed from  string to int
- Better visibility into which components were actually set to healthy
3. CLI Command ( cmd/gpud/set-healthy/command.go)
- Removed empty components check - allows calling without specifying components
- Displays actual successful components from server response instead of input list
4. Tests Updated
- E2E test now expects StatusOK instead of StatusBadRequest for empty components
- Client tests updated to handle new return signature

### DIFF
--- a/client/v1/v1_set_healthy.go
+++ b/client/v1/v1_set_healthy.go
@@ -14,15 +14,15 @@ import (
 )
 
 // SetHealthyComponents sets specified components to healthy state
-func SetHealthyComponents(ctx context.Context, addr string, components []string, opts ...OpOption) error {
+func SetHealthyComponents(ctx context.Context, addr string, components []string, opts ...OpOption) ([]string, error) {
 	op := &Op{}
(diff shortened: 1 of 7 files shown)


gyuho:
@MaxFedotov I have simplified server response struct in the last commit, please take a look

gyuho (approved):

PR #1097, opened 2025-10-03
fix(infiniband): fix int conversion overflow

gyuho:


pkg/nvidia-query/infiniband/class/value_parser.go
```diff
@@ -29,7 +33,23 @@ func newValueParser(v string) *valueParser {
...
+		return 0
+	}
+
+	v := vp.int64()
+	if vp.err != nil {
+		return 0
+	}
+
+	// Check bounds to prevent integer overflow on 32-bit systems
+	if v > math.MaxInt || v < math.MinInt {
```
eahydra: I'm curious why you not use math.MaxInt32 and math.MinInt32, and do we need to support 32-bit systems?

gyuho: > do we need to support 32-bit systems

No, let me drop

PR #1110, opened 2025-10-19
[LEP-2905] feat(nvidia/infiniband): make ib port drop event more persistent, rename more packages to nvidia/

gyuho:
Previously, IB port drops were "ephemeral" - as soon as thresholds recovered
(e.g., port came back up), the component would immediately flip from Unhealthy
back to Healthy. This created operator confusion because:
1. At time T: Component marked Unhealthy with HARDWARE_INSPECTION suggested
2. At time T+30s: Port recovers, component immediately becomes Healthy
3. Operators miss the alert or find contradictory states in logs

With this sticky window, drops remain unhealthy for a stabilization period,
giving operators time to observe and investigate, similar to how flaps work.
This creates consistency: both flaps and drops suggest inspection that persists.


eahydra:
> At time T: Component marked Unhealthy with HARDWARE_INSPECTION suggested
> At time T+30s: Port recovers, component immediately becomes Healthy

Why does it happen if the component marked Unhealthy with HARDWAR_INSPECTION suggested?   From the time series, it seems there is a race condition?

gyuho:
> At time T+30s: Port recovers, component immediately becomes Healthy
>
> Why does it happen if the component marked Unhealthy with HARDWAR_INSPECTION suggested

It did because the threshold check takes more priority, overwriting the previous failure.

Evaluated thresholds first. When the port recovers and thresholds pass, it sets the component to Healthy. Then when checking historical drop events, the old code would only process them if thresholds were currently failing  (len(unhealthyIBPorts) > 0). This meant once the port recovered, the drop event was effectively ignored, causing the immediate Unhealthy --> Healthy flip.

This adds a sticky window for new drop events:
1. Thresholds currently failing (original behavior)
2. Drop occurred within the sticky window (new)
3. Thresholds just recovered but still within stabilization period (new)

This prevents the confusing state flip and gives operators time to observe the issue.

cardyok (approved):

Issue #1107, opened 2025-10-16
no block device found -- something must be wrong with lsblk command -- LVM related?


martbhell:
Hello, we have a machine or two with some disks in an LVM. The other machines that do not have such an LVM do not see this lsblk issue from gpud.

```
$ gpud --version
gpud version v0.8.0-rc.13
$ lsblk --version
lsblk from util-linux 2.37.2
```

The first line is truncated. That's how it is in syslog :/

If you need more details just let me know.
The `lsblk` command prints disks if one runs it by itself. Maybe something is null that it expects or something like that?

```
[...]

PR #1111, opened 2025-10-20
[LEP-3109] fix(pkg/disk): handle raid/md devices, recursive lsblk parse for more depth

gyuho:
Address https://github.com/leptonai/gpud/issues/1107.


pkg/disk/lsblk_issue_1107_test.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+
+	// Both devices should be found with the complete fix
+	assert.Equal(t, 2, foundDevices, "both problematic devices should be found with complete fix")
+	assert.NotZero(t, foundDevices, "should NOT trigger 'no block device found' error")
+}
+
+// TestIssue1107_DocumentWhyBothChangesNeeded documents the necessity of both changes
+func TestIssue1107_DocumentWhyBothChangesNeeded(t *testing.T) {
```
eahydra: Why do we need such a test function?  we can add comments or write a detailed git message.

gyuho: Oh, forgot to implement this... removing

PR #1125, opened 2025-10-30
[LEP-3112] feat(cmd/gpud): remove "gpud login" command in favor of "gpud up", remove/move unused code

gyuho:


gyuho:
@eahydra Yes, it's removed

PR #1133, opened 2025-11-05
[LEP-3205] fix(components/nvidia): handle "GPU requires reset" NVML error to set suggested actions

gyuho:

### DIFF
--- a/pkg/nvidia-query/nvml/error.go
+++ b/pkg/nvidia-query/nvml/error.go
@@ -18,6 +18,24 @@ var (
 	ErrGPURequiresReset = errors.New("GPU requires reset")
 )
 
+// IsGPULostError returns true if the error indicates that the GPU is lost.
+// "if the target GPU has fallen off the bus or is otherwise inaccessible".
+func IsGPULostError(ret nvml.Return) bool {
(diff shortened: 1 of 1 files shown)


components/accelerator/nvidia/utilization/component.go
```diff
@@ -136,6 +137,17 @@ func (c *component) Check() components.CheckResult {
 			cr.err = err
 			cr.health = apiv1.HealthStateTypeUnhealthy
 			cr.reason = "error getting utilization"
+
+			if errors.Is(err, nvidianvml.ErrGPURequiresReset) {
+				cr.reason = "GPU requires reset, please reboot the machine"
```
eahydra: shall we remove `please reboot the machine`?

components/accelerator/nvidia/utilization/component.go
```diff
@@ -136,6 +137,17 @@ func (c *component) Check() components.CheckResult {
...
+				cr.suggestedActions = &apiv1.SuggestedActions{
+					Description: "GPU requires reset, please reboot the machine",
```
eahydra: and here

Issue #1144, opened 2025-11-14
MountVolume.SetUp failed for volume "host-dev-mem" : hostPath type check failed: /dev/mem is not a character device


brannondorsey:
We're deploying the v0.8.0 in a GKE environment. The `host-dev-mem` device is expected to be a `CharDevice` [here](https://github.com/leptonai/gpud/blob/5d09f94a2c3e6ffaf50905fef0aaf3b95f3fc952/deployments/helm/gpud/templates/daemonset.yaml#L85), but in our case, it isn't.

This results in all daemonset pods getting stuck in `ContainerCreating` with this warning event:

```
Warning  FailedMount  12s (x9 over 2m20s)  kubelet            MountVolume.SetUp failed for volume "host-dev-mem" : hostPath type check failed: /dev/mem is not a character device
```

At minimum, this should be configurable via the helm chart values.

brannondorsey:
I've enumerated the following volume types from [here](﻿kubernetes.io/docs/concepts/storage/volumes/#hostpath-volume-types) and all of them exhibited the same error.

* CharDevice
* BlockDevice
* File
* Directory
* Socket

Removing that volume and volume mount got the pods to finish creating at least.

gyuho:
@brannondorsey Hello, thanks for the detailed report. Running GPUd inside a container is definitely on our roadmap, and the work is still ongoing -- contributions are welcome. Because the early versions assumed privileged host access, we need to iterate on what permissions the daemonset/container must have. We’ll dig into this and follow up.

brannondorsey:
Well the good news is I got this daemonset working correct on GKE with the following changes.

All we had to do was:

* Remove the `/dev/mem` volume + mount from the underlying chart's `deployment.yaml`
* Publish our own docker images based off the Dockerfile in this repo
* Create a `PodMonitor` to scrape these daeomonset pods
* Use these helm values with the 0.8.0 chart...

```yaml
# Common values for gpud chart across all clusters

image:
  # We built and published our own image using the Dockerfile in this repo
  # Tag format: {version}-cuda{cuda_version}-{os}{os_version}
  tag: 0.8.0-cuda13.0.0-ubuntu22.04
  pullPolicy: IfNotPresent

# Overrides the default node selector to use example.com/gpu instead of nvidia.com/gpu.product
# (presumably because we use the k8s-device-plugin without the node feature discovery daemonset)
affinity:
  nodeAffinity:
[...]

PR #1149, opened 2025-11-19
[LEP-2916] feat(helm/gpud): support label reader, fix mounts (optional)

gyuho:
Address https://github.com/leptonai/gpud/issues/1144.

### DIFF
--- a/deployments/helm/gpud/templates/daemonset.yaml
+++ b/deployments/helm/gpud/templates/daemonset.yaml
@@ -27,6 +27,8 @@ spec:
     spec:
       # --- Pod Identity & Security ---
       serviceAccountName: {{ include "gpud.serviceAccountName" . }}
+      # K8s auto-mounts SA tokens unless explicitly disabled; see https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-automounting-service-account-tokens
+      automountServiceAccountToken: {{ .Values.serviceAccount.automount }}
       {{- with .Values.podSecurityContext }}
       securityContext:
         {{- toYaml . | nindent 8 }}
@@ -37,9 +39,11 @@ spec:
       {{- end }}
 
       # --- Node-level Settings & Lifecycle ---
-      # Required for the driver to interact with the host's network stack.
+      # Required for the driver to interact with the host's network stack (e.g. for correct IP reporting).
+      # GPUd is intentionally exposed via host networking, so there is no Service object for this chart.
+      # Consumers should reach GPUd via each node's IP/port (e.g. https://<node-ip>:15132).
       hostNetwork: true
-      # Required for the driver to inspect processes on the host.
+      # Required for the driver to inspect processes on the host (e.g. via /proc).
       hostPID: true
       restartPolicy: Always
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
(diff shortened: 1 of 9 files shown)


deployments/helm/gpud/values.yaml
```diff
@@ -8,29 +8,59 @@ nameOverride: ""
 fullnameOverride: ""
 
 image:
-  repository: gpud
+  repository: nvcr.io/dgx-cloud-lepton/gpud
```
brannondorsey, line 11: Woohoo! Love that an image is now going to be published.

This was one of the biggest barriers to entry for us getting started with this daemonset, so hopefully this will help others get started quicker than we did.

deployments/helm/gpud/templates/daemonset.yaml
```diff
@@ -131,23 +286,106 @@ spec:
...
               readOnly: true
+
+            {{- if .Values.gpud.mountHostDevMem }}
+            # Physical memory access for hardware diagnostics (read-only)
+            # Conditional mount - can be disabled for GKE/restricted environments
+            # via gpud.mountHostDevMem=false (addresses issue #1144)
             - name: host-dev-mem
               mountPath: /dev/mem
               readOnly: true
+            {{- end }}
```
brannondorsey, line 366: This looks looks good to me and should fix #1144 👍

deployments/helm/gpud/templates/daemonset.yaml
```diff
@@ -131,23 +286,106 @@ spec:
...
+              mountPath: /sys
+              readOnly: true
+
+            # Procfs - Kernel/process info (read-only for safety)
+            # - /proc/sys/fs/file-*: File descriptor limits and usage
+            # - /proc/sys/kernel/random/boot_id: Unique boot identifier
+            # - /proc/self/mountinfo: Current mount points
+            - name: host-proc
+              mountPath: /proc
+              readOnly: true
```
brannondorsey, line 347: Mounting both of these directories come with some unfortunate security considerations. But I'm not sure there is a way around this if the user wants to track these metrics.

It could make sense to wrap these in optional values as well, perhaps default `true`, to give the user control to opt-out of mounting these host directories in sensitive environments if they so choose.

That is, so long as `gpud` continues to function as you expect without access to these directories.

deployments/helm/gpud/templates/daemonset.yaml
```diff
@@ -96,31 +150,132 @@ spec:
...
             - name: GPUD_NO_USAGE_STATS
               value: {{ not .Values.gpud.telemetry.enabled | quote }}
 
           {{- if .Values.service.enabled }}
           ports:
-            - name: http
-              containerPort: {{ .Values.service.port }}
+            - name: https
+              # Must match the port in gpud.listenAddress (default 15132)
+              containerPort: {{ .Values.livenessProbe.httpGet.port }}
```
brannondorsey: Nice! I'm glad you resolved this as well, as we needed to change `.Values.service.port` to 15132 to get a ServiceMonitor to work, before switching to PodMonitor.

What do you think about removing the `.Values.service.port` entirely if you're going to use the liveness probe port instead?

brannondorsey: On second thought, it might actually be more semantic to use `.Values.service.port` here but default it to port 15132.

The reason being that someone could define a livenessProbe in their `values.yaml` file that doesn't use an `httpGet` at all, and this would fail.

deployments/helm/gpud/templates/rbac.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if and .Values.nodeLabelExporter.enabled .Values.nodeLabelExporter.rbac.create }}
```
brannondorsey, line 1: This is nice.

deployments/helm/gpud/templates/daemonset.yaml
```diff
@@ -96,31 +150,132 @@ spec:
...
+              fi
+              if [ -n "$MID" ]; then
+                set -- "$@" --machine-id="$MID"
+              fi
+
+              set -- "$@" --enable-auto-update={{ .Values.gpud.enableAutoUpdate }}
+              set -- "$@" --auto-update-exit-code={{ .Values.gpud.autoUpdateExitCode }}
+
+              echo "Starting: /usr/local/bin/gpud $*"
+              exec /usr/local/bin/gpud "$@"
```
brannondorsey, line 292: This `exec` is so important to make signal handling/shutdown continue to work when switching to an inline shell script entrypoint.

deployments/helm/gpud/templates/daemonset.yaml
```diff
@@ -96,31 +150,132 @@ spec:
...
+              set -- "$@" --log-level="{{ .Values.gpud.logLevel }}"
+
+              # Determine endpoint (label takes precedence over values.yaml)
+              ENDPOINT="{{ .Values.gpud.endpoint }}"
+              if [ -n "$EP" ]; then
+                ENDPOINT="$EP"
+              fi
+              if [ -n "$ENDPOINT" ]; then
+                set -- "$@" --endpoint="$ENDPOINT"
+              fi
```
brannondorsey, line 276: I haven't tested this, and feel free to ignore this comment if you disagree... but this new approach feels a bit fragile. 

In general I find these sorts of BASH + AWK parsing solutions to be prone to silent failures and issues down the line. The way I've solved this in the past is to add a small Go binary whose whole job it is is to print arguments (or invoke the entrypoint) given an environment.

That way you get modern language features and error handling, which is much easier to maintain into the future.

deployments/helm/gpud/values.yaml
gyuho: @brannondorsey Yes, we are actively working on the public image, will keep you posted!

cardyok (approved):

PR #1161, opened 2025-12-01
Fix note on install script architecture support

martbhell:
Corrected the note regarding architecture and OS support for the install script.


gyuho:
Thanks for the contribution!

PR #1163, opened 2025-12-01
[LEP-3306] feat(disk, memory): log warn level when threshold exceeds in case of node pressure eviction

gyuho:
Useful for debugging.

ref. https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/

### DIFF
--- a/components/disk/component.go
+++ b/components/disk/component.go
@@ -63,6 +63,13 @@ type component struct {
 	// Default is 500MB.
 	freeSpaceThresholdBytesDegraded uint64
 
+	// freeSpaceThresholdPercent is the threshold for free space as a percentage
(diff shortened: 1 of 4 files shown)


components/disk/component.go
```diff
@@ -121,6 +134,7 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 		statWithTimeoutFunc: pkgfile.StatWithTimeout,
 
 		freeSpaceThresholdBytesDegraded: defaultFreeSpaceThresholdBytesDegraded,
+		freeSpaceThresholdPercent:       defaultFreeSpaceThresholdPercent,
```
calmkart, line 137: maybe we should support this value change by config or env in future

PR #1180, opened 2025-12-17
[LEP-3777] fix(nvidia): do not exit on NVML init error (instead returns Unhealthy)

gyuho:

### DIFF
--- a/components/disk/component_test.go
+++ b/components/disk/component_test.go
@@ -2858,6 +2858,10 @@ func TestComponent_Ext4PartitionsTimeoutSetsReasonAndError(t *testing.T) {
 		_ = c.Close()
 	}()
 
+	// Set a non-zero retry interval to avoid race condition between
+	// context.Done and time.After in the select statement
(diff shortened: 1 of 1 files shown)


components/accelerator/nvidia/clock-speed/component.go
```diff
@@ -68,6 +68,12 @@ func (c *component) IsSupported() bool {
 	if c.nvmlInstance == nil {
 		return false
 	}
+	// Return true when there's an init error so the component can report unhealthy.
+	// This handles cases where NVML library loads but device enumeration fails,
+	// e.g., "error getting device handle for index 'N': Unknown Error"
+	if c.nvmlInstance.InitError() != nil {
```
eahydra: It's a bit odd that it still returns true even if initialization fails?
Perhaps we can simply return true, since I noticed that the NVMLExists function merely attempts to check whether the nvml library exists and should not cause any side effects.

eahydra (approved):

PR #1188, opened 2026-01-05
feat(infra): add gpud session method to update token and get token

calmkart:
feat(infra): add gpud session method to update token and get token

### DIFF
--- a/pkg/session/token.go
+++ b/pkg/session/token.go
@@ -1,0 +1,113 @@
+package session
+
+import (
+	"context"
+	"fmt"
+	"net/http/cookiejar"
+	"time"
+
+	"github.com/leptonai/gpud/pkg/log"
+	pkgmetadata "github.com/leptonai/gpud/pkg/metadata"
+)
+
+// processUpdateToken updates the session token in the database with the new token from the control plane.
+// It validates the new token before updating, and immediately reconnects the session with the new token.
+func (s *Session) processUpdateToken(payload Request, response *Response) {
+	if payload.Token == "" {
+		response.Error = "token cannot be empty"
+		log.Logger.Errorw("updateToken request with empty token")
+		return
+	}
+
+	// if in-memory cached token equal update token, just return is ok
+	cacheToken := s.getToken()
+	if cacheToken == payload.Token {
+		log.Logger.Info("cached token already matches the requested token")
+		return
+	}
+
+	if s.dbRW == nil {
+		response.Error = "database connection not available"
+		log.Logger.Errorw("updateToken failed: database connection is nil")
+		return
+	}
+
+	// Validate the new token before updating
+	log.Logger.Infow("validating new token with control plane")
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	if err := s.validateTokenWithHealthCheck(ctx, payload.Token); err != nil {
+		response.Error = fmt.Sprintf("token validation failed: %v", err)
+		log.Logger.Errorw("new token validation failed", "error", err)
+		return
+	}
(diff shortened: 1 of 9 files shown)


pkg/session/token.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/leptonai/gpud/pkg/log"
+	pkgmetadata "github.com/leptonai/gpud/pkg/metadata"
+)
+
+// processUpdateToken updates the session token in the database with the new token from the control plane.
+func (s *Session) processUpdateToken(payload Request, response *Response) {
```
eahydra, line 15: Shall we verify if the token is valid before updating token?
And after updating token, we should re-create session?

calmkart: make sense.i'll update later

eahydra: It is important to note that the session manager implementation on control-plane side may conflict with this token verification scenario.

pkg/session/token.go
```diff
@@ -0,0 +1,75 @@
...
+	if payload.Token == "" {
+		response.Error = "token cannot be empty"
+		log.Logger.Errorw("updateToken request with empty token")
+		return
+	}
+
+	// if in-memory cached token equal update token, just return is ok
+	cacheToken := s.token
```
gyuho: Use `s.getToken` to avoid racey acceess?

pkg/session/token.go
```diff
@@ -0,0 +1,75 @@
...
+	if cacheToken == payload.Token {
+		log.Logger.Printf("exist token is already equal update token")
```
gyuho:
```suggestion
		log.Logger.Info("cached token already matches the requested token")
```
Or something similar?

pkg/session/token_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	s.processGetToken(response)
+
+	if response.Error == "" {
+		t.Error("expected error when database is not available")
+	}
+}
+
+func TestTokenConcurrency(t *testing.T) {
+	t.Skip()
```
gyuho: Why skip? And can we improve this test to actually test the final set token is the expected one?

pkg/session/token.go
```diff
@@ -0,0 +1,75 @@
...
+	defer cancel()
+
+	token, err := pkgmetadata.ReadToken(ctx, s.dbRO)
+	if err != nil {
+		response.Error = err.Error()
+		log.Logger.Errorw("failed to read token from database", "error", err)
+		return
+	}
+
+	response.Token = token
```
gyuho, line 111: Missing setToken? So, we are not really cach-ing anything here?

pkg/session/token.go
calmkart: ok

pkg/session/token.go
calmkart: ok

pkg/session/token.go
calmkart: done

pkg/session/token_test.go
calmkart: ok，I'm concerned that these tests depend on the SQLite installation, and there might be environment issues preventing them from running. However, this isn't really a problem, as it's already enabled.

pkg/session/token.go
calmkart: done

pkg/session/token.go
calmkart: done

test_output.log
eahydra, line 1: what's this?

eahydra:
overall lgtm, defer to @gyuho

gyuho (approved):

PR #1199, opened 2026-01-23
make gpud port number configurable

aimichal:
I'd like to be able to choose the port number that gpud exposes.

This change renames `DefaultGPUdPort` to `defaultGPUdPort` to scope its visibility to the config package, and exposes a function `GPUdPortNumber()` that reads env var GPUD_PORT but falls back to defaultGPUdPort if it's missing or not an int.

It would be nice to factor this such that the port number is considered only when building the config (in DefaultConfig()) with a nice error message if it's malformed. But today `DefaultGPUdPort` (now `GPUdPortNumber()` is used throughout the codebase, so I decided to propose this surgical change to start.

### DIFF
--- a/pkg/config/default.go
+++ b/pkg/config/default.go
@@ -16,7 +16,7 @@ import (
 const (
 	DefaultAPIVersion = "v1"
 	DefaultDataDir    = "/var/lib/gpud"
-	defaultGPUdPort   = 15132
+	DefaultGPUdPort   = 15132
 )
 
 var (
@@ -146,12 +146,12 @@ func VersionFilePath(dataDir string) string {
 func GPUdPortNumber() int {
 	portStr, found := stdos.LookupEnv("GPUD_PORT")
 	if !found {
-		return defaultGPUdPort
+		return DefaultGPUdPort
 	}
 
 	port, err := strconv.Atoi(portStr)
 	if err != nil {
-		return defaultGPUdPort
+		return DefaultGPUdPort
 	}
 
 	return port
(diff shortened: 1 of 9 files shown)


pkg/config/default.go
```diff
@@ -141,3 +142,17 @@ func PackagesDir(dataDir string) string {
...
+
+func GPUdPortNumber() int {
+	portStr, found := stdos.LookupEnv("GPUD_PORT")
+	if !found {
+		return defaultGPUdPort
+	}
+
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return defaultGPUdPort
```
gyuho:
```suggestion
		return DefaultGPUdPort
	}

	port, err := strconv.Atoi(portStr)
	if err != nil {
		return DefaultGPUdPort
```

We can still keep this external, for godoc

pkg/config/default.go
```diff
@@ -14,8 +15,8 @@ import (
 
 const (
 	DefaultAPIVersion = "v1"
-	DefaultGPUdPort   = 15132
 	DefaultDataDir    = "/var/lib/gpud"
+	defaultGPUdPort   = 15132
```
gyuho:
```suggestion
	DefaultGPUdPort   = 15132
```

gyuho:
@aimichal The different port for listener is already supported by `--listen-address` flag, but some places you are changing are NOT using the flag value (as you correctly attempted to fix in this PR). Can we fix that instead, to use `--listen-address` flag value everywhere, rather than introducing the env variable?

gyuho:
@aimichal I am closing due to inactivity. Please re-open an issue to discuss this.

PR #1242, opened 2026-05-07
fix(containerd): detect NVIDIA runtime under containerd v2 plugin path and via imports

martbhell:
The nvidia runtime check searches /etc/containerd/config.toml for the literal `plugins."io.containerd.grpc.v1.cri".containerd.runtimes.nvidia`, which is the containerd 1.x plugin path. containerd 2.x uses `io.containerd.cri.v1.runtime`, so the check raises a false positive on every v2 host even when the nvidia runtime is configured correctly.

Two changes:

  - Accept either plugin path (1.x grpc.v1.cri or 2.x cri.v1.runtime).
  - Follow `imports = [...]` directives in the main config so drop-in files (e.g. /etc/containerd/conf.d/99-nvidia.toml written by nvidia-container-toolkit-daemonset on recent gpu-operator versions) are inspected too. Read errors on individual imports are ignored so the existing substring checks still run against whatever could be read.

Adds unit tests for the v2 plugin path, the new constant, and the import-following helper.

The error:
```
May 07 12:01:00 cluster-2 gpud[3478]: {"level":"warn","ts":"2026-05-07T12:01:00.934Z","caller":"log/log.go:175","msg":"nvidia GPUs found but nvidia-container-toolkit pod is not found"}
```

- [ ] Marking as draft I also haven't tested this (yet) and this is claude coding.

With this kind of containerd setup:

```
$ containerd --version
containerd containerd.io v2.2.3 77c84241c7cbdd9b4eca2591793e3d4f4317c590
$ find /etc/containerd/ -type f
/etc/containerd/config.toml
/etc/containerd/config.d/root-on-nvme.toml

$ find /etc/containerd/ -type f|xargs cat
imports = ["/etc/containerd/config.d/*.toml", "/etc/containerd/conf.d/*.toml"]
version = 3

[metrics]
  address = "0.0.0.0:1338"
  grpc_histogram = true
root = "/mnt/local_disk/containerd"
```

### DIFF
--- a/components/containerd/component.go
+++ b/components/containerd/component.go
@@ -7,6 +7,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


gyuho (approved):
lgtm!

PR #1245, opened 2026-05-21
[LEP-5065] feat(nfs): detect NFS hangs from kmsg and flip node Unhealthy

calmkart:
## Summary

  Adds a kmsg-based detection path to the `nfs` component that catches NFS
  hangs the existing read/write prober cannot — specifically the case where
  the NFS client kernel module is spinning on its own writeback locks while
  kubelet's main thread keeps reporting Ready.

  When a hang is detected, the component flips to `Unhealthy` with a
  `SuggestedActions{REBOOT_SYSTEM}`, which lets cluster-operator's
  auto-fixer cordon and reboot the node automatically. Repeated hangs
  across reboots are upgraded to `HARDWARE_INSPECTION` via the existing
  `eventstore.EvaluateSuggestedActions` machinery, so this does not
  introduce a reboot loop.

  ## Motivation

  A worker node ran for 3.5 days in a state where:

  - kubelet's main goroutine kept writing logs and reporting heartbeats
    (node `Ready=True`)
  - one CPU core was 100% pinned inside `nfs_lock_and_join_requests` →
    `_raw_spin_lock`, triggered by a user-issued `mksquashfs` writing a
    large file to `/nfs/home`
  - any goroutine touching the filesystem (kubelet's pod admit / volume
    mount / cAdvisor) blocked forever
  - new pods could be scheduled to the node but never admitted

  dmesg reported `watchdog: BUG: soft lockup` every ~28 seconds from the
  first 26-second stuck event onward, but nothing was listening for it.
  The current `components/cpu` matcher already recognises `soft lockup`
  text but only records it as an event — it never affects health state.

  This change makes that signal actionable, scoped to NFS so the
  generic `soft lockup` semantics in `components/cpu` are unchanged.

  ## What's in this PR

  Two commits, each independently buildable and testable:

  1. **`feat(nfs): add kmsg matcher and hang evaluator for NFS hang detection`**
     Pure additions, no wiring:
     - `kmsg_matcher.go`: four NFS-specific regexes (server not responding,
       server OK recovery, lock reclaim failed, NFS writeback functions
       in stack traces) + a `Match` function for `kmsg.Syncer`.
     - `hang_evaluator.go`: `collectNFSHangEvents()` applies conjunction
       rules — lock-reclaim failed counts unconditionally, "not responding"
       is cancelled by a later "OK", writeback stack hints require ≥ 2
       occurrences in the lookback window (which, combined with the
       300-second dedup, means ≥ 5 minutes of sustained hang and filters
       out one-shot sysrq dumps).

  2. **`feat(nfs): wire kmsg-based hang detection into component health`**
     Hooks the detector into the component:
     - New fields `eventBucket`, `kmsgSyncer`, `rebootEventStore`,
       `lookbackPeriod` (3 days, aligned with `eventstore.DefaultRetention`).
     - `newComponent()` testable inner constructor; `kmsg.Syncer` started
       on `linux + euid==0 + EventStore!=nil` with
       `WithCacheKeyTruncateSeconds(300)`.
     - `Check()` evaluates kmsg events before the prober. If a hang is
       present, query reboot history, sort ascending (see note below),
       and let `EvaluateSuggestedActions` decide whether to suggest
       reboot or hardware inspection. If a reboot already happened after
       the hang, fall through to the prober for live re-check rather than
       reporting stale state.
     - `Start()` now runs `Check()` immediately on the first iteration
       instead of waiting one tick, matching `components/disk`.
     - `Close()` shuts the syncer before the bucket.
     - `Events()` returns bucket contents instead of `nil`.
     - `checkResult.suggestedActions` surfaced through `HealthStates()`.

  ## Notable design points

  - **Ordering quirk**: `Bucket.Get` and `RebootEventStore.GetRebootEvents`
    both return events in **descending** order (latest first), but
    `eventstore.EvaluateSuggestedActions` expects **ascending** input
    (see `pkg/eventstore/suggested_actions_test.go` case 4 where
    `failureEvents[0]` is the earliest). This PR sorts ascending before
    calling. `components/disk` happens to pass the descending slice
    directly — that may be a latent issue worth addressing in a separate
    PR but is out of scope here.
  - **Lookback = 3 days** (not longer), because `EventStore` doesn't support
    per-bucket retention overrides and the global default is 3 days. Within
    this window we get accurate reboot-vs-hang ordering; beyond it we'd
    just be reading nothing.
  - **kmsg `Message` preserves leading whitespace** (parsed by
    `strings.SplitN(line, ";", 2)` in `pkg/kmsg/watcher.go`), so the
    writeback regex uses `(^|\s)…` to tolerate the leading space typical
    of kernel stack trace lines.

  ## Testing

  ```text
  go test ./components/nfs/ -short -race -count=1
  ok  github.com/leptonai/gpud/components/nfs  1.7s

  56 tests pass (all pre-existing + 25 newly added across matcher,
  evaluator, and component layers). Highlights:

  - Reboot-loop guard: TestCheckKmsgHangWithRebootLoop — two hangs
  with two interleaved reboots upgrades to HARDWARE_INSPECTION.
  - No-loop safety: TestCheckKmsgHangAfterReboot — hang followed by
  a later reboot returns nil from EvaluateSuggestedActions and falls
  through to the prober.
  - Robustness: TestCheckNilRebootEventStore — nil reboot store
  does not panic.
  - Probe skip: TestCheckSkipsProberWhenHangDetected — confirms the
  kmsg path returns before a probe is issued on a known-hung mount.
  - Cold-start latency: TestStartImmediateFirstCheck — Start()
  runs Check() before the first ticker fire (also fixes pre-existing
  behaviour where the first health state would only appear after a
  full minute).

  Risk & limitations

  - Repeated sysrq within 5 minutes could trigger a false positive
  (two writeback stack hints needed). This is a controlled human
  action and the resulting cordon is a safe side-effect.
  - gpud's own probe goroutines are still vulnerable to NFS hangs
  that occur mid-execution: the existing 5-second ctx.WithTimeout
  cannot cancel an in-kernel NFS syscall. This PR mitigates by
  skipping new probes on known-hung mounts but does not rescue
  already-stuck goroutines. Pre-existing; out of scope.
  - components/cpu is intentionally untouched: its existing
  soft lockup matcher still only writes events and does not affect
  health, since the generic CPU lockup signal is too noisy to act on
  unconditionally.

  Manual verification on a real node

  # 1. Inject "server not responding" (no OK to cancel) -> Unhealthy within 60s
  echo 'nfs: server 192.168.1.100 not responding, timed out' | sudo tee /dev/kmsg
  sleep 65
  curl -sk https://localhost:15132/v1/states | jq '.[] | select(.name=="nfs")'

  # 2. Single writeback stack hint -> still Healthy (below threshold)
  echo ' nfs_lock_and_join_requests+0x61/0x2b0 [nfs]' | sudo tee /dev/kmsg
  sleep 65

  # 3. Second writeback hint 5+ minutes later (defeats dedup) -> Unhealthy
  sleep 305
  echo ' nfs_lock_and_join_requests+0x82/0x2b0 [nfs]' | sudo tee /dev/kmsg
  sleep 65
  # Expect: .health == "Unhealthy"
  #         .suggested_actions.repair_actions == ["REBOOT_SYSTEM"]
  #         .suggested_actions.description == "NFS hang requires immediate reboot; drain may hang on NFS volumes"

### DIFF
--- a/components/nfs/component.go
+++ b/components/nfs/component.go
@@ -246,7 +246,15 @@ func (c *component) Check() components.CheckResult {
 			if len(hangEvents) > 0 {
 				var rebootEvents eventstore.Events
 				if c.rebootEventStore != nil {
-					rebootEvents, _ = c.rebootEventStore.GetRebootEvents(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+					var err error
+					rebootEvents, err = c.rebootEventStore.GetRebootEvents(c.ctx, cr.ts.Add(-c.lookbackPeriod))
+					if err != nil {
+						cr.err = err
+						cr.health = apiv1.HealthStateTypeUnhealthy
+						cr.reason = "failed to get reboot events"
+						log.Logger.Warnw(cr.reason, "error", cr.err)
+						return cr
+					}
 					// GetRebootEvents returns events in DESCENDING order
 					// (latest first), but EvaluateSuggestedActions expects
 					// ASCENDING order (oldest first) — see design doc §9 U2.
--- a/components/nfs/component.go
+++ b/components/nfs/component.go
@@ -261,18 +261,23 @@ func (c *component) Check() components.CheckResult {
 					sort.Slice(rebootEvents, func(i, j int) bool {
 						return rebootEvents[i].Time.Before(rebootEvents[j].Time)
 					})
+					if len(rebootEvents) > 0 {
+						hangEvents, hangReason = collectNFSHangEvents(nfsEventsOnOrAfter(evs, rebootEvents[0].Time))
+					}
 				}
-				sa := eventstore.EvaluateSuggestedActions(rebootEvents, hangEvents, 2)
-				if sa != nil {
-					sa.Description = "NFS hang requires immediate reboot; drain may hang on NFS volumes"
-					cr.health = apiv1.HealthStateTypeUnhealthy
-					cr.reason = hangReason
-					cr.suggestedActions = sa
-					return cr
+				if len(hangEvents) > 0 {
+					sa := eventstore.EvaluateSuggestedActions(rebootEvents, hangEvents, 2)
+					if sa != nil {
+						sa.Description = "NFS hang requires immediate reboot; drain may hang on NFS volumes"
+						cr.health = apiv1.HealthStateTypeUnhealthy
+						cr.reason = hangReason
+						cr.suggestedActions = sa
+						return cr
+					}
 				}
-				// sa == nil → a reboot was already observed after the
-				// hang event; fall through to the prober to verify the
-				// mount is currently healthy.
+				// No unresolved post-reboot hang evidence, or sa == nil
+				// because reboot history already resolved the hang; fall
+				// through to the prober for live verification.
 			}
 		}
 	}
--- a/components/nfs/component.go
+++ b/components/nfs/component.go
@@ -6,6 +6,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"runtime"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


gyuho (approved):
lgtm

PR #1247, opened 2026-06-01
feat(infiniband): opt-in recovery window for IB port flaps

martbhell:
IB port flap events are surfaced unconditionally and stay Unhealthy until an operator runs `gpud set-healthy` (or the 5-day event retention purges them). We run slurm and a NHC health checker to check healthiness of the gpud components.
During cluster bring-up/validation a port that flapped while the fabric was settling then stabilized keeps the node drained even though the link is up, forcing manual intervention.

I have seen this behavior once or twice now with cx8 & q3400.

Add an opt-in `--infiniband-flap-sticky-window` (default 0 = unchanged, always-sticky behavior). When set > 0, flaps follow the same recovery-window logic as drops: a flapping port that has been stably recovered (thresholds passing) for longer than the window stops being surfaced, so the node can auto-recover.

Recovery is keyed off thresholdRecoveryTime, not the flap event timestamp: the store stamps a flap at the snapshot where it first breached the flap-count threshold, which for a continuously-flapping port stays old, so "flap age > window" would wrongly clear a still-flapping port. A port that keeps flapping keeps dipping below thresholds, which resets the recovery timer, so it stays surfaced; only a stably-ACTIVE port clears.

### DIFF
--- a/cmd/gpud/command/command.go
+++ b/cmd/gpud/command/command.go
@@ -241,8 +241,8 @@ sudo rm /etc/systemd/system/gpud.service
 					Usage: "set the infiniband expected port states in JSON (leave empty for default, useful for testing)",
 				},
 				&cli.DurationFlag{
-					Name:  "infiniband-flap-sticky-window",
-					Usage: "opt-in recovery window for InfiniBand port flaps (e.g. 10m), analogous to the built-in drop sticky window; 0 (default) keeps flaps sticky until 'gpud set-healthy'",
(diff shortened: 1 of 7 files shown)


gyuho:
@martbhell The change makes sense, and the code/implementation look good to me, but the flag is a bit confusing

Do we have any better name than `--infiniband-flap-sticky-window`?

Something like

> --infiniband-flap-auto-clear-window

?

martbhell:
Something like this?

gyuho (approved):

PR #1251, opened 2026-06-16
fix(pkg/log): incorrect caller in log wrapper

JoonsooWon:
### What changed
- zap.AddCallerSkip(1) to `CreateLoggerWithConfig` to prevent the caller field from pointing to 'log/log.go'
- Left `CreateLoggerWithLumberjack` unchanged, caller-free behavior for file logging

### Why
When building and running the application locally using the default console logger, all log entries incorrectly point to `log/log.go`, making it impossible to trace code

```
make all
./bin/gpud run
```

Before
Screenshot 2026-06-17 at 2 12 10 AM
After
Screenshot 2026-06-17 at 2 13 42 AM

This happens because `gpudLogger` wraps the underlying `zap.SugaredLogger`

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -72,7 +72,7 @@ func CreateLoggerWithConfig(config *zap.Config) *gpudLogger {
 		config = DefaultLoggerConfig()
 	}
 
-	l, err := config.Build()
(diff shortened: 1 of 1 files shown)


gyuho (approved):
lgtm!

PR #1252, opened 2026-06-24
feat(session): support configurable reboot commands

gyuho:
## Summary
- add gpud run --reboot-commands and thread RebootCommands through config, server, and session setup
- preserve the default pkghost.Reboot(ctx, WithDelaySeconds(10)) behavior when the flag is empty
- make the Helm DaemonSet opt into host-namespace reboot via nsenter with documented hostPID/privileged/root requirements

## Validation
- go test -vet=off ./pkg/config ./pkg/session -run 'Test(DefaultConfig|WithRebootCommands|Session_processRequest|Session_configureRebootCommands|Session_runRebootCommandsWithConfiguredCommands|Session_triggerConfiguredRebootCommands|RunConfiguredRebootCommands|MockeyOpOption_WithRebootCommands)' -count=1
- go test -run '^$' ./pkg/config ./pkg/session ./pkg/server ./cmd/gpud/command ./cmd/gpud/run -count=1
- go vet ./pkg/config ./pkg/session ./pkg/server ./cmd/gpud/command ./cmd/gpud/run
- helm template gpud deployments/helm/gpud --namespace gpud --show-only templates/daemonset.yaml
- git diff --check

## Notes
- A broader local go test over the touched packages without the repository's mockey gcflags hit existing full-suite mockey/runtime-test behavior; GitHub Actions tests-unit is the authoritative full unit signal for the PR.

### DIFF
--- a/pkg/session/session_process_request.go
+++ b/pkg/session/session_process_request.go
@@ -235,6 +235,7 @@ func runConfiguredRebootCommands(ctx context.Context, commands string) error {
 		proc,
 		process.WithReadStdout(),
 		process.WithReadStderr(),
+		process.WithWaitForCmd(),
 		process.WithProcessLine(func(line string) {
 			log.Logger.Infow("configured reboot commands output", "line", line)
 		}),
--- a/pkg/session/session_process_request.go
+++ b/pkg/session/session_process_request.go
@@ -4,9 +4,17 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strings"
+	"time"
 
 	pkghost "github.com/leptonai/gpud/pkg/host"
 	"github.com/leptonai/gpud/pkg/log"
+	"github.com/leptonai/gpud/pkg/process"
+)
+
+const (
+	rebootDelaySeconds = 10
+	rebootDelay        = rebootDelaySeconds * time.Second
 )
 
 // processRequest handles all request processing logic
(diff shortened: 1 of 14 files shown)


eahydra (approved):

pkg/session/session_process_request.go
```diff
@@ -139,6 +147,100 @@ func (s *Session) processRequest(ctx context.Context, reqID string, payload Requ
...
+	if err := proc.Start(ctx); err != nil {
+		return err
+	}
+	defer func() {
+		if err := proc.Close(ctx); err != nil {
+			log.Logger.Warnw("failed to close configured reboot commands", "error", err)
+		}
+	}()
+
+	return process.Read(
```
gyuho, line 233: This needs to wait for the command result, e.g. by passing `process.WithWaitForCmd()` into `process.Read`. `process.Read` only propagates the child process exit status when that option is set; otherwise a script like `echo failed; exit 7` drains output and returns nil. In the delayed reboot path that means a failed configured reboot command is logged as `configured reboot commands completed`, and with the Helm default both host reboot attempts can fail without GPUd recording the failure. A focused test with `runConfiguredRebootCommands(ctx, "echo failed; exit 7")` expecting an error would catch this.

gyuho (commented):
Found one blocking issue in configured reboot command error handling, so I am not merging this yet.

PR #1256, opened 2026-06-29
fix(gpud): back off session reconnects

calmkart:
## Summary

  - Add structured reconnect signals for gpud session reader, writer, and health-check failures.
  - Parse `Retry-After` from 429/503 responses and use it as the lower bound for reconnect delay.
  - Replace fixed reconnect timing with exponential backoff plus jitter, including startup jitter.
  - Add jittered reconnect after `updateToken` so token rotation does not reconnect all agents at once.
  - Preserve auth failure persistence for 401/403 and token-validation 500 responses.

  ## Testing

  - `go test -gcflags="all=-N -l" ./pkg/session`
  - E2E single-machine replacement on worker node `fargate-ip-10-50-115-131.ec2.internal`
  - Verified real gpud service health, states, components, metrics, and control/worker machine status
  - Verified fake manager scenarios for health 503, reader 503, writer 429, auth 500, and token update

### DIFF
--- a/pkg/session/session_test.go
+++ b/pkg/session/session_test.go
@@ -189,6 +189,12 @@ func TestStartWriterAndReader(t *testing.T) {
 
 	// Initialize testable functions with controlled implementations to make
 	// goroutine lifecycles deterministic during the test.
+	s.jitterFunc = func(max time.Duration) time.Duration {
+		if max == startupJitterMax {
+			return 0
+		}
+		return max
+	}
 	s.timeAfterFunc = func(d time.Duration) <-chan time.Time {
 		// Block reconnection attempts after the first session so we can
 		// explicitly control shutdown ordering in the test.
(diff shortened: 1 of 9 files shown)


pkg/session/token.go
```diff
@@ -57,11 +57,15 @@ func (s *Session) processUpdateToken(payload Request, response *Response) {
 	log.Logger.Infow("token updated successfully in database and memory")
 
 	// Immediately close the current session to force reconnection with the new token
-	log.Logger.Infow("closing current session to immediately reconnect with new token")
+	delay := tokenReconnectJitterMin + s.jitter(tokenReconnectJitterMax-tokenReconnectJitterMin)
```
gyuho, line 60: tiny wording nit: this no longer happens immediately — it waits 2–30s. Maybe say “schedule a reconnect”?

pkg/session/session_test.go
```diff
@@ -738,17 +744,17 @@ func TestSessionKeepAlive(t *testing.T) {
 	originalStartWriter := s.startWriter
 
 	var exitMu sync.Mutex
-	var readerExitChans []<-chan any
-	var writerExitChans []<-chan any
+	var readerExitChans []<-chan reconnectSignal
```
gyuho, line 753: This test needs `jitterFunc` too. `timeAfterFunc` never fires, so the new startup jitter blocks before reader/writer start; the test then passes without exercising the session. Returning 0 for `startupJitterMax` (like the test above) should keep the old coverage.

gyuho (approved):

eahydra (approved):

pkg/session/token.go
calmkart: Good catch. Updated the comment and log wording to say that we schedule the reconnect, since the current session is now closed after the 2-30s jitter delay.

pkg/session/session_test.go
calmkart: Fixed. TestSessionKeepAlive now injects jitterFunc so startupJitterMax returns 0, and it also asserts that reader/writer goroutines actually start before waiting for their exit channels. Verified with the targeted tests and the full pkg/session test using the required Mockey gcflags.

PR #1264, opened 2026-07-06
feat(session): manage KAP mTLS agent credentials

cardyok:
## Summary

- keep only the KAP mTLS status and credential-update session commands; kubeconfig is now fixed during Lepton bootstrap
- validate each per-machine certificate, identity, endpoint, and CA fingerprint before writing an immutable credential generation
- atomically switch the `current` symlink, restart the agent, and acknowledge the update only after `/readyz` succeeds
- remove enable/disable state, kubeconfig mutation, pending markers, nested TLS preflight, and automatic rollback
- keep certificate and private-key material redacted from session audit logs

## Rollout

- the companion Lepton MR [dgxc-lepton/lepton!7510](https://gitlab-master.nvidia.com/dgxc-lepton/lepton/-/merge_requests/7510) installs the agent by default on new nodes and points kubelet at loopback from bootstrap
- existing nodes are not backfilled; Lepton reconciles only Machines that report the `kap-mtls-agent` package
- release this gpud change before or with the Lepton node-init rollout; failures leave new nodes blocked before kubelet registration

## Verification

- `GOFLAGS=-mod=mod go test -count=1 ./pkg/kapmtls`
- `GOFLAGS=-mod=mod go test -count=1 -gcflags='all=-N -l' ./pkg/session`
- `GOFLAGS=-mod=mod go test -race ./pkg/kapmtls`
- `GOFLAGS=-mod=mod go vet ./pkg/kapmtls ./pkg/session`

The repository-wide default `pkg/session` test command requires Mockey's `-gcflags='all=-N -l'` on macOS/arm64; the changed package scope passes with that required flag.

Jira: LEP-5336


pkg/kapmtls/manager.go
```diff
@@ -0,0 +1,691 @@
...
+	}
+	if _, err := m.runSystemctl(ctx, "enable", AgentService); err != nil {
+		return fmt.Errorf("enable KAP mTLS agent: %w", err)
+	}
+	if _, err := m.runSystemctl(ctx, "restart", AgentService); err != nil {
+		return fmt.Errorf("restart KAP mTLS agent: %w", err)
+	}
+	if !m.waitReady(ctx) {
+		return fmt.Errorf("KAP mTLS agent did not become ready")
+	}
```
eahydra, line 230: P1 — Failed activation is reported as successfully applied. UpdateCredentials switches current before enabling/restarting the service. If restart fails while the old agent remains alive, subsequent Status reports the new on-disk credentials together with AgentActive=true and AgentReady=true. The reconciler therefore considers the desired generation installed, although the running process still uses the old certificate/CA/endpoint. I reproduced this state. Track an applied generation only after readiness, roll back the symlink on activation failure, or verify the running certificate serial.

pkg/kapmtls/manager.go
```diff
@@ -0,0 +1,691 @@
...
+	return m.removeInactiveReleases(releaseID)
```
eahydra: P2 — Concurrent managers can delete the release referenced by current. After readiness, removeInactiveReleases(releaseID) cleans up using the caller’s stale release ID without synchronization. If two sessions overlap, manager B can select release B, then manager A’s delayed cleanup deletes B while current still points to it. I reproduced the resulting dangling symlink. Serialize the complete update across managers sharing a state directory and re-read the actual current target before cleanup.

eahydra (commented):
defer to @gyuho

gyuho (approved):

PR #1270, opened 2026-07-14
feat(session): add single-connection transport

cardyok:
## Summary

- add the v2 bidirectional gRPC session transport while preserving the legacy v1 transport
- default new sessions and service installations to `auto`: try v2 first and fall back to v1 when the v2 service is absent, including gRPC `UNIMPLEMENTED` and the legacy Envoy HTTP 464 response
- receive concrete protobuf request packets, preserve JSON results, and bound messages and handshake deadlines with gRPC transport keepalive
- preserve explicit session protocol selection across `gpud up` and auto-update, and reconnect either transport after token rotation
- send the control-plane `origin` metadata required by the gateway CSRF filter

## Rollout

New installations default to `auto`. They use the single-connection v2 transport when the matching Lepton gateway and gpud-manager support is available, and fall back to v1 when v2 is explicitly unsupported or rejected by a legacy gateway. Existing services that already persist `--session-protocol=v1` remain on v1 until reinstalled or explicitly reconfigured. Explicit `v1` and `v2` modes remain available.

Related Lepton MR: https://gitlab-master.nvidia.com/dgxc-lepton/lepton/-/merge_requests/7763

## Tested cases

- legacy GPUd v0.12.11 and the branch GPUd in forced v1 mode each maintain two v1 streams
- default auto mode falls back to two healthy v1 streams behind a legacy gateway and reaches Ready/Healthy
- forced v2 and auto modes each maintain exactly one bidirectional gRPC stream with a v2-capable control plane
- a default-auto node reaches Ready/Healthy on the single v2 stream
- v2 request/result traffic and periodic scrapes complete successfully
- v2 update-token audit data is redacted, and shutdown racing with request delivery exits without panic while queue-full backpressure remains intact
- deleting the owning gpud-manager pod causes the single stream to reconnect to a new owner, after which commands continue successfully
- upgrading an already Joined legacy GPUd under a v2-capable control plane replaces the two v1 streams with one v2 session; control-plane logs confirm the single session and successful scrapes for more than eight minutes, beyond the six-minute Offline threshold
- upgrading an already Joined legacy GPUd under a legacy control plane logs the automatic fallback and re-establishes v1 read/write sessions; control-plane logs confirm successful scrapes for more than nine minutes, beyond the six-minute Offline threshold

## Jira

[LEP-5352](https://jirasw.nvidia.com/browse/LEP-5352)


pkg/session/v2/session.proto
```diff
@@ -0,0 +1,67 @@
...
+}
+
+message HelloAck {
+  uint32 protocol_revision = 1;
+  string manager_instance_id = 2;
+  uint32 max_receive_message_bytes = 3;
+  uint32 heartbeat_interval_seconds = 4;
+}
+
+message Command {
```
eahydra: We have implemented new protocol based on gRPC, therefore, it's better to define concrete type for each command.

pkg/session/v2/session.proto
```diff
@@ -0,0 +1,67 @@
...
+  bytes payload_json = 2;
+  int64 timeout_millis = 3;
+}
+
+message CommandResult {
+  string request_id = 1;
+  bytes payload_json = 2;
+}
+
+message Heartbeat {
```
eahydra: Do we still need it as I thought gRPC has supported keepalive or heartbeart?

pkg/session/session_v2_adapter.go
```diff
@@ -0,0 +1,244 @@
...
+			legacy.CustomPluginSpecs = pluginSpecsFromV2(command.SetPluginSpecs.Specs)
+		} else if len(command.SetPluginSpecs.Specs) != 0 {
+			return Request{}, fmt.Errorf("plugin specs are present without a specs payload")
+		}
+	case *sessionv2.Request_UpdateToken:
+		if command.UpdateToken == nil {
+			return Request{}, fmt.Errorf("update-token command is missing")
+		}
+		legacy.Method = "updateToken"
+		legacy.Token = command.UpdateToken.Token
```
gyuho: can we make sure we redact this in audit logs?

pkg/session/session_v2.go
```diff
@@ -0,0 +1,353 @@
...
+			if err != nil {
+				return reconnectSignal{side: reconnectSideSingle, reason: "invalid v2 request", err: err}
+			}
+			data, err := json.Marshal(request)
+			if err != nil {
+				return reconnectSignal{side: reconnectSideSingle, reason: "failed to encode v2 request", err: err}
+			}
+			body := Body{ReqID: payload.Request.RequestId, Data: data}
+			select {
+			case s.reader <- body:
```
gyuho: can this panic, when we `Session.Stop`?

pkg/session/session_v2_adapter.go
cardyok: Yes. The v2 packet is converted to the legacy Request representation and then goes through the existing session audit path. Before logging, auditSessionRequestData recursively redacts token fields as well as the KAP certificate/private-key fields. TestAuditSessionRequestDataRedactsCredentials and TestAuditSessionRequestDataRedactsCaseVariants cover this.

pkg/session/session_v2.go
cardyok: Yes, there was a shutdown race here. The receive path now uses tryWriteV2Request, which checks context cancellation and converts a send-on-closed-channel race into context.Canceled instead of allowing it to panic. TestTryWriteV2RequestHandlesShutdownAndBackpressure explicitly calls Session.Stop before the write and covers this case.

eahydra (commented):
lgtm defer to @gyuho @calmkart

gyuho (approved):

PR #1274, opened 2026-07-16
Potential fix for code scanning alert no. 1: Workflow does not contain permissions

gyuho:
Potential fix for [https://github.com/leptonai/gpud/security/code-scanning/1](https://github.com/leptonai/gpud/security/code-scanning/1)

Add an explicit `permissions` block at the workflow root (top-level, alongside `name` and `on`) so all jobs inherit least-privilege defaults unless overridden.  
For this workflow, the minimal safe baseline is:

- `contents: read`

This preserves current functionality (checkout/build/upload artifact) while constraining token access and satisfying CodeQL.

Change location:

- File: `.github/workflows/build.yml`
- Insert `permissions:` after the trigger block (`on:` section) and before `jobs:`.

No imports, methods, or dependencies are needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches: ["**"]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,4 @@
-name: Build gpud binary
+name: Build and Store gpud binary
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


cardyok:
The explicit `permissions: contents: read` change looks correct, but removing the artifact upload is unrelated to the security fix and changes the workflow's output contract. The PR description also says the existing checkout/build/upload behavior is preserved.

The workflow run for the first commit (`41000b12`), which added `contents: read` while retaining `actions/upload-artifact`, completed successfully and produced the `gpud` artifact. Could we restore the upload step here? If artifact removal is intentional, it would be clearer to handle that separately with its consumers and behavior documented.

cardyok (approved):

PR #1275, opened 2026-07-18
fix(docker): address CVE-2026-45447

gyuho:


cardyok:
Two concerns before merging this security fix:

1. The Dockerfile supports both Ubuntu 22.04 and 24.04, but the fail-closed version check only covers Jammy. Ubuntu lists Noble's fixed OpenSSL version as `3.0.13-0ubuntu3.11`; a Noble build can currently succeed without proving that `openssl` and `libssl3t64` contain the fix. Please add a Noble version check alongside the Jammy check.
2. None of the current PR checks build the changed Dockerfile. The Docker image workflow runs only for tags, so package-name, repository, shell, and architecture failures would first appear during release. Please build at least the supported 22.04 and 24.04 image variants before merge, ideally with amd64 and arm64 coverage.

Ubuntu reference: https://ubuntu.com/security/CVE-2026-45447

cardyok (approved):

PR #1277, opened 2026-07-21
feat(providers): add OCI IMDSv2 detector

gyuho:


cardyok:
I found three issues worth addressing before merge:

1. **OCI provider metadata is incomplete.** `oci.New()` only wires provider detection and private IPv4, so OCI logins always report an empty `providerInstanceID` and no provider region. The control plane persists `providerInstanceID`, and every other cloud detector supplies an instance ID. OCI IMDSv2 exposes both `/instance/id` and `/instance/canonicalRegionName`; can we wire those through `InstanceID` and `RegionDetector` as part of this detector?

2. **The private-IP endpoint is fetched twice without retry.** `detectProvider` calls `FetchPrimaryVNICPrivateIPv4`, then `providers/all.Detect` calls `PrivateIPv4` again after detection. A transient failure or throttle on the second request leaves the machine identified as OCI but drops the primary VNIC IP. Oracle explicitly documents periodic IMDS downtime/throttling and recommends retrying 404/429/5xx responses. Please either cache the successful detection result or use a separate stable identity endpoint for detection, and add bounded retry/backoff for the metadata fetch.

3. **`gpud scan` ignores its caller context during the new provider lookup and makes the scan tests non-hermetic.** `Scan(ctx)` calls `machineinfo.GetProvider("")`, but `GetProvider` creates a new context from `context.Background()`. On hosts where metadata endpoints blackhole, Ctrl-C or a scan deadline cannot interrupt the sequential provider probes. Also, only `TestScan_Success` mocks `GetProvider`; the other scan tests that reach this line now access real metadata services. Please add a context-aware provider lookup (or inject it into `Scan`) and mock it consistently in scan unit tests.

The OCI endpoint/header usage itself matches Oracle's IMDSv2 documentation, and the focused OCI/provider tests pass locally with the repository's Go 1.25.12 toolchain and required Mockey gcflags.

eahydra (approved):

gyuho:
@cardyok Addressed all three points in [`93fd84ff`](https://github.com/leptonai/gpud/commit/93fd84ffdffe710775a5733dba0bc67cbb09ee9b) after rebasing onto the latest `main` (`c682518c`).

1. OCI now reads `/instance/id` and `/instance/canonicalRegionName`, exposed through `InstanceID` and `RegionDetector`.
2. Provider detection now uses the instance ID, so `/vnics/0/privateIp` is fetched once during metadata collection. OCI IMDS retries `404`, `429`, and `5xx` responses up to three times with bounded exponential backoff. Caller cancellation interrupts the wait.
3. `Scan(ctx)` now passes its context into provider detection through `GetProviderWithContext`. Scan unit tests inject the provider lookup and no longer call live metadata endpoints. The existing opt-in integration test still uses the real lookup.

Post-rebase normal tests, race tests, and `go vet` pass for `pkg/providers/oci/...`, `pkg/providers/all`, `pkg/machine-info`, and `pkg/scan`.

cardyok (approved):

PR #1302, opened 2026-08-18
[LEP-6152] feat(machine-info): report GPU fabric ClusterUUID, CliqueID, and chassis serial in MachineInfo

gyuho:
Surface ClusterUUID and CliqueID from the NVML fabric state API and the chassis serial number from nvidia-smi -q in the MachineInfo payload sent with login and gossip requests, so that gpud-manager can record them in the Machine CR status.

- api/v1: add clusterUUID, cliqueID, chassisSerial to MachineInfo and render them in gpud machine-info output
- pkg/machine-info: collect fabric identifiers from the first GPU with readable fabric state; collect chassis serial via nvidia-smi
- pkg/nvidia/nvidiasmi: new package parsing chassis serial number from nvidia-smi -q output


cardyok:
The fabric identifier path reads well: taking the first device with a readable fabric state is right, and the comment explains why the values are machine-wide.

**`GetChassisSerial` shells out to `nvidia-smi -q` on every collection.** The chassis serial never changes for the life of a node, and `nvidia-smi -q` is not a cheap call on NVL72-class systems: it walks every GPU to produce the full query output, which is exactly the hardware where this field is populated. Worth reading once and caching, or confirming the collection interval makes the repeated cost acceptable.

**The `nvidia-smi` invocation has no timeout of its own.** It inherits whatever ctx the caller passes. `nvidia-smi` is one of the first things to hang when the driver is wedged, which is a state gpud exists to report, so a bounded context here keeps a driver hang from stalling machine-info collection.

cardyok (approved):

PR #1304, opened 2026-08-19
[LEP-6173] fix(nvml): load GPU Operator managed driver

gyuho:


cardyok:
The resolution order is the right shape, and I like that `resolveNVMLLibraryPath()` returning empty preserves go-nvml's own lookup: a systemd install sees no behaviour change at all, and the test pins that.

**`mountNVIDIADriverRoot: true` as the default is the part worth a second look.** `DirectoryOrCreate` keeps the DaemonSet schedulable everywhere, but it also means every node running this chart gets `/run/nvidia/driver` created, including preinstalled-driver nodes that will never have anything there. Defaulting to `false` and opting in on GPU Operator clusters keeps the chart from touching hosts that do not need it. If the default stays `true`, it is worth saying in the values comment that the directory is created on nodes that do not use it.

**The candidate paths are a fixed list.** `/usr/lib/<arch>`, `/usr/lib64`, `/usr/lib` covers the current GPU Operator layout, so a layout change means a code change rather than a config change. `GPUD_NVML_LIBRARY_PATH` is a reasonable escape hatch for that, so this is a note rather than a request.

cardyok (approved):

PR #1305, opened 2026-08-19
[LS-1740] fix(nfs): require exact mount targets

gyuho:


cardyok:
Both findings are about the non-exact path this PR leaves behind.

**`FindMntTargetDevice` has no production callers left.** `components/nfs/component.go` was the only one, and it now uses `FindMntExactTargetDevice`. Leaving an exported function whose documented behaviour is "at or below the target" invites the next caller to hit the same bug this PR fixes.

**The prefix branch still has an unfixed edge.** `strings.HasPrefix(mountPoint, dir)` has no boundary check, so `dir="/lustre"` matches `/lustre-backup` as well as `/lustre/fs1`. If the function stays, the comparison wants to be:

```go
mountPoint == dir || strings.HasPrefix(mountPoint, dir+"/")
```

Either delete it or fix it, but a half-correct matcher on the exported surface is the part that will bite someone later.

cardyok (approved):

PR #1307, opened 2026-08-20
feat(session): add nodeCredentials for control-plane credential delivery

cardyok:
## What

Adds a `nodeCredentials` session method so the control plane can place credential material on a node without shipping a shell script for it.

The destination of each file travels in the request:

```json
{
  "method": "nodeCredentials",
  "node_credentials": {
    "kubelet": [
      {"path": "/var/lib/gpud/packages/kubelet/kubelet.yaml", "contents": "...", "mode": 420},
      {"path": "/var/lib/gpud/packages/kubelet/kubelet-client-current.pem", "contents": "..."}
    ]
  }
}
```

so adding a credential kind, or moving one, does not need a GPUd release.

## What it does not do

It writes files and nothing else: no process starts, no package installs, no service restarts. Whoever consumes the material decides when to act on it. That keeps the call safe to retry, and safe against a node whose state the caller cannot see.

## Three properties the tests pin

**Destinations are bounded.** `/var/lib/gpud` and `/etc/kubernetes` only, compared after `filepath.Clean` so a path spelled with `..` cannot leave an allowed tree. Caller-specified paths are what make the method general; without the bound it is an arbitrary root-owned write, and a control plane that is wrong or compromised could replace a unit file or an `authorized_keys`.

**All-or-nothing.** Every file is validated before any file is written, so one bad path leaves the node exactly as it was. The test lists the good file first, so a per-file validate-then-write loop would fail it.

**Atomic and never briefly exposed.** Each file is staged in its destination directory (same filesystem), given its final mode before any contents are written, flushed, then renamed into place. A reader sees the old contents or the new ones, never a partial write, and a private key is never readable under a permissive umask. The umask test runs at `umask 000`.

## Shape

`NodeCredentialsRequest` groups by the subsystem that owns the credential rather than flattening the fields, so a new kind is a new field instead of a change to an existing one. `updateKAPMTLSCredentials` can move under it later without disturbing what is here.

The whole `node_credentials` subtree is redacted from the audit log — redacting the subtree rather than naming fields keeps that true as fields are added.

## Testing

`go build ./...` and `go vet ./pkg/session` are clean.

Package tests do not run on my machine: `github.com/bytedance/mockey@v1.4.4` does not compile against the local Go toolchain, which breaks every package that imports it, on `main` as well as here. I verified the logic by running the file against an isolated module — path allow-list (6 rejections including `..` traversal and a `/var/lib/gpud-evil` prefix look-alike), empty contents, `0600` under `umask 000`, atomic replace with no staging residue, and the all-or-nothing property. CI runs the real tests.

### DIFF
--- a/pkg/session/kap_mtls.go
+++ b/pkg/session/kap_mtls.go
@@ -104,6 +104,14 @@ func redactSessionCredentials(value any) {
 				}
 				continue
 			}
+			if strings.EqualFold(key, "node_credentials") {
+				// Every file under here is credential material, and the
+				// per-subsystem grouping means new kinds appear without this
+				// function being touched. Redact the whole subtree rather than
+				// naming fields that will drift.
+				typed[key] = "<redacted>"
+				continue
+			}
 			redactSessionCredentials(child)
 		}
 	case []any:
--- a/pkg/session/node_credentials.go
+++ b/pkg/session/node_credentials.go
@@ -44,22 +44,39 @@ type NodeCredentialFile struct {
 	Mode uint32 `json:"mode,omitempty"`
 }
 
+// KubeletCredentials is what a kubelet needs to register. The fields are named
+// rather than a list, so each file's role is part of the contract instead of
+// something this side has to infer from a path.
+type KubeletCredentials struct {
+	// Config is the kubelet configuration, which carries the providerID.
+	Config *NodeCredentialFile `json:"config,omitempty"`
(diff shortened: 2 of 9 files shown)


eahydra (commented):
lgtm, defer to @gyuho @calmkart

calmkart (commented):
/LGTM

calmkart (approved):

PR #1309, opened 2026-08-21
[LEP-6075] feat(nvsentinel): support NVSentinel health data via gRPC sink

gyuho:


pkg/nvsentinel/source.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	subs := make([]chan HealthEvent, 0, len(s.subs))
+	for _, ch := range s.subs {
+		subs = append(subs, ch)
+	}
+	s.mu.Unlock()
+
+	for _, ch := range subs {
+		select {
+		case ch <- ev:
```
calmkart: **Blocking:** `record` adds the event to `recent` before any subscriber accepts it, and this non-blocking send can drop it. The RPC still returns success, so a full channel, a missing subscriber, or a later event-store insert failure can leave no durable event. A subsequent kmsg check still sees `Covers == true` and suppresses the native copy, permanently losing the incident. Coverage needs to represent successful component persistence, or the RPC must fail whenever delivery cannot be guaranteed.

components/accelerator/nvidia/xid/component_nvsentinel.go
```diff
@@ -0,0 +1,265 @@
...
+		return false
+	}
+
+	deviceUUID := xidErr.DeviceUUID
+	if uuid := convertBusIDToUUID(deviceUUID, c.devices); uuid != "" {
+		deviceUUID = uuid
+	}
+
+	return c.nvsSource.Covers(c.nvsDedupWindow, func(ev nvsentinel.HealthEvent) bool {
+		evXid, evUUID, ok := c.matchNVSentinelXid(ev)
```
calmkart, line 231: **Blocking:** this coverage predicate accepts healthy events, while `onNVSentinelEvent` explicitly discards every event with `IsHealthy == true`. A healthy event carrying XID N therefore enters the recent index, stores nothing, and can suppress a real native XID N during the dedup window. Please exclude healthy events from coverage, or implement explicit cancellation semantics that cannot hide a new native incident.

pkg/nvsentinel/source.go
```diff
@@ -0,0 +1,245 @@
...
+	s.nextSubID++
+	s.subs[id] = ch
+	s.mu.Unlock()
+
+	unsubscribe := func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		if sub, ok := s.subs[id]; ok {
+			delete(s.subs, id)
+			close(sub)
```
calmkart: `record` snapshots subscriber channels under the mutex and sends after releasing it. `unsubscribe` can close one of those snapshotted channels here before the send, causing `send on closed channel` and crashing gpud. Please synchronize subscriber lifetime with in-flight broadcasters, or avoid closing a channel that a broadcaster may still own. The same race exists when `Close` closes all subscriber channels.

components/accelerator/nvidia/sxid/component_nvsentinel.go
```diff
@@ -0,0 +1,234 @@
...
+		evSXid, _, ok := c.matchNVSentinelSXid(ev)
+		if !ok {
+			return false
+		}
+		// SXid events describe an NVSwitch link. The kmsg device UUID does
+		// not appear in the NVSentinel entity list as a UUID, so the
+		// comparison uses only the SXid number. Both detectors read the
+		// same kernel report — a repeated number within the window is
+		// the same incident.
+		return evSXid == sxidErr.SXid
```
calmkart: Matching only the SXID number is not sufficient on a multi-switch node: two NVSwitches can emit the same code within the dedup window. The NVSentinel event already carries `PCI`, `NVSWITCH`, and `NVLINK` entities, while the native error carries the switch PCI identifier. Please correlate at least the normalized PCI/switch identity, with a deliberate fallback only when identity is unavailable.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
```diff
@@ -0,0 +1,186 @@
...
+	ctx, cancel := context.WithTimeout(c.ctx, 5*time.Second)
+	events, err := c.eventBucket.Get(ctx, ts.Add(-c.nvsDedupWindow))
+	cancel()
+	if err != nil {
+		log.Logger.Warnw("failed to check stored infiniband events", "error", err)
+		return false
+	}
+
+	for _, stored := range events {
+		if stored.Name == eventName && !stored.Time.After(ts.Add(c.nvsDedupWindow)) {
```
calmkart: This node-wide match conflicts with the existing per-device policy for `access_reg_failed`. Two NVSentinel ACCESS_REG failures from `mlx5_0` and `mlx5_1` within the window have the same event name, so the second device is silently suppressed even though `EventKeyNVSentinelNICDevice` is stored. Please include NIC identity for ACCESS_REG dedup, while preserving the intentional fallback for events without a device.

pkg/gpud-manager/controllers/package_controller.go
```diff
@@ -303,13 +342,17 @@ func (c *PackageController) statusRunner(ctx context.Context) {
 		case <-ticker.C:
 			ticker.Reset(c.syncPeriod)
 		}
-		for _, pkg := range c.packageStatus {
-			if !pkg.IsInstalled {
+		for _, pkg := range c.packageStatusSnapshot() {
+			c.RLock()
+			installed := pkg.IsInstalled
+			scriptPath := pkg.ScriptPath
```
calmkart, line 348: This snapshot is used for `shouldSkip` and `status`, but the later `stop` and `start` commands still read `pkg.ScriptPath` directly without the lock. `reconcileLoop` can update that field concurrently under the write lock, so those residual reads are a data race. Please use this local `scriptPath` for both restart commands as well.

calmkart (commented):
Second-pass review against this PR head and the current remote main. Focused package tests pass, including the XID suite with its required Mockey compiler flags, but they do not cover the boundary sequences below. The first two inline comments are merge-blocking correctness issues; the remaining comments are confirmed concurrency or device-correlation defects.

pkg/nvsentinel/source.go
gyuho: Addressed in 00375fba. Coverage now means successful component persistence, not RPC receipt: the receive path only broadcasts, and XID/SXID components call RecordCoverage only after the event-store insert succeeds (or when a durable twin already exists). A full or missing subscriber and an insert failure therefore leave no coverage, so native kmsg detection is not suppressed. Added regression coverage for subscriber drops and insert failures.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
gyuho: Addressed in 00375fba. InfiniBand stored-twin detection now includes EventKeyNVSentinelNICDevice for ACCESS_REG events, so failures from mlx5_0 and mlx5_1 within the same window remain separate. The existing name-only fallback is preserved when either event has no device identity. Added per-device regression coverage.

pkg/gpud-manager/controllers/package_controller.go
gyuho: Addressed in 00375fba. Both restart commands now use the scriptPath snapshot captured under RLock; neither stop nor start reads pkg.ScriptPath again after releasing the lock. This removes the remaining race with reconcileLoop updates. The touched controller tests pass under the race detector.

components/accelerator/nvidia/sxid/component_nvsentinel.go
gyuho: Addressed in 00375fba. SXID coverage now normalizes and compares the native switch PCI address (for example, PCI:0000:05:00.0) with the NVSentinel PCI entity (0000:05:00.0). It falls back to SXID-number matching only when either side lacks a PCI identity. Added multi-switch regression coverage for equal SXID codes from different PCI addresses.

pkg/nvsentinel/source.go
gyuho: Addressed in 00375fba. Broadcast now holds the source mutex while performing non-blocking sends, so unsubscribe and Close cannot close a subscriber channel during an in-flight send. The sends never wait while holding the lock. Added a race regression test covering concurrent broadcast, unsubscribe, and Close; it reproduces the old send-on-closed-channel path and passes with this change.

components/accelerator/nvidia/xid/component_nvsentinel.go
gyuho: Addressed in 00375fba. Healthy events are excluded in Source.RecordCoverage, and the XID/SXID/InfiniBand coverage predicates also reject them explicitly. A healthy NVSentinel event can no longer suppress a later native incident. Added regression tests for healthy-event exclusion.

pkg/nvsentinel/source.go
```diff
@@ -0,0 +1,273 @@
...
+func (s *platformConnectorServer) HealthEventOccurredV1(_ context.Context, events *datamodels.HealthEvents) (*emptypb.Empty, error) {
+	receivedAt := time.Now().UTC()
+	for _, pbEv := range events.GetEvents() {
+		ev := healthEventFromProto(pbEv, receivedAt)
+		log.Logger.Debugw("received nvsentinel health event",
+			"agent", ev.Agent, "checkName", ev.CheckName,
+			"componentClass", ev.ComponentClass, "isHealthy", ev.IsHealthy)
+		s.src.record(ev)
+	}
+	return &emptypb.Empty{}, nil
```
calmkart, line 410: **Blocking:** The new coverage index prevents an unpersisted event from suppressing the native path, but this RPC still acknowledges delivery before any relevant component has accepted or persisted the event. `record` can drop on a full channel or deliver to zero subscribers, yet a nil response causes the NVSentinel gRPC sink to mark the queue item complete and skip retry. This is a permanent loss in the supported no-kmsg configuration and for XID/SXID codes rejected by the native catalog matcher. Please make successful component acceptance part of the RPC result, or enqueue into an acknowledged internal queue before returning success.

components/accelerator/nvidia/sxid/component_nvsentinel.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		var detail sxidErrorEventDetail
+		if err := json.Unmarshal([]byte(raw), &detail); err != nil {
+			continue
+		}
+		storedSXid, ok := intFromUint64(detail.SXid)
+		if !ok || storedSXid != sxidNum {
+			continue
+		}
+		if deviceUUID != "" && detail.DeviceUUID != "" && detail.DeviceUUID != deviceUUID {
```
calmkart: **Blocking:** The PCI fix only applies to `nvsentinelCoversSXid`. This stored-twin predicate still compares `deviceUUID`, which `matchNVSentinelSXid` derives from `GPU_UUID`, rather than the NVSwitch PCI identity. Two switches reporting the same SXID can therefore be treated as twins when they map to the same GPU, or when either UUID is empty. The twin branch then calls `RecordCoverage` for the second switch although its event was never stored, allowing its native report to be suppressed. Persist and compare normalized switch PCI/NVSWITCH identity here, and add an ingestion-level test that sends both NVSentinel events through `onNVSentinelEvent`.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
```diff
@@ -0,0 +1,217 @@
...
+		// A healthy event carries no new data point: it must never
+		// suppress a native incident. The source already excludes healthy
+		// events from coverage; this guard keeps the intent explicit here.
+		if ev.IsHealthy {
+			return false
+		}
+		if ev.CheckName != checkNameSyslogsNICDriverError || len(ev.ErrorCodes) == 0 {
+			return false
+		}
+		return ev.ErrorCodes[0] == pattern
```
calmkart: **High:** ACCESS_REG remains node-wide on the native-versus-NVSentinel dedup path. Coverage from `mlx5_0` matches an ACCESS_REG kmsg from `mlx5_1` because this predicate compares only the pattern. The native normalized message retains the PCI BDF, while NVSentinel provides the NIC name; correlate those identities for `access_reg_failed`. When they cannot be compared, retain the native event rather than suppressing a potentially distinct device incident.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	for _, stored := range events {
+		if stored.Name != eventName || stored.Time.After(ts.Add(c.nvsDedupWindow)) {
+			continue
+		}
+		if eventName == eventAccessRegFailed && device != "" {
+			// Both sides must name the same NIC; a stored event without a
+			// device (e.g. a native kmsg copy) keeps the name-only match.
+			if storedDevice := stored.ExtraInfo[EventKeyNVSentinelNICDevice]; storedDevice != "" && storedDevice != device {
```
calmkart: Treating a missing `EventKeyNVSentinelNICDevice` as a twin is not safe for native ACCESS_REG events. Native events do not populate this NVSentinel-specific key, but their normalized message can contain the PCI BDF. A stored native event from device A can therefore cause an NVSentinel event from device B to be skipped and then marked covered. Extract or map the stored BDF when present; if identity remains unavailable, do not claim a durable per-device twin.

pkg/nvsentinel/source.go
gyuho: Addressed in cae61843. The RPC now acknowledges against a bounded internal receive queue instead of a best-effort broadcast. HealthEventOccurredV1 enqueues each event and returns success only once the event is in the queue; a dispatcher drains it in order and delivers to every subscriber, waiting on a full subscriber channel instead of dropping, and holding the event when no subscriber has registered yet (components still starting). When the queue is full the RPC fails with codes.Unavailable, so the platform connector keeps the item and retries rather than marking it complete. New tests cover the queue-full failure, blocked-subscriber delivery, zero-subscriber hold, and the unsubscribe/Close race.

components/accelerator/nvidia/sxid/component_nvsentinel.go
gyuho: Addressed in cae61843. The twin check no longer compares the GPU UUID. onNVSentinelEvent persists the normalized PCI entity as nvsentinel_switch_pci, and hasStoredSXidTwin compares that switch identity against both stored formats — legacy kmsg copies carry PCI:<BDF> in device_uuid, NVSentinel copies carry the new key. Two switches reporting the same SXid with the same GPU_UUID (or an empty one) are now stored as distinct incidents, so the twin branch can no longer record coverage for an event that was never stored. When either side lacks a PCI address, the check falls back to the number-only match, same rule as nvsentinelCoversSXid. Added an ingestion-level test that pushes both switches' events through onNVSentinelEvent and asserts both are stored, plus a check that a redelivery for the same switch still twins.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
gyuho: Addressed in cae61843. For access_reg_failed the coverage predicate now extracts the PCI BDF from the native normalized message and maps the NVSentinel event's NIC name to its BDF via /sys/class/infiniband/<name>/device (resolved once, cached for the boot). Suppression only happens when both sides resolve to the same BDF. If the native line has no BDF or the NIC does not resolve, the native event is retained.

components/accelerator/nvidia/infiniband/component_nvsentinel.go
gyuho: Addressed in cae61843. A stored event without the NVSentinel NIC key is no longer an automatic twin: for native kmsg copies the check extracts the BDF from the stored normalized message and compares it with the incoming NIC's resolved BDF, and when that identity cannot be determined it claims no twin, so the incoming event is stored instead of skipped-and-covered. NVSentinel-stored copies still compare NIC names directly, and the name-only fallback now applies only when the incoming event itself names no device. Updated the old fallback test to assert the store-don't-skip behavior and added coverage for the BDF-correlated twin and distinct-device cases.

calmkart (approved):

PR #1311, opened 2026-08-26
[LEP-6173] fix(machine-info): detect GPU product via PCI when NVML is unavailable (GPU Operator-managed driver)

gyuho:
GPUd derives the GPU product name exclusively from NVML (nvmlDeviceGetName). On BYOK nodes where the NVIDIA driver is installed and managed by the GPU Operator, the NVML userspace library is not available to the gpud container during initial node discovery -- the Operator injects its toolkit only into GPU workload containers, and the driver install itself races gpud startup on a freshly added node. With no NVML, the no-op instance reports an empty product, so the control plane never learns the accelerator type (e.g., H100) and never creates the default GPU shape (e.g., gpu.h100-sxm).

Fall back to PCI sysfs enumeration when NVML yields no product name: scan /sys/bus/pci/devices for NVIDIA 3D/VGA controllers (kernel-side, needs no NVIDIA driver) and map the PCI device ID to the NVML-style product name, sanitized identically to the NVML code path. The NVML result always takes precedence when available.

Refs LEP-6173

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,3 +2,20 @@ ignore:
   - "e2e"
   - "pkg/session/v2/session.pb.go"
   - "pkg/session/v2/session_grpc.pb.go"
+
+coverage:
+  status:
+    # Project: coverage must stay within 0.25% of the base commit. The slack
+    # absorbs rounding noise when a PR adds mostly-covered files whose average
+    # sits slightly below the project average.
+    project:
+      default:
+        target: auto
+        threshold: 0.25%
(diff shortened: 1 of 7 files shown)


gyuho:
Superseded by #1314 and #1315. We are making a working NVIDIA driver an explicit prerequisite for registering NVIDIA GPU hardware: #1314 exposes and loads NVML correctly for both pre-installed and GPU Operator-managed drivers, and #1315 blocks login until NVML is ready while leaving CPU-only nodes unaffected. This removes the need to maintain a PCI device-ID-to-product-name fallback table.

PR #1314, opened 2026-08-31
[LEP-6173] feat(nvml): prefer pre-installed host driver, mount driver roots by default

gyuho:
c.f., https://github.com/leptonai/gpud/pull/1304
c.f., https://github.com/leptonai/gpud/pull/1311

Follow-up to #1304 (opt-in GPUD_NVIDIA_DRIVER_ROOT). Complementary to the NVML-free PCI product fallback in #1311, which covers the first-login driver-install race; this change restores full NVML monitoring once the driver exists, on both driver installation modes.

Library resolution order (pkg/nvidia/nvml/lib):
1. GPUD_NVML_LIBRARY_PATH (explicit, unchanged)
2. GPUD_NVIDIA_HOST_ROOT (new): host root fs mounted at /host, probing usr/lib/<arch>, usr/lib64, usr/lib -- a pre-installed host driver wins over the Operator driver root when both are present
3. GPUD_NVIDIA_DRIVER_ROOT (Operator-managed /run/nvidia/driver)
4. go-nvml default system lookup (unchanged when no env is set)

This mirrors the GPU Operator's DRA driver DaemonSet, which mounts /host and /driver-root and selects the pre-installed driver first at runtime.

Helm chart:
- gpud.mountHostRoot (new, default true): host root mounted read-only at /host with HostToContainer propagation, setting GPUD_NVIDIA_HOST_ROOT
- gpud.mountNVIDIADriverRoot: default flipped to true so GPU Operator clusters work out of the box; on preinstalled-driver nodes the DirectoryOrCreate hostPath may leave an empty, unused /run/nvidia/driver (set to false to avoid it)
- driver-root mount gains HostToContainer propagation so an Operator driver (re)install is visible without a pod restart

Backward compatible: with GPUD_NVIDIA_HOST_ROOT unset the resolution is identical to before.

### DIFF
--- a/pkg/nvidia/nvml/lib/options.go
+++ b/pkg/nvidia/nvml/lib/options.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	nvlibdevice "github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	nvinfo "github.com/NVIDIA/go-nvlib/pkg/nvlib/info"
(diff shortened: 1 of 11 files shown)


pkg/nvidia/nvml/lib/options.go
```diff
@@ -53,23 +60,42 @@ func (op *Op) applyOpts(opts []OpOption) {
...
 
-	driverRoot := os.Getenv(EnvNVIDIADriverRoot)
-	if driverRoot == "" {
-		return ""
+	// A pre-installed host driver takes precedence over a GPU
+	// Operator-managed driver root when both are present, mirroring the
+	// Operator's own driver-validation order (pre-installed wins when both
+	// validate). Loading the host's own libnvidia-ml guarantees the
+	// userspace library matches the loaded kernel module.
+	if hostRoot := os.Getenv(EnvNVIDIAHostRoot); hostRoot != "" {
```
eahydra: IMO, it's better to load the env from `/run/nvidia/validations/driver-ready`, which is generated by gpu-operator validation pod.

gyuho: Already fully addressed by commit c9e06ab4 ("discover GPU Operator driver root from driver-ready contract"), which post-dates the comment. I verified the implementation against the actual gpu-operator source (cmd/nvidia-validator): the contract is IS_HOST_DRIVER / NVIDIA_DRIVER_ROOT / DRIVER_ROOT_CTR_PATH / NVIDIA_DEV_ROOT / DEV_ROOT_CTR_PATH, and gpud's parsing matches it correctly:

- Reads <hostRoot>/run/nvidia/validations/driver-ready via the /host mount (the only path visible in-container; /run/nvidia/validations isn't mounted directly)
- Probes NVIDIA_DRIVER_ROOT resolved under the host root — covers non-default spec.hostPaths.driverInstallDir
- Skips / (host-driver selection, e.g. validator's hostRootFlag default) since host probes already ran
Correctly ignores DRIVER_ROOT_CTR_PATH (DRA container's mount path, not gpud's layout)

eahydra (approved):

PR #1317, opened 2026-09-04
[LEP-6439] fix(config): support "-name" denylist in gpud run --components

gyuho:
## Summary

Fixes LEP-6439. The `--components` flag documents `-name` entries as disabling individual components, but on v0.13.0-rc.1 (and 0.12.x) a denylist-only list disabled **all** components, and `--components=all,-foo` never disabled `foo`.

Found while hot-fixing AWS GPU Operator BYOK clusters (LEP-6440, [Slack thread](https://nvidia.slack.com/archives/C08PW10F3FC/p1788456735158699)), where we needed to silence exactly four noisy components (`accelerator-nvidia-fabric-manager`, `accelerator-nvidia-persistence-mode`, `library`, `docker`) and the documented denylist form would have taken down every health check.

## Root cause

1. `ShouldEnable` treated the list as a pure allowlist: `-`-prefixed entries were skipped, leaving `selectedComponents` empty — so `--components=-a,-b` disabled everything.
2. `ShouldDisable` stored denylist keys **with** the `-` prefix while `pkg/server` looks up **bare** component names — it never matched (dead code). The unit tests masked this by querying with prefixed names.
3. The `"all"`/`"*"` short-circuit returned before the denylist was consulted, so `all,-foo` did not disable `foo`.

## Fix

Parse `Components` once into allowlist + denylist maps, with the denylist stored under bare names:

| Configuration | Behavior |
|---|---|
| empty / `*` / `all` | all enabled (unchanged) |
| `none` (or any non-matching positive entry) | all disabled (unchanged, documented) |
| `comp1,comp2` | allowlist, only those enabled (unchanged) |
| `-comp1,-comp2` | **all enabled except comp1/comp2 (fixed)** |
| `all,-comp1` / `*,-comp1` | **all enabled except comp1 (fixed)** |
| `comp1,-comp2` | comp1 enabled, comp2 disabled, others disabled |

No interface changes; `pkg/server` registration call sites untouched.

## Test plan

- `go test ./pkg/config/... -count=1` — passes, including new regression coverage:
  - `TestConfig_Components_Denylist_LEP6439` uses the real BYOK component names
  - denylist-only, `all,-x`, `*,-x`, mixed allowlist+denylist, and `none` cases
  - updated the tests that encoded the buggy behavior (prefixed-name queries, early-return internals)
- `go build ./...` and `go vet ./pkg/config/` pass
- `pkg/server` test failures verified pre-existing on `main` (11 failures with and without this change; unrelated token/machine-ID/admin tests)

## Follow-ups (not in this PR)

- Once released, switch the affected-cluster hot-fix (helm `commandOverride` with a 28-name allowlist) to the denylist form: `--components=-accelerator-nvidia-fabric-manager,-accelerator-nvidia-persistence-mode,-library,-docker`
- Consider exposing a `gpud.components` value in the helm chart so this does not require `commandOverride`'s full wrapper replacement

Ref: LEP-6439, LEP-6440


pkg/config/config.go
```diff
@@ -195,61 +200,73 @@ func (config *Config) Validate() error {
 	return nil
 }
 
+// parseComponentSelectors lazily builds the allowlist and denylist lookup
+// maps from Config.Components. Entries prefixed with "-" form the denylist,
+// stored under their bare component names because callers (e.g. pkg/server)
+// look components up by bare name. All other entries form the allowlist,
+// where "*" and "all" select every component. A list containing only
+// "-"-prefixed entries enables every component except the denylisted ones.
+func (config *Config) parseComponentSelectors() {
```
gyuho, line 219: **Why the denylist is stored under bare names:** the only production caller, `pkg/server/server.go` (`shouldEnable := config.ShouldEnable(name)` / `config.ShouldDisable(name)` at ~line 358), passes the *bare* registered component name (e.g. `"library"`). The old code stored the map key as `"-library"` — prefix included — so `ShouldDisable("library")` could never match and the disable path was dead code. This was masked by the unit tests, which queried `ShouldDisable("-comp1")` with the prefix still attached instead of mirroring the production call. Evidence from LEP-6439 triage: on GPU Operator BYOK clusters there was no working way to disable individual components at all.

pkg/config/config.go
```diff
@@ -195,61 +200,73 @@ func (config *Config) Validate() error {
...
+	config.parseComponentSelectors()
 
-		for _, c := range config.Components {
-			if c == "*" || c == "all" {
-				// enable all components
-				return true
-			}
+	// the denylist wins in both allowlist and denylist-only modes:
+	// "--components=all,-foo" and "--components=-foo" both disable "foo"
+	if _, disabled := config.disabledComponents[componentName]; disabled {
```
gyuho, line 263: **Why the denylist is consulted before any allowlist logic:** the old `Components[0] == "*" || Components[0] == "all"` early return (and the identical short-circuit inside the map-init loop) bailed out before "-" entries were ever read, so `--components=all,-library` silently kept `library` enabled. Concrete case: the BYOK hot-fix for GPU Operator BYOK clusters needed exactly `all` minus four components; with the short-circuit, that combination was a no-op. Consulting `disabledComponents` first makes `all,-x` / `*,-x` behave the way the flag help text promises.

pkg/config/config.go
```diff
@@ -195,61 +200,73 @@ func (config *Config) Validate() error {
...
 
-			// prefix "-" is used to disable a component
-			if strings.HasPrefix(c, "-") {
-				continue
-			}
-			config.selectedComponents[c] = struct{}{}
-		}
+	// "all"/"*", or a denylist-only list ("--components=-foo,-bar"):
+	// everything not denylisted stays enabled
+	if config.allComponentsSelected || !config.hasComponentAllowlist {
```
gyuho, line 274: **Why a denylist-only list must default-enable:** the flag contract in `cmd/gpud/command/command.go` is "prefix component name with `-` to disable it" — i.e. `--components=-a,-b` means "everything except a and b". Previously this produced an *empty* allowlist, so `ShouldEnable` returned false for every component and `pkg/server` registered **zero** components. We know how bad that is from live evidence: on GPU Operator BYOK clusters we observed nodes running 13 vs 32 registered components (all 19 `accelerator-nvidia-*` missing) when GPU visibility differed — invisible health checks look identical to healthy ones from the control plane. The `!config.hasComponentAllowlist` branch preserves the documented escape hatch at the same time: `--components=none` is a positive, non-matching entry, so it still disables everything.

pkg/config/config_test.go
```diff
@@ -864,10 +970,74 @@ func TestConfig_WildcardAndAll_EdgeCases(t *testing.T) {
...
 	})
 }
+
+// TestConfig_Components_Denylist_LEP6439 is a regression test for LEP-6439:
+// "gpud run --components=-a,-b" must disable exactly the denylisted
+// components. Previously a denylist-only list disabled ALL components
+// (ShouldEnable treated the list as a pure allowlist and skipped "-" entries,
+// leaving the selected set empty), and ShouldDisable never matched because it
+// stored "-"-prefixed keys while pkg/server looks up bare component names.
+func TestConfig_Components_Denylist_LEP6439(t *testing.T) {
```
gyuho, line 990: **Why these exact component names:** they are the four components observed `Unhealthy` via `/v1/states` on the two p5.48xlarge GPU nodes of GPU Operator BYOK clusters (2026-09-03, gpud 0.13.0-rc.1): `accelerator-nvidia-fabric-manager` (nv-fabricmanager binary not on container PATH; present under `/run/nvidia/driver/usr/bin`), `accelerator-nvidia-persistence-mode` (off on all 8 H100s), `library` (libcuda/libnvidia-ml false negative — libs exist under the driver root), and `docker` (containerd node). The cluster is currently hot-fixed with a helm `commandOverride` carrying a 28-name allowlist; once this PR ships, that collapses to the denylist form asserted here. The positive-side assertions (`gpu-counts`, `error-xid`, `containerd`, `os`, `network-latency` stay enabled) guard against the old failure mode where a denylist-only list registered zero components.

deployments/helm/gpud/values.yaml
```diff
@@ -27,6 +27,21 @@ gpud:
...
+  #   select (allowlist): list bare component names; ONLY those run.
+  #     components: ["containerd", "os", "accelerator-nvidia-error-xid"]
+  #   disable (denylist): prefix names with "-"; everything else still runs.
+  #     components: ["-docker", "-library"]
+  #   disable everything: ["none"]
+  #
+  # The denylist ("-name") form requires a gpud image with the LEP-6439 fix;
+  # on older images a denylist-only list disables ALL components -- use an
+  # explicit allowlist there instead.
+  components: []
```
gyuho, line 43: **Why a list type with default empty:** empty list renders no `--components` flag at all, so stock behavior (all components enabled) is preserved byte-for-byte — verified with `helm template` (no `GPUD_COMPONENTS` env when unset). The denylist form (`["-docker"]` → `--components=-docker`) only works because of the `ShouldEnable`/`ShouldDisable` fix in this PR; on older images the same input disables ALL components, which is why the inline comment above carries the version requirement. Concrete use case: the GPU Operator BYOK clusters hot-fix currently runs a full 28-name allowlist via `commandOverride` — with this PR + chart value it becomes `components: ["-accelerator-nvidia-fabric-manager", "-accelerator-nvidia-persistence-mode", "-library", "-docker"]` (LEP-6440).

calmkart (approved):

cardyok (approved):

PR #1318, opened 2026-09-04
[LEP-6440] fix(components): report correctly on GPU Operator/containerized BYOK nodes

gyuho:
## Summary

Fixes LEP-6440. On containerized GPU Operator BYOK clusters (containerized gpud 0.13.0-rc.1, GPU Operator 26.3.2, p5.48xlarge H100 nodes), four components reported false `Unhealthy` after the 0.13.0-rc.1 driver-root mounts fixed GPU inventory — and the unhealthy status blocked Slurm scheduling. All findings below were verified live against the cluster ([Slack thread](https://nvidia.slack.com/archives/C08PW10F3FC/p1788456735158699)).

| Component | Reported | Verified reality |
|---|---|---|
| `library` | `libcuda.so`/`libnvidia-ml.so` "does not exist" | Both exist at `/run/nvidia/driver/usr/lib/x86_64-linux-gnu/` (Operator driver root, mounted by default since 0.13.0-rc.1) — gpud's own NVML loader already loads them from there |
| `accelerator-nvidia-fabric-manager` | "nv-fabricmanager executable not found" | Binary exists at `/run/nvidia/driver/usr/bin/nv-fabricmanager`; FM **process** is running (started by the driver DaemonSet with `fabricmanager.cfg`) but port 6666 is not listening on the host |
| `accelerator-nvidia-persistence-mode` | "all 8 GPU(s) disabled persistence mode" | True state — the Operator driver DaemonSet doesn't enable it; needs to be tolerable per deployment |
| `docker` | "docker installed but docker is not running" | Host has no docker at all (containerd 2.2.1 node); the CLI that triggered the check is **bundled in the gpud image itself** |

Also fixes the related observability gap: on nodes where NVML never loaded, `library` vacuously reported Healthy ("all libraries exist") — the same silence class as the original empty-`gpuInfo` incident.

## Changes

- **`pkg/nvidia/driverroot` (new)**: shared, existence-based probe helpers for the well-known driver roots (`/host`, `/run/nvidia/driver`) with library/bin dir expansion. No-ops on bare-metal/systemd installs.
- **`library`**: search dirs extended with the driver roots; when NVML is absent but PCI shows NVIDIA hardware, reports Unhealthy instead of vacuously passing (CPU-only nodes still skip cleanly).
- **`fabric-manager`**: `checkFMExists` probes driver-root `bin` dirs after PATH; `checkFMActive` accepts a running `nv-fabricmanager` process (hostPID `/proc` scan) when port 6666 is closed — covering FM-in-driver-container deployments.
- **`docker`**: `CheckDockerInstalledHostAware` determines installation from the host root mount when containerized; falls back to PATH on bare metal.
- **`persistence-mode`**: unchanged. PM-off on Operator nodes is genuinely true state (not a detection bug), so there is nothing to fix in detection — and no new flag: an earlier revision of this PR added `--persistence-mode-expected=any`, but it was dropped because PR #1317's denylist covers the operational need with first-class chart UX (`gpud.components: ["-accelerator-nvidia-persistence-mode"]`), while the flag had no chart knob and would have required the `commandOverride` path both PRs are eliminating.

The helm chart knob for component selection (`gpud.components`, list type, denylist syntax) shipped separately in PR #1317 with the LEP-6439 denylist fix.

## Test plan

- New/updated unit tests: `driverroot` path probes, `library` no-NVML gate (CPU/GPU/error/nil-detector), `checkFMExistsInRoots` (Operator layout, dir-not-file), `fabricManagerProcessRunning` (fake proc root with the exact argv observed on the cluster), `checkDockerInstalledOnHost` (host paths, fallback).
- `go build ./...`, `go vet`, `gofmt` clean; all touched packages' tests pass.
- Pre-existing failures on `main` (`pkg/server` ×11, `cmd/gpud/run` ×1) verified unchanged with this branch stashed.
- Chart unchanged in this PR (the `gpud.components` list value is in PR #1317).
- Post-merge validation plan: upgrade the target BYOK cluster, drop the temporary 28-name allowlist `commandOverride`, expect all-Healthy with `gpud.components: ["-accelerator-nvidia-persistence-mode"]` (the PR #1317 denylist) unless the platform enables PM.

Ref: LEP-6440, LEP-6439


components/library/component.go
```diff
@@ -105,6 +115,16 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
...
 			searchDirs[dir] = struct{}{}
 		}
+		// Also probe the well-known driver roots that the gpud DaemonSet mounts
+		// by default (helm values gpud.mountHostRoot and gpud.mountNVIDIADriverRoot).
+		// Without these, containerized gpud cannot see libnvidia-ml/libcuda installed
+		// by the GPU Operator (/run/nvidia/driver/usr/lib/...) or pre-installed on the
+		// host (/host/usr/lib/...), producing false "library does not exist" unhealthy
+		// reports even though gpud's own NVML loader already loads them from those
+		// trees (LEP-6440, observed on aws-iad-nkxdev-1 with GPU Operator 26.3.2).
+		for _, dir := range driverroot.LibraryDirs(driverroot.Existing()...) {
```
gyuho, line 129: **Why these extra search dirs:** verified on the target GPU Operator BYOK cluster (2026-09-03) — `libnvidia-ml.so`/`libcuda.so` exist at `/run/nvidia/driver/usr/lib/x86_64-linux-gnu/` (GPU Operator 26.3.2 driver tree, mounted into the gpud pod by default since chart 0.13.0-rc.1), and gpuds own NVML loader already loads them from there, yet the `library` component reported `library "libcuda.so" does not exist; library "libnvidia-ml.so" does not exist` because its search dirs only covered system paths. False Unhealthy on both p5.48xlarge nodes.

components/library/component.go
```diff
@@ -177,6 +197,33 @@ func (c *component) Check() components.CheckResult {
...
 	}()
 
+	// When NVML is absent, no libraries are configured for checking. On a
+	// CPU-only node that is correct ("nothing to check"), but on a node with
+	// NVIDIA GPU hardware it means gpud cannot monitor its GPUs at all --
+	// report it instead of vacuously passing with "all libraries exist".
+	// (LEP-6440: this silent gap is how the empty gpuInfo on aws-iad-nkxdev-1
+	// went unnoticed; the control plane recorded empty GPU fields with no
+	// unhealthy signal from gpud.)
+	if len(c.libraries) == 0 {
```
gyuho, line 212: **Why this gate exists:** on nodes where NVML never loaded, `c.libraries` stays empty and the old loop over zero entries reported Healthy `all libraries exist` — vacuously true. Observed live: the three m6i.xlarge CPU nodes on the target GPU Operator BYOK cluster reported exactly that. On a node WITH NVIDIA hardware this silence is the same failure class as the original empty-`gpuInfo` incident (gpud ran for days with blank GPU fields and no unhealthy signal). Now: CPU-only node → Healthy skip; GPU hardware present but NVML absent → Unhealthy. PCI detection errors stay log-and-skip so weird systems dont get noisy.

components/accelerator/nvidia/fabric-manager/component.go
```diff
@@ -553,12 +558,31 @@ func (c *component) hasNothingToDoEvent() bool {
...
+// PATH-only lookup reports a false "executable not found" (LEP-6440, observed
+// on aws-iad-nkxdev-1 with GPU Operator 26.3.2).
 func checkFMExists() bool {
 	p, err := exec.LookPath("nv-fabricmanager")
-	if err != nil {
-		return false
+	if err == nil && p != "" {
+		return true
+	}
+	return checkFMExistsInRoots(driverroot.Existing())
```
gyuho, line 575: **Why probe the driver roots:** the old `exec.LookPath("nv-fabricmanager")` only searches the containers PATH. Verified on the target GPU Operator BYOK cluster: binary absent from host system paths (`/host/usr/bin`, `/host/usr/sbin`) but present at `/run/nvidia/driver/usr/bin/nv-fabricmanager` inside the Operator driver tree — which the gpud DaemonSet mounts by default since 0.13.0-rc.1. The old check produced `fabric manager supported but nv-fabricmanager executable not found` on both H100 nodes despite the binary being visible inside the pod.

components/accelerator/nvidia/fabric-manager/component.go
```diff
@@ -580,7 +604,47 @@ func checkFMExists() bool {
...
+	}
+
+	// The port check alone is insufficient when the GPU Operator runs fabric
+	// manager inside its driver container: the FM process is alive (started by
+	// the driver DaemonSet with the nvswitch fabricmanager.cfg) but does not
+	// listen on port 6666 on the host (LEP-6440, observed on aws-iad-nkxdev-1:
+	// nv-fabricmanager pid present, :6666 closed). gpud's DaemonSet runs with
+	// hostPID, and on bare metal /proc is the host's, so scanning /proc covers
+	// both deployment modes.
+	return fabricManagerProcessRunning("/proc")
```
gyuho, line 623: **Why a process-based fallback for activeness:** port 6666 is only opened by a standalone host FM service. On Operator-managed nodes FM runs *inside the driver DaemonSet container*: verified on the target GPU Operator BYOK cluster — `nv-fabricmanager -c /usr/share/nvidia/nvswitch/fabricmanager.cfg` alive as host pid 39063, while `:6666` has no listener in `/proc/net/tcp` and the systemd unit is inactive. gpud runs hostPID (the DaemonSet needs it for nsenter-based reboots), so `/proc` here IS the hosts process table; on bare metal it is too. argv[0] basename match (not `/proc/<pid>/comm`, which is truncated to 15 bytes → `nv-fabricmanage`).

pkg/docker/docker.go
```diff
@@ -74,6 +76,38 @@ func CheckDockerInstalled() bool {
...
+// CheckDockerInstalledHostAware reports whether docker is installed on the
+// node. When gpud runs inside a container with the host root filesystem
+// mounted (the gpud DaemonSet's gpud.mountHostRoot, default on), a docker CLI
+// bundled in the container image itself must not count: the node only "has
+// docker" if the HOST has the binary. (LEP-6440: the gpud image ships
+// /usr/bin/docker, so the PATH-only check reported "docker installed but
+// docker is not running" on containerd-only nodes such as aws-iad-nkxdev-1.)
+// Without the host-root mount (bare-metal/systemd installs) it falls back to
+// the PATH lookup, preserving the previous behavior.
+func CheckDockerInstalledHostAware() bool {
```
gyuho, line 90: **Why host-aware detection:** the trigger on the target GPU Operator BYOK cluster was not host docker at all — the host has no docker binary (`/host/usr/bin/docker` absent, no `/var/run/docker.sock`) and runs containerd 2.2.1. The `docker` CLI that `LocateExecutable` found lives at `/usr/bin/docker` *inside the gpud container image itself*, so every containerd-only node running containerized gpud reported `docker installed but docker is not running` (all 5 nodes, CPU and GPU alike). With the host-root mount present (chart default since 0.13.0-rc.1), only host binaries count; without it (bare metal), the PATH behavior is unchanged.

components/accelerator/nvidia/persistence-mode/expectation.go
```diff
@@ -0,0 +1,47 @@
...
+	ExpectedModeEnabled ExpectedMode = "enabled"
+
+	// ExpectedModeAny records the persistence mode state but never reports
+	// unhealthy for disabled persistence mode. Intended for GPU
+	// Operator-managed clusters where no component enables persistence mode
+	// and long-running GPU clients (DCGM, device plugin) already keep the
+	// driver initialized (LEP-6440, observed on aws-iad-nkxdev-1 where the
+	// GPU Operator 26.3.2 driver DaemonSet leaves persistence mode off on all
+	// GPUs).
+	ExpectedModeAny ExpectedMode = "any"
```
gyuho, line 26: **Why an opt-out knob instead of just fixing the nodes:** disabled persistence mode here is *real* (all 8 H100s `supported=true, enabled=false`), but nothing on GPU Operator 26.3.2 turns it on — no nvidia-persistenced, no driver-container init. Meanwhile DCGM and the device plugin hold the driver initialized continuously, so PM-off is functionally benign on exactly the clusters where it is structurally off. `any` lets those deployments record the state without a permanent Unhealthy that blocks Slurm scheduling; default `enabled` keeps every existing installation behaving as before. The platform-side alternative (enable PM via the driver DaemonSet) is tracked in LEP-6440 as well.

pkg/nvidia/driverroot/driverroot.go
```diff
@@ -0,0 +1,77 @@
...
+	// mounts the GPU Operator's driver installation tree (helm value
+	// gpud.mountNVIDIADriverRoot).
+	OperatorRoot = "/run/nvidia/driver"
+)
+
+// Existing returns the well-known driver roots that currently exist,
+// in probe order: the host root first (a pre-installed host driver wins,
+// mirroring the GPU Operator's own driver-validation order), then the GPU
+// Operator driver root.
+func Existing() []string {
```
gyuho, line 42: **Why existence-based probing with host-root first:** mirrors the probe order in `pkg/nvidia/nvml/lib` (a pre-installed host driver wins over the Operator tree, matching the Operators own driver-validation order) and is a no-op when neither mount exists — so systemd/bare-metal installs (e.g. the PDX prod nodes where gpud runs as a host service) keep their standard system-path behavior with zero risk. This package is the shared home so `library` and `fabric-manager` stop hardcoding divergent path lists.

cardyok:
How will a Helm deployment set this option?

The main use case here is containerized GPU Operator/BYOK nodes, but the DaemonSet startup wrapper does not pass --persistence-mode-expected and values.yaml has no corresponding setting. Without replacing the whole wrapper through commandOverride, users will still get the default "enabled" behavior, so disabled persistence mode will continue to report Unhealthy.

Could we add something like gpud.persistenceModeExpected, defaulting to "enabled", and pass it through the normal startup wrapper?

cmd/gpud/command/command.go
```diff
@@ -328,6 +329,11 @@ sudo rm /etc/systemd/system/gpud.service
 					Name:  "nvlink-expected-link-states",
 					Usage: "set the nvlink expected link states in JSON (leave empty for default, useful for testing)",
 				},
+				&cli.StringFlag{
+					Name:  "persistence-mode-expected",
```
calmkart: The new persistence-mode policy is not configurable through the default Helm deployment path that this PR is intended to fix.

The CLI accepts `--persistence-mode-expected=any`, but the chart has no corresponding value and the startup wrapper never appends the argument. Rendering the chart with `gpud.persistenceModeExpected=any` produces no persistence-mode argument, so GPU Operator nodes still use the default `enabled` policy and continue reporting Unhealthy.

`commandOverride` can technically pass the flag, but it replaces the complete startup wrapper and requires callers to reproduce the token, machine ID, driver, reboot, disk, and runtime arguments.

Please add a validated Helm value and append `--persistence-mode-expected` to the normal wrapper, with a render test for both `enabled` and `any`.

pkg/nvidia/driverroot/driverroot.go
```diff
@@ -0,0 +1,88 @@
...
+	return ExistingFrom(HostRoot, OperatorRoot)
```
calmkart: Please keep this driver-root discovery consistent with the existing NVML resolver.

The NVML loader reads `/host/run/nvidia/validations/driver-ready` and supports a custom `NVIDIA_DRIVER_ROOT` such as `/opt/nvidia/driver`. This helper only returns `/host` and `/run/nvidia/driver`.

As a result, NVML can load successfully from `/host/opt/nvidia/driver/usr/lib/...`, while the library component still reports `libcuda`/`libnvidia-ml` missing and the Fabric Manager component cannot find `nv-fabricmanager`.

The chart documentation already says a custom GPU Operator `driverInstallDir` is discovered automatically. Could we expose and reuse the canonical resolved driver roots here instead of maintaining a second, fixed list?

components/accelerator/nvidia/persistence-mode/component.go
```diff
@@ -259,11 +264,21 @@ func (c *component) Check() components.CheckResult {
...
 		if len(notEnabled) == len(cr.PersistenceModes) {
-			cr.reason = fmt.Sprintf("all %d GPU(s) disabled persistence mode", len(devs))
+			reason = fmt.Sprintf("all %d GPU(s) disabled persistence mode", len(devs))
+		}
+
+		// ExpectedModeAny tolerates disabled persistence mode (e.g. GPU
+		// Operator-managed clusters where nothing enables it) while still
+		// recording the state; ExpectedModeEnabled keeps the historical
+		// unhealthy behavior (LEP-6440).
+		if c.expectedMode == ExpectedModeAny {
```
calmkart: `ExpectedModeAny` is applied after the Warning event has already been inserted.

The component runs every 30 seconds, and `eventstore.Find` includes `ev.Time.Unix()` in its exact match. The next check therefore does not find the previous event and inserts another Warning for every GPU, even though disabled persistence mode is explicitly tolerated.

On an 8-GPU node with the default 14-day retention, this can retain roughly 322,560 Warning rows per node.

Please either skip the Warning event in `any` mode or deduplicate it by state transition/device identity without including the current timestamp. The test should advance time by at least one ticker interval.

cmd/gpud/command/command.go
gyuho: Resolved by removing the flag instead: 485e7648 drops `--persistence-mode-expected` from this PR entirely (the branch no longer touches the persistence-mode component). Tolerating PM-off on GPU Operator nodes is handled by the denylist fixed in #1317, which has first-class chart support: `gpud.components: ["-accelerator-nvidia-persistence-mode"]` renders `--components=-accelerator-nvidia-persistence-mode` in the default startup wrapper — no `commandOverride` needed, and this PR no longer carries a knob the chart cannot express. The PR body validation plan was updated accordingly.

pkg/nvidia/driverroot/driverroot.go
gyuho: Addressed in 8ef810e9: `pkg/nvidia/driverroot` now owns the canonical ordered candidate list — host root, then any `NVIDIA_DRIVER_ROOT` selected by the driver-ready contract (resolved under the host root), then the Operator root — and `pkg/nvidia/nvml/lib` consumes it via `driverroot.Candidates` instead of maintaining its own contract parser. With a custom `driverInstallDir` (e.g. `/opt/nvidia/driver`), the library and fabric-manager components now search the same tree NVML loaded from. The contract parsing edge cases (missing contract, `"/"` host selection, relative-path rejection, quoting, multiple selections, existence filtering) moved to `driverroot` unit tests, and the resolver-level contract tests in `options_test.go` are preserved against the shared path.

components/accelerator/nvidia/persistence-mode/component.go
gyuho: The flag this concerned is dropped (485e7648) — `ExpectedModeAny` no longer exists and this file is byte-identical to main, so the tolerated-but-still-warning scenario is gone. On the dedup observation itself: confirmed, `findEvent` matches on timestamp+name+type(+message), so a persistent condition re-inserts an event every check interval; that behavior predates this PR and is shared by every component using the event store, so changing it belongs in a separate change. Happy to file a follow-up issue for transition/device-identity-based dedup if you would like it tracked.

gyuho:
@cardyok Good catch — resolved by removing the flag instead: 485e7648 drops `--persistence-mode-expected` entirely (the branch no longer touches the persistence-mode component). PM-off tolerance on GPU Operator nodes is handled by the #1317 denylist, which does have first-class chart support: `gpud.components: ["-accelerator-nvidia-persistence-mode"]` → `--components=-accelerator-nvidia-persistence-mode` in the default wrapper, no `commandOverride`. PR body updated to match.

calmkart (approved):

cardyok (approved):

PR #1319, opened 2026-09-05
fix(package-controller): recover installed packages before dependency checks

cardyok:
## Summary

- detect packages that are already installed before validating dependencies needed for a fresh install
- preserve dependency gates for missing, uninstalled, or version-incompatible dependencies
- cover restart recovery, dependency boundaries, and concurrent-install guards

## Why

Package controller state is rebuilt in memory when GPUd starts. Previously, a newly declared but unavailable dependency prevented `isInstalled` from running, so an existing package could remain `Unknown` and become ineligible for updates after a restart.

## Testing

- `GOFLAGS=-mod=mod GOTOOLCHAIN=go1.25.12 go test ./pkg/gpud-manager/controllers -run TestInstallRunner_WithMockedRunCommand -count=100 -shuffle=on`
- `GOFLAGS=-mod=mod GOTOOLCHAIN=go1.25.12 go test -race ./pkg/gpud-manager/controllers -run TestInstallRunner_WithMockedRunCommand -count=20 -shuffle=on`
- `GOFLAGS=-mod=mod GOTOOLCHAIN=go1.25.12 go test ./pkg/gpud-manager/controllers -count=5 -shuffle=on`
- `GOFLAGS=-mod=mod GOTOOLCHAIN=go1.25.12 go test -race ./pkg/gpud-manager/controllers -count=1`
- `GOFLAGS=-mod=mod GOTOOLCHAIN=go1.25.12 go build ./...`
- custom `golangci-lint v2.7.2`: 0 issues for `./pkg/gpud-manager/controllers/...`

### DIFF
--- a/pkg/gpud-manager/controllers/mock_package_controller_runner_test.go
+++ b/pkg/gpud-manager/controllers/mock_package_controller_runner_test.go
@@ -161,6 +161,7 @@ func TestInstallRunner_WithMockedRunCommand(t *testing.T) {
 			Name:           "dep",
 			IsInstalled:    true,
 			CurrentVersion: "2.0.0",
+			ScriptPath:     "/tmp/dep.sh",
 		}
 
 		controller.packageStatus["needs-missing-dep"] = &packages.PackageStatus{
@@ -175,19 +176,39 @@ func TestInstallRunner_WithMockedRunCommand(t *testing.T) {
 			ScriptPath: "/tmp/version.sh",
 			TotalTime:  200 * time.Millisecond,
 		}
+		controller.packageStatus["uninstalled-dep"] = &packages.PackageStatus{
+			Name:       "uninstalled-dep",
+			Dependency: [][]string{{"missing", "*"}},
+			ScriptPath: "/tmp/uninstalled-dep.sh",
+			TotalTime:  200 * time.Millisecond,
+		}
+		controller.packageStatus["needs-uninstalled-dep"] = &packages.PackageStatus{
+			Name:       "needs-uninstalled-dep",
+			Dependency: [][]string{{"uninstalled-dep", "*"}},
+			ScriptPath: "/tmp/needs-uninstalled-dep.sh",
+			TotalTime:  200 * time.Millisecond,
+		}
 		controller.packageStatus["skip"] = &packages.PackageStatus{
 			Name:       "skip",
 			ScriptPath: "/tmp/skip-install.sh",
 			TotalTime:  200 * time.Millisecond,
 		}
-		controller.packageStatus["installed"] = &packages.PackageStatus{
-			Name:       "installed",
-			ScriptPath: "/tmp/installed.sh",
+		controller.packageStatus["installed-with-missing-dep"] = &packages.PackageStatus{
+			Name:       "installed-with-missing-dep",
+			Dependency: [][]string{{"missing", "*"}},
+			ScriptPath: "/tmp/installed-with-missing-dep.sh",
 			TotalTime:  200 * time.Millisecond,
 		}
 		controller.packageStatus["install"] = &packages.PackageStatus{
 			Name:       "install",
+			Dependency: [][]string{{"dep", "2.0.0"}},
 			ScriptPath: "/tmp/install.sh",
+			TotalTime:  200 * time.Millisecond,
+		}
+		controller.packageStatus["installing"] = &packages.PackageStatus{
+			Name:       "installing",
+			Installing: true,
+			ScriptPath: "/tmp/installing.sh",
 			TotalTime:  200 * time.Millisecond,
 		}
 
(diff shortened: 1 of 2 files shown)


gyuho:
Moving `isInstalled` before the dependency check breaks an important controller invariant: **`IsInstalled == true` previously guaranteed that all declared dependencies were satisfied.**

Bypassing dependency checks when setting `IsInstalled = true` introduces several issues across the controller runners:

---

### 1. `updateRunner` Bypasses Dependencies (Premature Upgrades)
`updateRunner` only checks `if !installed { continue }` and does not validate `pkg.Dependency`. It relies on `installRunner` gating `IsInstalled` to ensure updates only happen when dependencies are satisfied.
- **Scenario:** Package `A` (v1.0.0) is installed. A new target version (v2.0.0) declares a new or updated dependency `B` (`#GPUD_PACKAGE_DEPENDENCY=B:2.0.0`) which is missing or uninstalled.
- **Result:** With this PR, `isInstalled` succeeds for `A` (since v1.0.0 is on disk), setting `A.IsInstalled = true`. `updateRunner` immediately detects `version != targetVersion` and executes `upgrade` on `A` without `B` being installed, leading to broken installations.

---

### 2. `statusRunner` Infinite Restart Loops
`statusRunner` also checks `if !installed { continue }` without checking dependencies.
- **Scenario:** Package `A` is present on disk, but its dependency `B` is missing or not installed (e.g. `kubelet` depending on `containerd` or `nvidia-driver`).
- **Result:** `A.IsInstalled` is set to `true`. `statusRunner` checks `status`, which fails because dependency `B` is not running. `statusRunner` then triggers `stop` and `start` every 3 seconds, putting `A` into an infinite restart loop.

---

### 3. Invalidation of Downstream Dependencies
- **Scenario:** Package `C` depends on `A`, and `A` depends on missing/uninstalled `B`.
- **Result:** Since `A` is marked `IsInstalled = true` via `isInstalled`, package `C`'s dependency check sees `A` as installed and proceeds to install `C`, violating transitive dependency requirements.

---

### 4. Excessive Subprocess Spawning for Blocked Packages
- **In `main`:** When an uninstalled package's dependency was not yet installed, `installRunner` skipped it in memory (0 subprocesses spawned).
- **In this PR:** For every uninstalled package waiting on a dependency, `installRunner` now executes both `shouldSkip` and `isInstalled` bash subprocesses every 3 seconds (`syncPeriod`). Each invocation creates and truncates `shouldSkip.log` and `isInstalled.log` on disk (`os.O_TRUNC`).

---

### Suggestions
If the goal is to discover existing on-disk packages across GPUd restarts so they are known:
1. `updateRunner` and `statusRunner` should explicitly validate `pkg.Dependency` before attempting `upgrade` or `status`/restart actions, OR
2. Decouple on-disk presence/version probing from `IsInstalled` state so that packages are not considered ready/active until their dependencies are satisfied.

pkg/gpud-manager/controllers/package_controller.go
```diff
@@ -261,6 +229,39 @@ func (c *PackageController) installRunner(ctx context.Context) {
 				continue
 			}
 
+			var skipInstall bool
```
gyuho, line 232: Moving the dependency check after `isInstalled` breaks the invariant that `IsInstalled == true` guarantees declared dependencies are met. This has several unintended side effects across the controller:

1. **`updateRunner` bypasses dependencies**: `updateRunner` only checks `if !installed { continue }` and never validates `pkg.Dependency`. If package `A` is present on disk and has an update to target version `v2.0.0` that declares a new/updated dependency `B`, `A.IsInstalled` is set to `true`, and `updateRunner` immediately attempts to `upgrade` `A` even when `B` is missing or uninstalled.
2. **`statusRunner` restart loops**: `statusRunner` monitors any package with `IsInstalled == true`. If a package cannot run properly without its dependency (e.g. `kubelet` missing `containerd` or `nvidia-driver`), its status check will fail, causing `statusRunner` to continuously `stop` and `start` the service every 3 seconds.
3. **Downstream dependencies unblocked**: Any downstream package `C` that depends on `A` will see `A.IsInstalled == true` and proceed to install, even though `A`'s own dependencies are missing.
4. **Redundant polling**: For uninstalled packages blocked on dependencies, moving `shouldSkip` and `isInstalled` before this check causes `installRunner` to spawn bash subprocesses and truncate/write log files every 3s instead of skipping in-memory.

pkg/gpud-manager/controllers/mock_package_controller_runner_test.go
```diff
@@ -262,7 +390,8 @@ func TestInstallRunner_WithMockedRunCommand(t *testing.T) {
 		require.Eventually(t, func() bool {
 			controller.RLock()
 			defer controller.RUnlock()
-			return controller.packageStatus["installed"].IsInstalled
+			installed := controller.packageStatus["installed-with-missing-dep"]
```
gyuho, line 393: Asserting that a package with a missing dependency is marked `IsInstalled = true` codifies the bypass: if dependency `missing` is absent, marking the package as installed allows `updateRunner` to upgrade it without dependencies, `statusRunner` to restart it when it fails, and dependents to install assuming it is ready.

pkg/gpud-manager/controllers/mock_package_controller_runner_test.go
```diff
@@ -271,18 +400,50 @@ func TestInstallRunner_WithMockedRunCommand(t *testing.T) {
 			return controller.packageStatus["install"].Progress == 100
 		}, 10*time.Second, 10*time.Millisecond)
 
+		require.Eventually(t, func() bool {
+			return missingProbeCalls.Load() > 0 &&
```
gyuho, line 404: `missingProbeCalls > 0` and `dependentProbeCalls > 0` show that `isInstalled` is now being executed every tick for packages whose dependencies are missing or uninstalled. In production, this causes continuous subprocess execution (`runCommand`) and log file truncation (`os.O_TRUNC`) every 3 seconds for every uninstalled package waiting on a dependency.

pkg/gpud-manager/controllers/mock_package_controller_runner_test.go
```diff
@@ -271,18 +400,50 @@ func TestInstallRunner_WithMockedRunCommand(t *testing.T) {
...
+		assert.Zero(t, oldVersionInstallCalls.Load())
+		assert.Zero(t, oldVersionStartCalls.Load())
+		assert.Positive(t, oldVersionProbeCalls.Load())
+		assert.Zero(t, uninstalledDepInstallCalls.Load())
+		assert.Zero(t, uninstalledDepStartCalls.Load())
+		assert.Positive(t, uninstalledDepProbeCalls.Load())
+		assert.Zero(t, dependentInstallCalls.Load())
+		assert.Zero(t, dependentStartCalls.Load())
+		assert.Positive(t, dependentProbeCalls.Load())
+		assert.Positive(t, installingShouldSkipCalls.Load())
```
gyuho, line 442: `installingShouldSkipCalls > 0` demonstrates that `shouldSkip` is called even when `Installing` is true (because `shouldSkip` is executed before `if installing { continue }`). If a package is actively installing in the background, running `shouldSkip` and potentially marking it as `Skipped = true`, `IsInstalled = true`, and `Progress = 100` can race with the active install goroutine.

pkg/gpud-manager/controllers/package_controller.go
```diff
@@ -198,38 +198,6 @@ func (c *PackageController) installRunner(ctx context.Context) {
...
-					log.Logger.Infof("[package controller]: %v dependency %v version %v does not meet required %v, skipping", pkg.Name, dep[0], depVersion, dep[1])
-					skipCheck = true
-					break
-				}
-			}
-			if skipCheck {
-				continue
-			}
 
 			var shouldSkipResult string
```
calmkart: Moving these probes before the dependency gate causes every dependency-blocked package to execute `shouldSkip` and `isInstalled` on every 3-second sync.

`installRunner` processes packages serially, and `runCommand` has no per-command timeout; it only observes the controller's root context. A single slow or hung `isInstalled` script can therefore block installation and recovery for every package that follows it.

Please give recovery probes a bounded timeout and avoid running them continuously while dependencies remain blocked. A test with one non-terminating probe followed by an otherwise installable package would cover the failure mode.

gyuho (approved):

PR #1320, opened 2026-09-07
fix(providers): survive slow first IMDS connect for private IP detection

gyuho:
## Problem

On nscale, the first TCP connection to the instance metadata service (169.254.169.254) after an idle period was observed taking 5.1-5.2s on production nodes — occasionally exceeding the 5s HTTP client timeout entirely — while warm connects complete in under a millisecond. When the private-IP fetch lost that race, the attempt was not retried and gpud reported an empty private IP in the login request, so downstream consumers (e.g., node machine-local-IP labeling) never received a value.

## Changes

- `pkg/providers/nscale/imds`: raise the IMDS HTTP client timeout from 5s to 10s so the cold first connect can complete. Scoped to nscale only; other providers' clients are unchanged.
- `pkg/providers`: new opt-in `WithPrivateIPv4Retry()` detector option (variadic on `NewIMDSWithRegion`, so existing constructors compile unchanged) with a `SupportsPrivateIPv4Retry` predicate mirroring the `SupportsIMDS` pattern. `detector.PrivateIPv4` is untouched.
- `pkg/providers/all`: the private IPv4 fetch goes through `fetchRequiredMetadata` — the same retry-with-backoff path already used for region and instance ID — **only for detectors that opted in (currently nscale)**. All other providers (aws, azure, gcp, nebius, oci) keep the exact single-attempt behavior from before this PR.
- `pkg/providers/all`: fetch order is now required-identity-first (instance ID, region), then best-effort fields (public IP, VM environment), with private IP last — an opted-in retry (4 attempts × 10s + backoff ≈ 40.7s worst case) can no longer starve the mandatory fields of the shared one-minute budget (review feedback).
- `pkg/providers/nscale`: the nscale detector opts into the retry, since its fabric IPs are not RFC1918 and a missed fetch has no NIC-based fallback.

## Tests

- `TestDetect_IMDSRetriesPrivateIPv4` (opted-in detector succeeds on 3rd attempt) and `TestDetect_IMDSPrivateIPv4RetryExhaustion` (4 attempts then empty).
- `TestDetect_IMDSPrivateIPv4SingleAttemptWithoutOptIn`: a non-opted-in detector with a failing fetcher is called exactly once — proven with a one-hour backoff that would hang the test if a retry were attempted.
- `TestDetect_PrivateIPv4RetryDoesNotStarveRequiredFields`: private fetch hangs until the shared context expires while region/instance-ID mocks fail on a dead context, pinning the required-first fetch order (verified to fail against the pre-reorder code).
- `TestSupportsPrivateIPv4Retry` covers the option on/off/nil-detector paths.
- `go vet`, `go test -gcflags="all=-N -l"`, and `-race -d=checkptr=0` pass for `./pkg/providers/...` and consumers (`pkg/machine-info`, `pkg/login`, `cmd/gpud/machine-info`, `cmd/gpud/up`).

### DIFF
--- a/pkg/providers/all/all.go
+++ b/pkg/providers/all/all.go
@@ -77,7 +77,12 @@ func DetectWithRegionOverride(ctx context.Context, regionOverride string) (*pkgp
 		info.PublicIP = publicIP
 	}
 
-	privateIP, err := detector.PrivateIPv4(ctx)
+	// Private IP goes through the same IMDS retry path as region and instance ID:
+	// on some providers (e.g., nscale) the first metadata request after an idle
+	// period can exceed the HTTP client timeout while the connection path warms
+	// up, and a single attempt would return empty even though a retry succeeds
+	// immediately.
+	privateIP, err := fetchRequiredMetadata(ctx, detector, "private IP", detector.PrivateIPv4)
 	if err != nil {
 		log.Logger.Warnw("failed to get private IP", "provider", detector.Name(), "error", err)
 	} else {
--- a/pkg/providers/all/all.go
+++ b/pkg/providers/all/all.go
@@ -68,21 +68,16 @@ func DetectWithRegionOverride(ctx context.Context, regionOverride string) (*pkgp
 		IMDSDetected: pkgproviders.SupportsIMDS(detector),
 	}
 
-	// Metadata fields are best-effort: provider detection has already succeeded,
-	// so one failed optional fetch should not discard provider identity.
-	publicIP, err := detector.PublicIPv4(ctx)
+	// Fetch the required identity fields first: when the metadata service is
+	// slow or hanging, the optional best-effort fetches (public IP, VM
+	// environment, private IP) must not consume the shared context budget
+	// that the login-gating instance ID and region need.
+	instanceID, err := fetchRequiredMetadata(ctx, detector, "instance ID", detector.InstanceID)
 	if err != nil {
-		log.Logger.Warnw("failed to get public IP", "provider", detector.Name(), "error", err)
+		log.Logger.Warnw("failed to get instance ID", "provider", detector.Name(), "error", err)
 	} else {
-		info.PublicIP = publicIP
-	}
-
-	privateIP, err := fetchPrivateIPv4(ctx, detector)
-	if err != nil {
-		log.Logger.Warnw("failed to get private IP", "provider", detector.Name(), "error", err)
-	} else {
-		info.PrivateIP = privateIP
-		log.Logger.Infow("successfully detected private IP", "provider", detector.Name(), "privateIP", privateIP)
+		info.InstanceID = instanceID
+		log.Logger.Infow("successfully detected instance ID", "provider", detector.Name(), "instanceID", instanceID)
 	}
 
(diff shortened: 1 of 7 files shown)


cardyok:
I am worried this retry can consume most or all of the provider-detection timeout.

Private IP is best-effort, but it now gets the same 4-attempt retry path as required fields such as region and instance ID. With the 10-second nscale request timeout, private IP alone can take about 40.7 seconds. Provider detection may already have taken 10 seconds, so region and instance ID may only run after the outer 1-minute context has expired. Login would then fail because those fields are required.

This also affects IMDS providers that do not implement private IP lookup. Azure, GCP, and Nebius currently use a nil private-IP fetcher, so they will always retry four times, add 700 ms, and emit misleading warnings.

Could we keep this retry nscale-specific, or add a capability/not-supported result and give optional private-IP lookup a smaller timeout budget? Fetching the required fields before optional private IP would also prevent it from blocking login.

pkg/providers/all/all.go
```diff
@@ -77,7 +77,7 @@ func DetectWithRegionOverride(ctx context.Context, regionOverride string) (*pkgp
 		info.PublicIP = publicIP
 	}
 
-	privateIP, err := detector.PrivateIPv4(ctx)
+	privateIP, err := fetchPrivateIPv4(ctx, detector)
```
calmkart: This optional private-IP retry runs before the required region and instance-ID fetches, while all three share the login provider-detection timeout.

For nscale, four attempts at the new 10-second HTTP timeout plus backoff can consume about 40.7 seconds of the one-minute total budget. Any time already spent detecting the provider or fetching the public IP leaves even less time for region and instance ID. Login then fails because those later fields are mandatory, even though private IP itself is optional.

Please fetch the required identity fields first, or give private IP a separate and substantially smaller retry budget. The current tests use immediate failures and zero backoff, so a budget/cancellation test would also be useful.

gyuho: Addressed in three pushed commits:
- bc189506: retry only when the detector actually has a private-IP fetcher — the nil fetchers (Azure, GCP, Nebius) return immediately instead of 4 retries + backoff + misleading warnings.
- f10b7534: the retry is now an nscale-only opt-in (`WithPrivateIPv4Retry`); no other provider gets it.
- f9000a58: `DetectWithRegionOverride` fetches instance ID and region first and moves private IP last, and adds `TestDetect_PrivateIPv4RetryDoesNotStarveRequiredFields` — the private fetch blocks until the shared context expires while region/instance-ID mocks fail on a dead context, pinning the fetch order (verified to fail against the pre-reorder code).

gyuho:
@cardyok All three asks landed: bc189506 skips the retry entirely for detectors with no private-IP fetcher (Azure/GCP/Nebius nil fetcher — no more 4 retries / 700ms / misleading warnings), f10b7534 makes the retry an nscale-only opt-in (`WithPrivateIPv4Retry`), and f9000a58 fetches the required identity fields (instance ID, region) before any optional metadata, with a regression test that pins the order against a hanging private-IP fetch.

calmkart (approved):

cardyok (approved):

PR #1322, opened 2026-09-08
[LEP-6503][LEP-6504] fix(helm/gpud): persistence checks on GPU Operator nodes

gyuho:
# [LEP-6503][LEP-6504] Fix BYOK startup and persistence checks on GPU Operator nodes

## Summary

This fixes three failures observed after upgrading the Nightly BYOK cluster to `gpud` `0.13.0-rc.2`:

- `[LEP-6503]` Slow pre-server login could exhaust the liveness budget and restart `gpud` forever.
- `[LEP-6503]` A stale GPU Operator driver-root overlay could expose NVIDIA libraries by exact path while returning an empty directory listing. `gpud` found `libnvidia-ml.so.1`, missed its companion libraries, and failed the CUDA version query with `NVML_ERROR_NOT_FOUND`.
- `[LEP-6504]` On GB200 with driver `595.91.07`, `nvmlDeviceGetPersistenceMode` returned `SUCCESS` with mode `0` even though `nvidia-persistenced` held all four GPU devices open and persistence was effective.

The fixes are narrow and do not add an `nvidia-smi` dependency:

- Add a configurable Kubernetes startup probe. Liveness remains unchanged and starts only after startup succeeds.
- Keep the existing allowlisted NVIDIA companion preload, but add exact-path probes when directory enumeration returns no matches.
- When raw NVML reports persistence disabled, inspect the running persistence daemon's open `/dev/nvidiaN` handles and match them to NVML device minor numbers.

### `[LEP-6503]` Startup restart loop

`gpud run --refresh-session-token` performs control-plane login and machine-info collection before opening its HTTPS listener. On affected GB200 nodes, blocked IMDS calls consumed approximately 30 seconds:

```text
nscale: 10s
azure:  10s
oci:     5s
other providers + ASN/login: several seconds
```

The `0.13.0-rc.2` chart started liveness checks immediately:

```yaml
livenessProbe:
  httpGet:
    path: /healthz
    port: 15132
    scheme: HTTPS
  periodSeconds: 10
  failureThreshold: 3
  timeoutSeconds: 1
```

The third failure arrived as login completed, so kubelet killed the process before port `15132` opened. The observed pods restarted 21 and 56 times. Setting `livenessProbe.initialDelaySeconds=120` stopped the loop, confirming that the process needed startup grace rather than a different liveness policy.

### `[LEP-6503]` Driver-root companion discovery

Node `ipp2-0290` had a separate deterministic crash loop:

```text
failed to create nvml instance: failed to get driver version: Not Found
```

Its stale GPU Operator overlay had inconsistent directory semantics:

```text
ls /run/nvidia/driver/usr/lib/x86_64-linux-gnu
# total 0

stat /run/nvidia/driver/usr/lib/x86_64-linux-gnu/libcuda.so.1
# succeeds
```

The existing loader found `libnvidia-ml.so.1` by exact `os.Stat`, then used `filepath.Glob` to find companions. The empty `readdir` result made glob return no matches, so `libcuda.so.1` was not preloaded. A direct `go-nvml` probe reproduced the split behavior:

```text
nvmlInit_v2                    -> SUCCESS
nvmlSystemGetDriverVersion     -> SUCCESS (580.126.20)
nvmlSystemGetCudaDriverVersion -> NVML_ERROR_NOT_FOUND
```

The fix keeps the existing probe order and preload allowlist. It adds direct `os.Stat` fallbacks for `libcuda.so.1`, `libnvidia-cfg.so.1`, and version-matched `libnvidia-gpucomp.so.<driver-version>`. It does not use `LD_LIBRARY_PATH`, which would expose the driver root's bundled glibc and crash `gpud` with a `GLIBC_PRIVATE` symbol error.

### `[LEP-6504]` Daemon-managed persistence

On `gb-nvl-124-compute17`, persistence mode was the only unhealthy gpud component:

```text
all 4 GPU(s) disabled persistence mode
```

For those same four GPUs:

- Direct `nvmlDeviceGetPersistenceMode` returned `SUCCESS`, mode `0`.
- `nvidia-persistenced --persistence-mode` was running.
- The daemon process held `/dev/nvidia0` through `/dev/nvidia3` open.
- `nvidia-smi` independently reported `Enabled` for every matching UUID during investigation, but the implementation does not invoke or depend on it.

NVIDIA's persistence daemon source provides the decisive invariant: enabling a GPU calls `nvCfgOpenPciDevice`; disabling it calls `nvCfgCloseDevice`; current per-GPU mode is tracked by those handles. Process presence or startup arguments are insufficient because the daemon remains alive when devices are disabled and RPC can change individual GPUs after startup.

### `[LEP-6503]` Helm startup probe

- Add an HTTPS startup probe for `/healthz` on port `15132`.
- Default budget: approximately five minutes (`periodSeconds: 10`, `failureThreshold: 30`, `timeoutSeconds: 5`).
- Keep the existing liveness probe unchanged.
- Allow `startupProbe: null` to disable the probe.
- Document Helm deep-merge behavior for alternate handlers and listener/probe port coupling.

### `[LEP-6503]` NVML companion fallback

- Keep glob-based selection for normal driver roots.
- If globbing returns no entries, probe stable companion SONAMEs directly.
- Derive `libnvidia-gpucomp`'s version from the fully resolved NVML symlink.
- Keep the allowlist limited to NVIDIA companions; never add the driver directory to `LD_LIBRARY_PATH`.

### `[LEP-6504]` Dependency-free persistence fallback

- Run only when NVML reports a supported GPU as disabled.
- Capture each NVML device's minor number.
- Inspect all live `nvidia-persistenced` processes and aggregate open `/dev/nvidiaN` handles.
- Treat a GPU as daemon-managed only when its exact minor number is held open.
- Preserve the raw NVML result as `nvml_reported_enabled`; identify the effective source as `nvidia-persistenced`.
- Fail closed: missing daemon, unreadable/incomplete FD snapshots, or unavailable minor numbers retain the NVML-disabled state and warning event.
- Do not execute host binaries and do not depend on `nvidia-smi`.

### Live BYOK validation

- Patched NVML probe on `ipp2-0290` preloaded all companions despite empty directory listings and completed full instance creation:

```text
NVMLExists=true driver=580.126.20 cuda=13.0 product=NVIDIA-A10 devices=1
```

- On arm64 GB200, the live `nvidia-persistenced` process held all four target devices open:

```text
/dev/nvidia0
/dev/nvidia1
/dev/nvidia2
/dev/nvidia3
```

- NVIDIA's daemon implementation confirms that these open handles are the mechanism used for enabled persistence mode.

### Automated validation

- `go test -race ./components/accelerator/nvidia/persistence-mode/... ./pkg/nvidia/nvml/...`
- `go vet ./components/accelerator/nvidia/persistence-mode/... ./pkg/nvidia/nvml/...`
- `go test ./... -run '^$'`
- `helm lint deployments/helm/gpud`
- Default Helm render contains the startup probe.
- `--set startupProbe=null` omits it.
- `git diff --check`

## References

- NVBug 6737605 / LEP-6503
- NVBug 6737677 / LEP-6504
- NVIDIA persistence daemon implementation: https://github.com/NVIDIA/nvidia-persistenced/blob/main/nvidia-persistenced.c
- NVIDIA Driver Persistence: https://docs.nvidia.com/deploy/driver-persistence/persistence-daemon.html
- Kubernetes probes: https://kubernetes.io/docs/concepts/workloads/pods/liveness-readiness-startup-probes/

### DIFF
--- a/components/accelerator/nvidia/persistence-mode/component.go
+++ b/components/accelerator/nvidia/persistence-mode/component.go
@@ -47,9 +47,10 @@ type component struct {
 	nvmlInstance           nvidianvml.Instance
 	getPersistenceModeFunc func(uuid string, dev device.Device) (PersistenceMode, error)
 
-	// queryEffectivePersistenceModesFunc queries nvidia-smi's current per-GPU
-	// persistence state when the raw NVML API reports disabled.
-	queryEffectivePersistenceModesFunc func(context.Context) (map[string]bool, error)
+	// daemonPersistenceModesFunc returns GPU minor numbers currently held open
+	// by nvidia-persistenced when raw NVML reports disabled.
+	daemonPersistenceModesFunc func(string) (map[int]bool, error)
+	procRoot                   string
 
 	eventBucket eventstore.Bucket
 
@@ -66,9 +67,10 @@ func New(gpudInstance *components.GPUdInstance) (components.Component, error) {
 		getTimeNowFunc: func() time.Time {
 			return time.Now().UTC()
 		},
-		nvmlInstance:                       gpudInstance.NVMLInstance,
-		getPersistenceModeFunc:             GetPersistenceMode,
-		queryEffectivePersistenceModesFunc: queryEffectivePersistenceModes,
+		nvmlInstance:               gpudInstance.NVMLInstance,
+		getPersistenceModeFunc:     GetPersistenceMode,
+		daemonPersistenceModesFunc: daemonPersistenceModes,
+		procRoot:                   "/proc",
 	}
 
 	if gpudInstance.EventStore != nil {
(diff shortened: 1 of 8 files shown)


gyuho:
Validated the released `v0.13.0-rc.3` chart and image against the original failure modes on the affected BYOK environment.

### `[LEP-6503]` Startup and NVML loading

- Both affected GB200 nodes took approximately 31 seconds to complete login, exceeding the old approximately 30-second liveness budget.
- The new startup probe protected startup until `/healthz` became available.
- Both nodes became Ready with zero container restarts.
- The A10 node that previously crash-looped on `NVML_ERROR_NOT_FOUND` also became Ready with zero restarts.
- Runtime logs confirmed that the released image found and preloaded the matching `libcuda`, `libnvidia-cfg`, and `libnvidia-gpucomp` companion libraries.

### `[LEP-6504]` Daemon-managed persistence

- On both GB200 nodes, raw `nvmlDeviceGetPersistenceMode` still reported disabled, reproducing the underlying driver behavior.
- The dependency-free fallback matched all four GPU minor numbers to the open `/dev/nvidiaN` handles owned by `nvidia-persistenced`.
- Repeated checks reported the persistence component as Healthy.
- The component API reported four GPUs with:

```json
{
  "enabled": true,
  "supported": true,
  "nvml_reported_enabled": false,
  "effective_state_source": "nvidia-persistenced"
}
```

- No `nvidia-smi` invocation or runtime dependency is used.

### Result

The released artifacts fix both reported failure modes on every reachable affected node tested. All tested containers remained at zero restarts throughout validation.

One deployment detail: `helm upgrade --reuse-values` does not introduce a newly added chart default such as `startupProbe`. Existing installations must either supply the startup probe values explicitly during this upgrade or upgrade with a values set that includes the new chart defaults.

The overall DaemonSet did not reach full readiness because pre-existing unreachable nodes could not recreate their pods. This was independent of the rc.3 changes; every reachable node tested was healthy on rc.3.

cardyok (approved):
Reviewed the startup-probe, NVML companion-loading, and daemon-managed persistence changes. I found no blocking correctness issues.

The fallback remains fail-closed, matches GPU minors to the exact /dev/nvidiaN handles held by nvidia-persistenced, and the startup probe does not alter liveness semantics. Helm lint/render checks passed, including startupProbe: null disabling the probe; CI and the rc.3 live BYOK validation are also green.

The documented --reuse-values rollout caveat is important: existing releases must explicitly receive the new startupProbe values (or use a complete values set) for that part of the fix to take effect.

PR #1326, opened 2026-09-11
[LEP-6028] feat(containerd): restore dangling pod detection via kubelet read-only port

gyuho:
## Summary

PR #1298 removed the dangling pod detection along with the kubelet component, but the **containerd** component was the actual consumer of that check. Instead of bringing back the whole kubelet component, this PR:

- Moves the kubelet read-only port (`10255`) pod listing code into `components/containerd/kubelet.go` as **private** helpers (`listPodsFromKubeletReadOnlyPort`, `kubeletPodStatus`), keeping only the namespace/name fields needed for the dangling comparison
- Brings back the dangling pod check logic in the containerd component's `Check()` with private thresholds (`danglingDegradedThreshold` = 5, `danglingUnhealthyThreshold` = 10)
- Wires the port probe and pod listing through injectable component funcs (`isKubeletReadOnlyPortOpenFunc`, `listKubeletPodsFunc`), matching the component's existing testability pattern; the check is skipped when the kubelet read-only port is closed

## Behavior (unchanged from before #1298)

- `dangling > 10` → Unhealthy + suggested reboot
- `dangling > 5` → Degraded
- `dangling != 0` → reason annotated
- kubelet read-only port closed or listing error → check skipped, no health impact

## Tests

- `kubelet_test.go`: httptest-based tests for the read-only port client (success, parse error, connection error) and the pod-list parser
- `component_test.go`: `danglingPodCount` unit tests plus table-driven `Check()` tests covering port-closed skip, list-error tolerance, and all threshold transitions

```
go test -race -gcflags="all=-N -l" ./components/containerd/   # PASS
```


cardyok:
We cannot assume kubelet's unauthenticated read-only port (10255) is enabled. Our nodes have it disabled for security reasons, so this implementation would silently skip dangling-pod detection on exactly the nodes where we need it restored. The port-closed unit test currently confirms that blind spot rather than demonstrating restored coverage. Re-enabling 10255 should not be a prerequisite.

Could we make the pod-listing source work with hardened kubelet configurations instead?

- Prefer kubelet's authenticated HTTPS `/pods` endpoint (normally port 10250), with proper serving-certificate/hostname verification and an explicitly authorized client identity. Do not assume the existing kubelet client certificate automatically has permission, or use `InsecureSkipVerify`/anonymous access as a workaround.
- Validate least-privilege authorization for the supported Kubernetes versions. Where fine-grained kubelet authorization is available, prefer node-scoped `get` on `nodes/pods`; avoid granting broad `nodes/proxy` access just for this check. See [Kubernetes kubelet authentication/authorization](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/).
- If listing is unavailable, expose that detection is unavailable (reason/metric), rather than making a skipped check indistinguishable from a successful comparison with zero dangling pods. That does not mean an auth/network failure should itself trigger a reboot.
- Add an E2E case with 10255 disabled and authenticated pod listing working, plus tests for 401/403, TLS failures, timeouts, and a valid empty pod list.

An API-server pod list filtered by `spec.nodeName` could be another design, but it is not automatically equivalent to kubelet's local pod view; its deletion/termination semantics would need explicit validation before substituting it.

cardyok:
The update in `41e269f` removes the 10255 dependency and explicitly reports listing failures without suggesting a reboot. I still see two correctness issues in the API-server-based comparison:

1. **The 10-minute guard measures sandbox lifetime, not how long it has been absent from the API server.**

   At [component.go:641](https://github.com/leptonai/gpud/blob/41e269f7ad34d92a1faae63aa4ab6bb29fe0935f/components/containerd/component.go#L641), the check uses `now - sandbox.CreatedAt`. A day-old pod that was force-deleted just now already exceeds that threshold. If its sandbox is still READY while kubelet tears it down, it is immediately counted as dangling. With more than 10 such sandboxes and another pod still present in the API list, this can produce Unhealthy plus a suggested reboot during ordinary teardown, without waiting 10 minutes after deletion.

   Please base the grace period on sustained, successfully observed absence, keyed by sandbox ID / pod UID, rather than creation time. Reset the observation when the pod reappears or the sandbox disappears, and do not treat query failures as evidence of absence. Add a regression where an old sandbox has only just lost its API pod, then verify no immediate dangling verdict and that a genuinely persistent orphan becomes eligible after the grace period.

2. **A successful empty API pod list still disables detection.**

   [component.go:625-626](https://github.com/leptonai/gpud/blob/41e269f7ad34d92a1faae63aa4ab6bb29fe0935f/components/containerd/component.go#L625) returns zero whenever the reference list is empty, even if containerd still has many orphaned READY sandboxes. A valid empty result is not the same as a failed/unavailable query. The assertion at [component_test.go:4195-4196](https://github.com/leptonai/gpud/blob/41e269f7ad34d92a1faae63aa4ab6bb29fe0935f/components/containerd/component_test.go#L4195) currently locks in this blind spot.

   Please compare against the valid empty set using the same absence grace period, while keeping query failures on the explicit unavailable path. Add a component-level regression for a successful empty API list with persistent orphan sandboxes, including the health thresholds and suggested action.

These are findings from the current code and tests, not a live-node reproduction.

cardyok (approved):

gyuho:
Addressed in 41e269f: dropped 10255 entirely.

The check now lists pods from the API server (`GET /api/v1/pods?fieldSelector=spec.nodeName=<node>`) using kubelet's own client credentials already present on every node:

- Client certificate: `/etc/kubernetes/pki/kubelet-client-current.pem` (re-read every check to follow kubelet client-certificate rotation)
- Cluster CA for full serving-certificate verification: `/etc/kubernetes/pki/ca.crt` — no `InsecureSkipVerify`
- Endpoint from kubelet's kubeconfig (`/root/.kube/config`), honoring `tls-server-name`
- Node name taken from the client certificate CN (`system:node:<name>`)

On authorization: agreed that the kubelet client certificate cannot be assumed to have `nodes/pods` on 10250 — the Node authorizer has no rule for kubelet-serving subresources, so that path would 403 by design (our own clusters grant `nodes/pods` explicitly to per-node agents like machine-daemon/ingress-tunnel-agent for exactly this reason). The API-server list sidesteps it: the node identity may list pods, and our kube-apiserver-proxy force-scopes node-identity pod lists to `spec.nodeName=<self>`, so no new credentials or RBAC are needed and the query is hard-limited to the local node.

Unavailability is explicit: missing identity (non-Kubernetes host) skips with a stated reason; 401/403/TLS/timeout/network failures report "dangling pod detection unavailable" and never suggest a reboot. On the API-view divergence you flagged: graceful deletes keep the pod object until kubelet confirms teardown, so the remaining gap is the force-delete teardown window — that is handled by the sustained-absence grace period added in the follow-up commit (see next comment).

Tests: 401/403, TLS verification failure, client-cert rejection, timeout, connection error, valid empty pod list, and `tls-server-name` override. One honest gap: a true E2E with 10255 disabled needs a Kubernetes cluster fixture, which the current e2e harness (gpud binary + mocks) does not have — the httptest TLS servers with `RequireAndVerifyClientCert` exercise the full production path (handshake → client-cert auth → field selector → decode) in the meantime.

gyuho:
Both findings addressed in 6d1048ba.

1. **The verdict now keys on sustained absence, not sandbox age.** Per-sandbox-ID tracking records when a pod was first missing from a successfully observed API list; a sandbox counts as dangling only after 10m of continuous absence; the observation resets when the pod reappears and is forgotten when the sandbox leaves READY or disappears. Query failures are never treated as evidence of absence — and they also do not reset a running observation. A day-old pod force-deleted a second ago is no longer immediately dangling.

2. **The empty-list short-circuit is gone.** A valid empty result is authoritative, not a blind spot: every READY sandbox on a pod-less node becomes eligible after the same grace period, while failed queries keep the explicit "dangling pod detection unavailable" path.

Regressions added for both, per your suggestions: an old sandbox that has only just lost its API pod (no immediate verdict, eligible after grace); a successful empty API list with persistent orphan sandboxes crossing the unhealthy threshold *with* the suggested reboot action; pod-reappearance resets the observation; and failure-between-observations neither seeds nor resets absence.

PR #1329, opened 2026-09-15
fix(kap-mtls): reload renewed client certificates without restarting agent

cardyok:
## Summary

- Keep original credential validation and immutable-generation atomic writes. Running agents receive SIGHUP; inactive updates return immediately. The existing package controller and package init.sh own all service installation, startup and recovery.
- Remove certificate-metrics parsing, historical-generation scans, loaded-certificate confirmation, restart/rollback orchestration and all dedicated lifecycle commands.
- Keep one durable pending-notification marker to cover a failed HUP or GPUd exiting after selection. Read-only Status reports selected disk credentials and masks AgentReady while notification is pending; the existing Activate command retries the notification without reissuing credentials. No new wire fields or RPCs.
- Success means selected files plus delivered notification, not proof that asynchronous certificate loading finished. CA, client-CA fingerprint, endpoint and server-name changes require separate maintenance; corrupt selected configuration fails closed. Historical generations and process startup state are not inferred.

Related: [LEP-6633](https://jirasw.nvidia.com/browse/LEP-6633).

## Verification

- Independent KAP race tests repeated three times: passed. Shared package-lock race, focused session KAP/wire/compatibility tests and full-session non-race tests passed under Go 1.25.12 with the repository's Mockey compiler flags.
- Focused builds, formatting, Markdown checks and seven standard linters passed. Architecture, adversarial error-flow and final semantic audits report no HIGH/MEDIUM findings.
- DEV02 cardy new-node join used this minimal candidate from first login and reached Joined/Healthy, with startup entirely owned by the existing package flow.
- On the plain candidate, a staged pending-notification marker was cleared by the real existing control manager's Activate path. The same certificate remained selected, and agent/kubelet PID, start time and restart counters remained unchanged.
- Two actual manager-issued certificate rotations passed 51 checks, including at least 125 seconds of unarmed stability after the second rotation. Agent and kubelet did not restart; one persistent TLS HTTP/1.1 connection never reconnected. Fresh verified TLS GET /version returned HTTP 200.
- Restored the plain candidate, then stopped only this disposable node's agent. The existing package controller recovered it, confirmed by its `kap-mtls-agent status not ok, restarting` log. Credential renewal itself required no service restart.
- The test Machine was released/deleted, exact DEV EC2 termination waiter completed, and the temporary loopback API forward was closed. No shared environment resources were changed.

The renewal stimulus was a test-only Go overlay outside the PR. It only made
outgoing status appear due for the exact Machine behind a root-owned,
serial-bound, short-lived marker. Certificate issuance, authenticated delivery,
Manager.UpdateCredentials and the notification code were real and unchanged.
The plain candidate covered initial join, notification retry and service recovery.

The DEV02 run above used b80cddb8 before the unrelated bootstrap audit work was
split out. This scope-only revision leaves `pkg/kapmtls` byte-identical to that
tested implementation and restores all four session files to the original PR
base. No new node was created for this split. Session compatibility tests were
rerun with race three times; separate bootstrap audit/capability work is not
part of this PR and will be reviewed with the shared ECR trust-anchor work.

Full configured local lint could not load the nilaway plugin; seven standard
linters ran separately, and remote golangci-lint passed. Repository-wide local
build/test/lint was not run. Remote CI passed all 18 checks on b80cddb8;
that result predates the scope-only split. The split is pushed as 11281e35;
new-head CI is running and is not yet claimed passing.
No production, shared R2, NodeClass,
controller or other-node changes were made.

## Compatibility

- Requires a SIGHUP-capable mtls-agent and functioning generic package recovery.
- Certificate fields describe selected files. AgentReady combines the existing readiness probe with absence of pending notification; it does not prove which certificate or startup configuration the process uses.
- Failed asynchronous reload remains visible in mtls-agent logs. This PR intentionally does not poll metrics or restart the agent to establish activation proof.
- The older Lepton KAP spec's dedicated restart/rollback description describes old GPUd; no Lepton runtime change is required for this notification retry path.


pkg/kapmtls/manager.go
```diff
@@ -186,13 +208,48 @@ func (m *Manager) UpdateCredentials(ctx context.Context, machineID string, crede
...
 	if err := m.swapCurrentSymlink(releaseID); err != nil {
+		if loaded.installed {
+			if current, readErr := m.currentReleaseID(); readErr != nil || current != loaded.releaseID {
+				return m.rollbackReload(ctx, loaded, err)
+			}
+		}
 		return err
 	}
+	if loaded.installed && loaded.sameStartupConfig(selected) &&
+		(loaded.releaseID == releaseID || loaded.serial != selected.serial || !loaded.notAfter.Equal(selected.notAfter)) {
```
gyuho: The certificate metrics identify the loaded cert/key, but they do not identify the process startup configuration. The agent’s SIGHUP handler reloads only the cert/key; the gateway endpoint and CA pool remain those loaded at process startup. If GPUd is interrupted after selecting generation B (with an endpoint/CA change) but before restarting the agent, and the agent subsequently receives SIGHUP, `loadedRelease()` associates B’s serial/expiry with B’s on-disk startup configuration even though the process is still using A’s. This branch then treats a retry of B as already activated and skips the required restart; `Status()` also reports B’s configuration as installed, so reconciliation can leave the old endpoint/trust pool in use indefinitely. I reproduced this with a focused race-enabled test: stage/select B, model the certificate-only reload, then retry `UpdateCredentials`; it returns success without a restart and status reports the new endpoint. Please track startup activation separately from certificate identity (with process-bound activation evidence or runtime configuration reporting), and use restart recovery when that evidence is unavailable.

gyuho (approved):

PR #1332, opened 2026-09-16
[LEP-6792] fix(containerd): add runtime diagnostics without changing health checks

gyuho:
When the existing NVIDIA containerd configuration check fails, GPUd reports a generic missing-configuration reason. This change supplements that reason with the running daemon's native CDI setting, default handler, configured handlers, and an example READY sandbox requiring a missing handler.

The existing health policy is unchanged. All on-disk checks, import handling, NVIDIA-handler requirements, toolkit startup grace, and their original tests are preserved. Socket, service, CRI, and dangling-pod checks are unchanged as well. This PR does not relax the health check for CDI-only configurations.

Only after the existing check sets Unhealthy does GPUd issue an additional verbose CRI status request, bounded to five seconds. It appends diagnostic details without changing health, the original reason, error, or suggested actions. Missing or malformed status, RPC failures, and apparently healthy live settings cannot clear or downgrade the existing verdict. Healthy checks do not issue this extra request.

Example appended detail:

```text
live containerd: native CDI enabled=true, default handler "runc", configured handlers ["runc"]; pod gpu-operator/device-plugin requires runtime handler "nvidia", but containerd does not have it configured
```

READY sandboxes supply examples of handler use; they are not an exhaustive inventory of workload requirements. Pending pods without sandboxes are not covered by these supplemental details. The disk checks remain in place to preserve configuration and restart-safety coverage. No dependencies, Kubernetes permissions, or chart behavior changes are introduced.

Validation:
- Original configuration and component tests are restored unchanged from the base branch.
- Added tests compare health, error, suggested actions, and original reason with diagnostics enabled versus disabled. Cases cover CDI-only live status, missing required handlers, broken disk configuration with healthy live settings, RPC/malformed/missing status, healthy config, startup grace, missing toolkit, and disk-read failure.
- `go test -race -gcflags='all=-N -l -d=checkptr=0' ./components/containerd` passed.
- `go build ./components/containerd` and `go vet ./components/containerd` passed.
- Repository-enabled golangci-lint linters passed using a temporary config removing only the unavailable, inactive nilaway custom-plugin registration.
- `git diff --check` passed. README markdownlint reports the same six pre-existing MD013 violations as the base.

Operational validation of the separate configuration repair: both investigated nodes recovered to Healthy on their existing GPUd version after adding the NVIDIA drop-in import and restarting containerd. The running container IDs were preserved on both nodes. This PR has not been deployed to them.


gyuho:
For context, the two investigated nodes had an NVIDIA runtime drop-in on disk, but their main containerd configuration did not import it. Live CRI status showed only the `runc` handler loaded, with native CDI enabled.

The missing connection in `/etc/containerd/config.toml` is a top-level import, before any TOML section:

```diff
 version = 4
+imports = ["/etc/containerd/conf.d/99-nvidia.toml"]
 
 [plugins."io.containerd.cri.v1.runtime".containerd]
   default_runtime_name = "runc"
 
 [plugins."io.containerd.cri.v1.runtime".containerd.runtimes.runc]
   runtime_type = "io.containerd.runc.v2"
```

The existing `/etc/containerd/conf.d/99-nvidia.toml` contains the relevant settings:

```toml
version = 4

[plugins."io.containerd.cri.v1.runtime"]
  enable_cdi = true

[plugins."io.containerd.cri.v1.runtime".containerd.runtimes.nvidia]
  runtime_type = "io.containerd.runc.v2"

[plugins."io.containerd.cri.v1.runtime".containerd.runtimes.nvidia.options]
  BinaryName = "/usr/local/nvidia/toolkit/nvidia-container-runtime"
```

These are shortened examples, not replacement configuration files. An existing imports list should be preserved and extended as appropriate; the merged configuration must be validated before activation.

Native-CDI workloads using `runc` do not inherently need the NVIDIA handler. However, on one of the nodes we verified a device-plugin pod specifying `runtimeClassName: nvidia`, and that RuntimeClass maps to handler `nvidia`. Its missing handler is a real configuration mismatch and could prevent sandbox recreation; we did not reproduce that failure or independently verify the second node's pod requirement.

This PR separates those cases: native CDI can pass without an NVIDIA handler, while a READY CRI sandbox requiring a missing handler remains Unhealthy with a pod-specific reason. No node configuration changes or deployment of this PR have been performed.

eahydra (approved):

PR #1333, opened 2026-09-18
[LEP-6855] fix(containerd): support configurable runtime installations

gyuho:
## Summary
- Add explicit containerd endpoint, config path, service name, and systemctl command-prefix configuration with matching `gpud run` flags.
- Keep legacy defaults and exported helpers; pin explicit configurations to one runtime so missing sockets cannot be hidden by image binaries or another runtime.
- Use the configured runtime for health checks and machine metadata during login, API reads, and gossip.
- Derive Helm directory mounts from the selected paths and document RKE2/k3s examples inline in values.yaml.

## Root cause
Containerized GPUd can find the containerd binary bundled in its own image while probing the standard host socket and systemd unit. RKE2 uses a different socket/config location and runs containerd under its agent/server service. The standard-path probe therefore reports a missing socket even when the actual runtime is healthy. Overriding the existing service-active command alone cannot fix this because socket validation happens first.

## Compatibility
- Unset runtime fields preserve existing detection, CRI-O fallback, and health thresholds.
- Existing active-command overrides retain precedence. For explicitly configured runtimes, arbitrary overrides disable uptime-based grace rather than guessing their service manager.
- Default Helm rendering retains the existing service-active command; null now means chart-generated default, while an explicit empty string still removes the override.
- New runtime Helm settings require a GPUd image supporting the new flags.

## Validation
- Scoped race tests passed with repository-required mockey compiler flags across containerd, config, systemd, run/command, login, server, session, machine-info, and Helm tests.
- Real local Unix-socket CRI regression test covers configured endpoint enumeration, missing/stale sockets, and socket recreation/recovery.
- Scoped Go build passed.
- Repository custom-gcl lint: 0 issues.
- Helm lint and default/RKE2/override render tests passed.

Live-cluster rollout verification remains pending.


cmd/gpud/command/command.go
```diff
@@ -272,6 +272,22 @@ sudo rm /etc/systemd/system/gpud.service
 					Usage: "command used to check whether the containerd service is active (e.g. 'nsenter --target 1 --mount -- systemctl is-active containerd' to query the host service manager from inside a container; exit code 0 means active); leave empty to use the built-in systemd check",
 					Value: "",
 				},
+				cli.StringFlag{
```
eahydra, line 275: I don't think it works on BYOK as GPUd daemonset is installed by us not users. We don't even know what the runtime is.

eahydra: never mind. I thought we should we can detect what the runtime is. We should keep these flags for unknown cases. And Zhiting created a MR for integrating these flags.

eahydra (approved):

